PR #1, opened 2016-02-18
Add a physics ruleset describing collation and numbering for physics.

InconceivableVizzini:
Dunno where to shove these thoughts so just creating a PR so they are somewhere...
```

XXXX ——

<div data-type=“page”> should be <section data-type=“section”>. These aren’t “pages”, they are highest-level sections. This is a format describing a book structure, not our storage structure.


XXXX ——

This markup is already dirty… Having <div data-type=“document-title”> all over the document, etc etc.
We need to design more before we implement.

Get rid of <div data-type='document-title’> and use <h1>?



WAT?

<section data-depth="1" id="fs-id3108942" class="section-summary"><h1 data-type="title"><em data-effect="italics"><strong data-effect="bold"><strong data-effect="bold"><em data-effect="italics">Summary</em></strong></strong></em></h1><em data-effect="italics"><strong data-effect="bold"><strong data-effect="bold"><em data-effect="italics">
            <ul id="fs-id3155213">
                <li id="import-auto-id1961798">Science seeks to discover and describe the underlying order and simplicity in nature.</li>
                <li id="import-auto-id2575847">Physics is the most basic of the sciences, concerning itself with energy, matter, space and time, and their interactions.</li>
                <li id="import-auto-id2757516">Scientific laws and theories express the general truths of nature and the body of knowledge they encompass. These laws of nature are rules that all natural processes appear to follow.</li>
            </ul>
        </em></strong></strong></em></section>

XXXX ——

“types” are (optionally) nest-able content.
“parts” are the various parts out of which types are produced.

<div data-type=“chapter”>
  <div data-part=“title”>
   <span class=“titling” />
  </div>
</div>

```

### DIFF
--- a/cte-collation-poc/physics-ruleset.less
+++ b/cte-collation-poc/physics-ruleset.less
@@ -1,0 +1,148 @@
+.\=(@counters) {
+  counter-reset: @counters;
+}
+.\+(@counters) {
+  counter-increment: @counters;
+}
+div[data-type='book']{
+  .\=(chapter);
+}
+div[data-type='chapter'] {
+  .\=(section figure table list);
+  .\+(chapter);
+
+  dl.definition {
+    copy-to: eoc-key-terms;
+  }
+  section.section-summary {
+    move-to: eoc-section-summary;
+  }
+  section.conceptual-questions {
+    move-to: eoc-conceptual-questions;
+  }
+  section.problems-exercises {
+    move-to: eoc-exercises;
+  }
+
+  // <div data-type="document-title">Stuff</div>
+  // becomes
+  // <div data-type="document-title"><span class="titling">1</span><span class="titling">Stuff</span></div>
+  > div[data-type='document-title'] {
+    string-set: chapter-title content();
+    content: '';
+
+    &::span {
+      class: 'titling';
+      content: counter(chapter);
+    }
+    &::span {
+      class: 'titling';
+      content: string(chapter-title);
+    }
+  }
+
+  div[data-type="page"] {
+    .\=(subsection);
+    .\+(section);
+
+    > div[data-type='document-title'] {
+      string-set: section-title content();
+      content: '';
+
+      &::span {
+        class: 'titling';
+        content: counter(chapter) '.' counter(section);
+      }
+      &::span {
+        class: 'titling';
+        content: string(section-title);
+      }
+    }
+    > section {
+        .\+(subsection);
+    }
+  }
+
+  .eoc-section-summary > section h1.title {
+    string-set: section-title content();
+    content: '';
+
+    &::span {
+      class: 'titling';
+      // This implies that group-by stores a "xref" attribute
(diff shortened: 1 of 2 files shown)


InconceivableVizzini:
There's also a lot of weird stuff in this DOM...

```
<em data-effect="italics"><strong data-effect="bold"><strong data-effect="bold"><strong data-effect="bold">
    <section data-depth="1" id="fs-id3137968" class="conceptual-questions" data-element-type="conceptual-questions"><h1 data-type="title">Conceptual Questions</h1>
 etc
```

reedstrm:
@InconceivableVizzini could you copy these PR comments into a file under ../docs?

Anything other than the toplevel structure you see in this DOM comes from the existing CNXML->HTML transform.  Modifying that is in scope for CTE, but we want to keep it under control (no throw-it-all-out-start-from-scratch :-) )

cte-collation-poc/physics-ruleset.less
```diff
@@ -0,0 +1,148 @@
+.\=(@counters) {
```
philschatz: Ooh, interesting mixin syntax!

InconceivableVizzini: I really dislike that backslash... .+ .- .= etc seemed so cute :(

reedstrm:
Going to merge this as is, even though I have some misgivings. Wondering if we should use standard selector (:after :before), but w/ custom declaration, like the proposed "label". something like content-wrapper: "block" or "inline" (or "div" or "span")

philschatz:
For what it's worth, I'm a fan of using `::after(123)` so that:

1. nested selectors work for adding things like that label
2. it's more CSS-y

so, adding a label would look something like:

```less
// LESS example
[data-type='page']::after(1) {
  &::after(1) { content: 'Homework'; }
  content: pending(hw-problems);
}

// "compiled" CSS
[data-type='page']::after(1)::after(1) { 
  // the "label"
  content: 'Homework';
}

[data-type='page']::after(1) {
  content: pending(hw-problems);
}

```

InconceivableVizzini:
I really like this style of nesting pseudo-element-generating selectors.

reedstrm:
O.k. Here's my condensed thoughts on this. copy-to, move-to, stacked pseudo-classes, including their numbering are all part of a 2003(!) draft CSS3 spec (CSS3-content) that was revised in 2013, to not include _any_ of that, and even that version is marked "Rewriting" with the note "Severally outdated".

So, that means _no one else has ever implemented this_. Which means there's no proof-by-existence that it's even implementable. So, I'm a little resistant to try and do the full-spec, just because it's there. I'm going to implement just enough to do what we need for cooking books.

More practically, there's no indication that we're going to actually generate nodes and wrap this stuff. I'm going to propose:
```
display: inline-wrapped
display: block-wrapped
```
or
```
display: wrapped-inline
display: wrapped-block
```

PR #2, opened 2016-02-19
Math comparison tool

Dantemss:
- Added optional mathjax_version as the last param of fullbook.py (will insert a mathjax script tag for that version)
- Fullbook title is now a link to archive that also shows the uuid and version on the tooltip
- Added extractmath.py to extract all the math from the fullbook output (preserves the book title)
- Added fullbook_mathonly.py to do both steps together

### DIFF
--- a/cte-collation-poc/extractmath.py
+++ b/cte-collation-poc/extractmath.py
@@ -23,8 +23,7 @@ def main(html_in, html_out=sys.stdout):
 
     math = math_xpath(body)
 
-    for c in body.iterchildren():
-        body.remove(c)
+    body.clear()
 
     for m in math:
         body.append(m)
--- a/cte-collation-poc/extractmath.py
+++ b/cte-collation-poc/extractmath.py
@@ -26,6 +26,7 @@ def main(html_in, html_out=sys.stdout):
     body.clear()
 
     for m in math:
+        m.tail = None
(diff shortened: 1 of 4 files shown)


philschatz:
:+1: for the feature!!! What does the file look like? (could you add a screenshot to the PR body?)

Dantemss:
Here it is:

screen shot 2016-02-19 at 3 45 08 pm

edwoodward:
Can you add a link to the content the math came from or at least the id?

Dantemss:
This particular screenshot above came from 27275f49-f212-4506-b3b1-a4d5e3598b99
That is, http://cnx.org/contents/JydfSfIS@2.2:HR_VN3f7@3/Introduction-to-Science-and-th

Dantemss:
Oh I see what you mean :) I'll add it to the script

Dantemss:
OK the book title is now a link to archive.

Dantemss:
Ready for review. This works fine with College Physics at least, as far as I can tell. If there's a book that's too big for Mathjax to handle properly, we could split up the processing into smaller chunks.

reedstrm:
I converted the optional positional argument to an optional dashed argument '-m' '--mathjax_version' which takes an optional version number, defaulting to 'latest'

PR #4, opened 2016-03-04
Draft an `algebra.html` for discussion

pumazi:
This is the results of doing

```
export epub (returns internal epub) -> transform to single-html (returns html file)
```
which is the first part of
```
export epub (returns internal epub) ->
collate [transform to single-html (returns html file) ->
         cook single-html  (returns html file) ->
         reconstitute (returns models)] (returns models) ->
export to epub (returns epub)
```

### DIFF
--- a/single-html/README.md
+++ b/single-html/README.md
@@ -1,5 +1,24 @@
 # Drafting of the Single HTML format
 
+## Structure
+
+Recreate the `structure.html` file using:
+
+```py
+>>> import os
+>>> import cnxepub
+>>> cnxepub_dir = os.path.abspath(os.path.dirname(cnxepub.__file__))
+>>> book_dir = os.path.join(cnxepub_dir, 'tests', 'data', 'book')
+>>> epub = cnxepub.EPUB.from_file(book_dir)
+>>> binder = cnxepub.adapt_package(epub[0])
+>>> html = bytes(cnxepub.SingleHTMLFormatter(binder))
+>>> with open('structure.html', 'w') as f:
+...    f.write(html)
+```
+
+
+## Example: Algebra
+
 Recreate the `algebra.html` file using:
 
 ```
--- a/single-html/README.md
+++ b/single-html/README.md
@@ -1,0 +1,11 @@
+# Drafting of the Single HTML format
+
+Recreate the `algebra.html` file using:
+
+```
+ssh cte-cnx-dev.cnx.org /var/cnx/venvs/archive/bin/cnx-archive-export_epub /etc/cnx/archive/app.ini 9b08c294-057f-4201-9f48-5d6ad992740d@3.278 algebra.epub
+scp cte-cnx-dev.cnx.org:~/algebra.epub .
+cnx-epub-single_html algebra.epub > algebra.html
+```
+
+Note, we choose algebra, because it's about 1/5 the size of physics.
--- a/single-html/structure.html
+++ b/single-html/structure.html
@@ -1,0 +1,335 @@
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:lrmi="http://lrmi.net/the-specification">
+  <head itemscope="itemscope" itemtype="http://schema.org/Book">
+
+    <title>Book of Infinity</title>
+    <meta itemprop="inLanguage" data-type="language" content="None"/>
+
+    <!-- These are for discoverability of accessible content. -->
+    <meta itemprop="accessibilityFeature" content="MathML"/>
+    <meta itemprop="accessibilityFeature" content="LaTeX"/>
+    <meta itemprop="accessibilityFeature" content="alternativeText"/>
+    <meta itemprop="accessibilityFeature" content="captions"/>
+    <meta itemprop="accessibilityFeature" content="structuredNavigation"/>
+
+
+    <meta itemprop="dateCreated" content="2013/03/19 15:01:16 -0500"/>
+    <meta itemprop="dateModified" content="2013/06/18 15:22:55 -0500"/>
+  </head>
+  <body xmlns:bib="http://bibtexml.sf.net/" xmlns:data="http://www.w3.org/TR/html5/dom.html#custom-data-attribute" itemscope="itemscope" itemtype="http://schema.org/Book">
+    <div data-type="metadata">
+      <h1 data-type="document-title" itemprop="name">Book of Infinity</h1>
+      <span data-type="cnx-archive-uri" data-value="9b0903d2-13c4-4ebe-9ffe-1ee79db28482@1.6"/>
+      <div class="authors">
+        By:
+<span id="author-1" itemscope="itemscope" itemtype="http://schema.org/Person" itemprop="author" data-type="author">
+            <a href="https://github.com/marknewlyn" itemprop="url" data-type="github-id">Mark Horner</a>
+          </span>, <span id="author-2" itemscope="itemscope" itemtype="http://schema.org/Person" itemprop="author" data-type="author">
+            <a href="https://cnx.org/member_profile/sarblyth" itemprop="url" data-type="cnx-id">Sarah Blyth</a>
+          </span>, <span id="author-3" itemscope="itemscope" itemtype="http://schema.org/Person" itemprop="author" data-type="author">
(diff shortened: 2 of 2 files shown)


InconceivableVizzini:
`<div data-type="resources" style="display: none;"></div>`

Can this be avoided?

pumazi:
The resources section was a newly created metadata bit that comes out of [Implement the encapsulation of Resources within a Binder](https://trello.com/c/1b73hNXN/31-implement-the-encapsulation-of-resources-within-a-binder).

It's a necessary bit of information which will be needed in the reconstruction of the data back into machine readable data. Basically, all resources are mostly registered by hrefs, src, etc. (aka `Reference`s) in the content, but this isn't the case for Binders/Books, which does not contain content type data and thus contains no references.

InconceivableVizzini:
`<body .....>`

can be

`<body data-type="book">`

-

The "most top level" selector a framework based stylesheet would use.

InconceivableVizzini:
`<div data-type="metadata">`

This should probably just be a section? For my part I think all of the metadata and table of contents should be under a top level section type.

`<section data-type="section" class="titlepage">... metadata ... ...toc ... </section>`

-

Similarly, `<div data-type="page">...</page>`

can be `<section data-type="section">...</section>`

InconceivableVizzini:
Under metadata, `data-type=""` can instead be `class=""` and `data-part=""`. eg, `<h1 data-part="title"></title>` and `<span data-type="cnx-archive-uri">` can be `<span class="cnx-archive-uri">`, etc.

InconceivableVizzini:
For the `id="toc"` nav element, we should investigate further with the various book structures and develop classes for the various parts of the toc. An example of how we use these can be seen with the high school physics design.

InconceivableVizzini:
`<div data-type="document-title">`

Helene and I have ideas with regards to a set of markup for all titles that will allow enough flexibility for all of our designs.

-

Something along the lines of...

`<div data-part="title"><span class="titling" /><span class="titling" /><span class="titling" /><span class="titling" /></div>`

Primarily we tend to require numbering, a divider, a label, and an element to contain the wording (text node). This split gives us enough control for various types of visual distinction, and why we need it can be seen with the high school physics design.

InconceivableVizzini:
`<p id="delete_me"></p>`

Um...

helenemccarron:
avoid different methods for handling titles <div data-type="document-title">  <h2 data-type="title">.

 <div data-type="document-title">  for section titles is confusing.

helenemccarron:
Can we avoid duplicate data-type/class? 
``<div data-type="solution" class="solution" >`` seems redundant

InconceivableVizzini:
After considering the "page" to "section" switch, the old CNXML tag `<section>` becomes strange. I propose these transform into `<div data-type="subsection"></div>` or `<div data-type="topic">...</div>`.


The storage chunk "page" is logically a top level "section" of the book, and I would like the data-types to reflect that. The CNXML "section" is a subheading/subsection in a section of the book.

InconceivableVizzini:
`
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.
All chapters are broken down into multiple sections, the titles of which can be viewed in the Table of Contents.`

In the preface, is this in the content, or is there a bug?

InconceivableVizzini:
`div data-type="abstract">` We need to get together and nail down this whole "Chapter Outline" and "Learning Objectives" stuff...

helenemccarron:
``<math xmlns="http://www.w3.org/1998/Math/MathML" display="block">`` or ``<math xmlns="http://www.w3.org/1998/Math/MathML" display="inline"> ``
Should we avoid inline styles?

InconceivableVizzini:
@helene regarding the math tag, and some other tags, such as svg, inline styles may be unavoidable and is the generally accepted way they are entered into markup. Eg the stroke="" etc attributes. I don't really like this either...

helenemccarron:
``<table id="Table_04_00_01" summary="">`` I can't remember if this applies to algebra but in general, we need to make sure that all classes applied to tables are transferred.

pumazi:
Reply to metadata...

Keep in mind that everything within the `<div data-type="metadata">` block is added by cnx-epub. This block cannot be moved to the top of the document. It needs to remain with the individual documents themselves (for reconstruction). (*In publishing, we remove this block before committing the content to archive.*)

The metadata blocks could potentially be moved to the top of the document and the blocks could be linked by id. But we can't collapse that metadata into one blob without doing a lot of work to parse it back out again.

pumazi:
Replying to titling...

The `data-type="document-title"` is something we add from cnx-epub. The other `<h# data-type="title"`>` is from transforms (cnxml `<title>` -> html `<h# data-type="title">`). So they aren't the same and as such need to be treated differently. For me, using `document-title` was that solution.

In the beginning I started with `<h1>` for the document-title, then ran into issues because that element is actually used in the content. And there needed to be a distinction between the document-title and a title used in the content (for parsing out by xpath).

InconceivableVizzini:
I don't like that this document is tied to any type of logic about reconstruction etc etc... this document should be precisely what we need for rendering, not for any other purpose.

pumazi:
@helenemccarron in response to:

> Can we avoid duplicate data-type/class? 
`<div data-type="solution" class="solution" >` seems redundant

This is done in the transform cnxml->html5. ([xsl snippet of the section where it happens](https://github.com/Connexions/rhaptos.cnxmlutils/blob/master/rhaptos/cnxmlutils/xsl/cnxml-to-html5.xsl#L162-L165)) Looks like we can probably remove this, but I'd suggest we not do those kind of changes at the same time. Let's add an issue to rhaptos.cnxmlutils to remove this?

pumazi:
@InconceivableVizzini in response to
> I don't like that this document is tied to any type of logic about reconstruction etc etc... this document should be precisely what we need for rendering, not for any other purpose.

Please see my awful diagram of inputs and outputs for the various pieces of logic. We do need to reconstruct the cooked html output so that it can be put back into the archive.

If it's the PDF single-html file you are thinking of, **we can create another formatter** (similar to `cnxepub.adapters.SingleHTMLFormatter`) to format the EPUB (models) to a single-html file that contains ONLY what you need.

reedstrm:
Mostly good comments, the expectation is that PDF rendering will start from the encapsulated epub, which will not contain this particular single html file. If you absolutely hate the idea of a few warts to cover the 'cook and burst' case, we could add an option or alternative formatter to strip that out.

I'm envisioning a textbook styling workflow going something like - 
### Textbook-two-step: collation & numbering, then styling

-   install cnx-epub
-    grabbing a full epub, unzipping it locally, 
-    run local cook against that directory with a local  recipe CSS - 
         in this case, local cook does the assemble/cook/burst steps
-    check collation/numbering behavior
-    rinse/repeat
-    update styling
-    check page-by-page style (webview way)
-    generate single-html (potentially sub-sampling)
-    check fullbook styling (in browser, but testing for PDF gen)
-    run pdfgen against directory (includes it's own fullbook assembly step)
-    check PDF
-    commit diffs to book recipe CSS and styling CSS

Alternatively, local cook could leave the 'all one file' around as well as bursting it.

pumazi:
Furthermore, the PDF transform circuit looks more like this:

```
export epub containing cooked contents  (returns internal epub) ->
send to transformation services [
    transform to single-html  (returns html file) ->
    produce pdf  (returns pdf file)
    ]  (returns list of artifacts containing a reference to the pdf)
```

So again, the single-html for the PDF could be a specific adaptation of the models into a format that meets *exactly* what you want.

helenemccarron:
@reedstrm the one thing I am wondering in the scenario you posted is testing. That workflow is perfect for development but how do testers generate the pdf in conditions that are similar to production and ensure that everything will look the same?

Or are we uploading PDF manually all the time, without any pdf generation on the server at all (which means that nobody can create a pdf other than Openstax)?

reedstrm:
@helenemccarron No, the automated PDF generation will use the path @pumazi outlined in the comment directly above yours. :-)

pumazi:
During the moving of elements to composite (or virtual) pages, can we use `data-type="composite-page"` over `data-type="page"`? This will help us in the reconstitution of the models. I have a feeling it will also be beneficial down the line when working with the content in authoring.

reedstrm:
I'm going to close this PR - I think all the comments have been addressed, or documented elsewhere.

PR #6, opened 2016-04-05
All rulesets

helenemccarron:
All rules are present, will undergo modifications as we develop them.


philschatz:
Would it make sense to start creating mixins for the books? It seems like there is a lot of copy/pasta to move all the pieces around. ie (terrible names, I know):

```less
section.practice-test { .x-move-src(eoc-practice-test); }
.x-move-dest(eoc-practice-test; eoc-all);

// the mixins
.x-move-src(@bucket) { move-to: @bucket; }
.x-move-dest(@bucket; @dest-bucket) {
  &::after {
    container: div;
    class: @bucket;
    content: pending(@bucket);
    move-to: @dest-bucket;
  }
}
```

Also, my poor little laptop has trouble scrolling through the diff (& probably downloading this repo)

helenemccarron:
@philschatz: I've been thinking about it for a while, I will go over it and refactor the less when I'm done actually creating each element. Right now, we keep changing it as we go and I prefer to keep visible. As a matter of fact, even that include is problematic right now. It's definitively not optimized, that's for sure :)

InconceivableVizzini:
One of those mixins seems superfluous?

`section.practice-test { move-to: eoc-practice-test; }
.x-move-dest(eoc-practice-test; eoc-all);

.x-move-dest(@bucket; @dest-bucket) {
  &::after {
    container: div;
    class: @bucket;
    content: pending(@bucket);
    move-to: @dest-bucket;
  }
}`

PR #17, opened 2016-05-19
TOC links format

helenemccarron:
Document TOC doesn't have h1 or div, has links on each li element and is pulling the inner structure of the titles (spans)


books/rulesets/common/toc.less
```diff
@@ -24,31 +22,33 @@ body {
   }
   > div[data-type='chapter'] {
       > h1[data-type='document-title'] {
-      pass: 20;
-      copy-to: eoc-toc;
+        container: div;
```
reedstrm: duplicate container directive

Issue #18, opened 2016-05-19
Statistics - Answer Collation Issues


Stackblocks:
There is a set of mysteriously collated answers at the start of the Chapter 1 answer key. These don't have a readily searchable identifier, so i'm not sure where they came from. Appears to continue intermittently through the rest of the answer key.

![image](https://cloud.githubusercontent.com/assets/12678235/15430535/54cdf666-1e6b-11e6-8056-c34f459679f2.png)

![image](https://cloud.githubusercontent.com/assets/12678235/15430543/5d396ea2-1e6b-11e6-960e-5c557c949999.png)

Stackblocks:
There is another set of mysteriously collated answers below that.

![image](https://cloud.githubusercontent.com/assets/12678235/15407900/03479268-1dd3-11e6-99cd-a974ad622730.png)

![image](https://cloud.githubusercontent.com/assets/12678235/15407881/ef689562-1dd2-11e6-90ec-89e0b470a448.png)

Stackblocks:
Mysteriously appearing answers also present in other chapters.

Stackblocks:
Showing as resolved in book baked with new ruleset. 

![image](https://cloud.githubusercontent.com/assets/12678235/15551982/11b1747c-227e-11e6-91e4-ab1a3958f2f6.png)
![image](https://cloud.githubusercontent.com/assets/12678235/15552037/545678cc-227e-11e6-9fee-a972fadb198f.png)

PR #19, opened 2016-05-25
Fix issue #18, trash Try It solutions

tomjw64:
Correctly ordered solutions and removed the Try It solutions, which should not be present


reedstrm:
Here's the sort of description I would have wanted for review, if I hadn't worked through this with you in person:
This changes the pass for solution moving to a later pass, so they have already been sorted with their respective exercises at end of chapter. Try-it solutions are not included in the book, so moving to trash.

PR #22, opened 2016-05-26
Passtravaganza

tomjw64:
Replace pass numbers with variables. Variables and commented descriptions of processes that take place on each pass located in mixins.less. Passes were also re-ordered such that similar processes will take place on the same pass (titles all containers on pass 30, etc.). Does not affect output.


books/rulesets/common/collations.less
```diff
@@ -1,3 +1,5 @@
+@import url("common/mixins.less");
```
philschatz: Does this import work? I would have expected the line to be `@import './mixins';`

books/rulesets/common/mixins.less
```diff
@@ -0,0 +1,12 @@
+@5-sol-move: 5; /*Assign exercise solutions to migrate to their own container*/
```
philschatz: _Minor 💭 :_ Why include the pass number in the name of the variable? It's still easy to search-and-replace . What about something like:

```
@pass-numbering-10: 10;   // Number chapters and static sections
@pass-sol-receive-15: 15; // Actually create solution container and move solutions
@pass-numbering-20: 20;   // Number solutions/EOC content
@pass-index-25: 25;       // Create everything associated with the index
@pass-index-symbol-26: 26; // Create separate section in index for symbols
@pass-title-30: 30;       // Title existing EOC/EOB containers
@pass-toc-35: 35;         // Create everything associated with the toc
@pass-toc-nav-40: 40;     // Create toc nav
@pass-metadata-50: 50;    // Add metadata to all locations
@pass-meta-title-51: 51;  // Add title to metatdata sections
@pass-meta-uri-52: 52;    // 
```

Then, autocomplete would work in text-editors (like https://atom.io but I'm sure others as well)

philschatz:
Maybe `cte` could decide on some LESS conventions and add to `CONTRIBUTING.md? There's https://github.com/openstax/napkin-notes/blob/master/phil/conventions.md#lesscss-conventions :smile:

books/rulesets/common/mixins.less
helenemccarron: The pass number in the variable is useful in my opinion when reading the declarations and not having to remember in which order different passes are happening. I may not remember that the toc is happening after the index. 

For the auto complete, sounds like a good idea. I just would write it down the way I logically think about it, for example @pass-10-numbering

helenemccarron:
@philschatz: agreed, we are still very much in the process of defining those conventions but we can start following the guidelines and and updating our contributing

books/rulesets/output/statistics.css
```diff
@@ -1,3 +1,15 @@
...
+/*Actually create solution container and move solutions*/
+/*Number solutions/EOC content*/
+/*Create everything associated with the index*/
+/*Create separate section in index for symbols*/
+/*Title existing EOC/EOB containers*/
+/*Create everything associated with the toc*/
+/*Create toc nav*/
+/*Add metadata to all locations*/
+/*Add title to metatdata sections*/
+/**/
```
reedstrm: if you use // comments, they don't pass through to the css.

tomjw64: Whoops... Good to know.

helenemccarron: That said, that particular file will not make it to the css anyway since it's all less :)

PR #23, opened 2016-05-26
Numbering

helenemccarron:


tomjw64:
Try It solutions are not yet numbered, but should be numbered according to the example that precedes them

mjbuchholz:
Example problems are numbered but the Solution contained within every Example problem still need to be numbered(with the same numbering).

PR #31, opened 2016-06-08
target-counters

helenemccarron:


philschatz:
yay, nice!

reedstrm:
@phil it does target-string, too :-)

PR #41, opened 2016-08-12
Refactor

helenemccarron:


books/rulesets/common-refactor/_obj-composite-page.scss
```diff
@@ -0,0 +1,37 @@
...
+  @return map-merge($obj_compositePage, $params);
+}
+
+$obj_chapterCompositePage: obj_compositePage((
+  hasSolutions: false,
+  isGlossary: false,
+  sortBy: null,
+));
+
+@function obj_chapterComposite($params) {
```
philschatz: It seems like this is just a way to handle defaults. Is that right? If so, could we just skip this complication?

helenemccarron: Yes, that was already planned :) Thomas really liked objects and wanted to play with them but we recognized this was superfluous.

philschatz:
Ooh, thanks for removing the `.less` files! Is there a reason to include the `.map` and `.scssc` files? If not, I can add them to `.gitignore` 😄

PR #44, opened 2016-09-02
simplify instructions for fetching and baking books

philschatz:
This adds instructions to the README for getting set up, baking, and updating dependencies.

So, the instructions now look like:

```sh
./scripts/setup       # Only need to run this once
./scripts/fetch-book  statistics ${UUID}
./scripts/bake-book   statistics
./scripts/update      # If you want to update the dependencies
./scripts/generate-docs
```

## TODO

- [x] simplify the `.gitignore` file to only include files that are actually generated
- [ ] figure out if the `generated-html/` directory can be removed
- ~find a better way to list the python dependencies so there is less copy/pasta~
- ~get `ssh` and `scp` to run in 1 command (fewer passwords to type)~

/cc @edwoodward @Stackblocks because this removes 2 of the files you've edited before in `docs/` (hopefully simplifying the process ❤️ )

**Note:** The organization of the `scripts/` directory is "inspired" by http://githubengineering.com/scripts-to-rule-them-all/


edwoodward:
Looks good. :+1:

scripts/update
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+set -e
+
+echo 'Step 1/4: Updating SASS'
+
+gem update --quiet sass
+
+echo 'Step 2/4: Uninstalling python packages'
```
philschatz: @reedstrm: Is there a better way to list these dependencies to install/upgrade? Currently, they are copy/pasta'd in the `scripts/setup` and `scripts/update` files. /cc @pumazi

README.md
```diff
@@ -1,2 +1,21 @@
-# cte
-Placeholder repository for Textbook Experience project artifacts
+# Install
+
+1. run `./scripts/setup` to install the dependencies. You only have to run this once!
+
+**Note:** To update the dependencies, run `./scripts/update`
+
```
helenemccarron: maybe add something about activating the virtualenv?

README.md
```diff
@@ -1,2 +1,21 @@
...
+# Cook a New Book
+
+1. Find the UUID of the book you want
+  - point your browser to `cte-cnx-dev.cnx.org` and find a book you want.
+  - click "More Information" and copy the **ID**
+1. run `./scripts/fetch-book statistics ${ID}`
+  - **Note:** This will require you to log in via ssh. Also, make sure the **ID** contains the version.
```
helenemccarron: Maybe provide instructions + example, something along the lines of:

run: `./scripts/fetch-book [bookName] ${ID}`
for example: `./scripts/fetch-book statistics 69619d2b-68f0-44b0-b074-a9b2bf90b2c6@11.330`

README.md
```diff
@@ -1,2 +1,21 @@
-# cte
-Placeholder repository for Textbook Experience project artifacts
+# Install
+
+1. run `./scripts/setup` to install the dependencies. You only have to run this once!
+
+**Note:** To update the dependencies, run `./scripts/update`
+
+# Cook a New Book
+
```
helenemccarron: I would separate the instructions in two steps:
 "Generate the epub and HTML on the server then retrieve the HTML file", and "Bake the html using the ruleset" for clarity.

README.md
```diff
@@ -1,2 +1,21 @@
...
+1. Find the UUID of the book you want
+  - point your browser to `cte-cnx-dev.cnx.org` and find a book you want.
+  - click "More Information" and copy the **ID**
+1. run `./scripts/fetch-book statistics ${ID}`
+  - **Note:** This will require you to log in via ssh. Also, make sure the **ID** contains the version.
+1. run `./scripts/bake-book statistics`
```
helenemccarron: This assumes that the HTML file name is the same as the CSS but that's not always the case, some books use a common file (all the books in the Econ series are using economics.css for example).

philschatz: Could the scripts get updated to take an optional 3rd-argument when the need arises?

README.md
philschatz: I updated the README to describe that you can run the 2nd step multiple times, but this comment didn't disappear because I didn't edit this line (it's a blank line ; (

README.md
```diff
@@ -1,2 +1,22 @@
...
+**Note:** To update the dependencies, run `./scripts/update`
+
+# Cook a New Book
+
+1. Find the UUID of the book you want
+  - point your browser to `cte-cnx-dev.cnx.org` and find a book you want.
+  - click "More Information" and copy the **ID**
+1. run `./scripts/fetch-book statistics ${ID}`
+  - **Example:** `./scripts/fetch-book statistics 69619d2b-68f0-44b0-b074-a9b2bf90b2c6@11.330`
+  - **Note:** This will require you to log in via ssh. Also, make sure the **ID** contains the version.
```
philschatz: Add the 3 options:

- if your username is the same on the local machine then you just type in your password
- create a key (`ssh-keygen`)
- set an environment variable for the username

scripts/update
reedstrm: `pip -r requirements.txt` is the usual spelling.  Perhaps w/ `-U` to force upgrading. Is that the point of the uninstall/reinstall dance?

philschatz: Cool, I didn't know `requirements.txt` could [point directly to a git repo](https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source)!

Also, @pumazi mentioned running `git clone https://github.com/connexions/${REPO_NAME} && cd ./python-packages/ && pip install --quiet -e .` so the package could be upgradeable but if `-U` works then I wonder if there's any other advantage of doing the `git clone` manually

philschatz: hmm, running `pip install -U ./requirements.txt` resulted in:

```
Could not find a version that satisfies the requirement requirements.txt (from versions: )
No matching distribution found for requirements.txt
```

philschatz: I ended up using `requirements.txt` to list all the packages for initial setup, but during the update script I explicitly listed the packages to uninstall.

PR #46, opened 2016-09-06
Create an overview document for outside use

philschatz:
The more I wrote this the more it seemed like we could use SASS Doc to generate it (by writing up a documentation-only SASS file; thoughts?

It seems sassdoc does not generate documentation for a file that only contains selectors. So, I started using https://github.com/kss-node/kss-node

### DIFF
--- a/docs/html-overview.md
+++ b/docs/html-overview.md
@@ -1,0 +1,77 @@
+Each Page contains the following:
+
+- `div[data-type='page']` : Everything in here is specific to the page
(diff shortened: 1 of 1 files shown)


philschatz:
This work has been recreated in https://github.com/Connexions/cnx-recipes/compare/simplify-instructions...generate-guide but is waiting on #44 to be merged first

philschatz:
Closing in favor of #48

PR #47, opened 2016-09-07
Add chapter separated option to EOB solutions

helenemccarron:
End of book solutions can now be separated by chapter, by section or both. When separating by section, the End of Chapter section must be section separated as well to retrieve the appropriate nesting. 

Expected output when using both:
```
<div class="eob solutions-container" data-type="composite-page" id="ebc57378-4f0d-4609-a258-84067c34ba0b@1">
<h1 data-type="document-title"><span class="text">Answer Key</span></h1><div data-type="metadata">
<div class="solutions-chapter-area">
<h2 class="title">Chapter 1</h2>
<section class="solutions">
<h3 xmlns:c="http://cnx.rice.edu/cnxml" xmlns:md="http://cnx.rice.edu/mdml" xmlns:qml="http://cnx.rice.edu/qml/1.0" xmlns:mod="http://cnx.rice.edu/#moduleIds" data-type="document-title" id="auto_9ff3ce08-0471-4d79-81d7-6a57a308d142@12_30568_copy_1"><span class="number">1.4</span><span class="divider"> | </span><span class="text">How Economies Can Be Organized: An Overview of Economic Systems</span></h3>
<div data-type="solution" class="solution" id="auto_e911f49c-bb38-49cf-bb03-066373c43a76@8_fs-idp44451856">
<a id="solution-1" href="#problem-1"><span class="number">1</span></a>
<p id="auto_e911f49c-bb38-49cf-bb03-066373c43a76@8_fs-idp17723088">Scarcity means human wants for goods and services exceed the available supply. Supply is limited because resources are limited. Demand, however, is virtually unlimited. Whatever the supply, it seems human nature to want more. </p></div><div data-type="solution" class="solution" id="auto_e911f49c-bb38-49cf-bb03-066373c43a76@8_fs-idm49024864"><a id="solution-2" href="#problem-2"><span class="number">2</span></a><p id="auto_e911f49c-bb38-49cf-bb03-066373c43a76@8_fs-idm49858304">100 people / 10 people per ham = a maximum of 10 hams per month if all residents produce ham. Since consumption is limited by production, the maximum number of hams residents could consume per month is 10. </p>
</div>
</section>
</div>
</div>
```

### DIFF
--- a/books/rulesets/common/_compose.scss
+++ b/books/rulesets/common/_compose.scss
@@ -216,49 +216,64 @@
       }
     }
   }
+  @function map-deep-get($map, $keys...) {
+    @each $key in $keys {
+      $map: map-get($map, $key);
+    }
+    @return $map;
+  }
   @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
     $solutionSource: map-get($solutionPage, source);
+    //$isSeparated: map-has-key($solutionPage, separated);
+    $isSeparated:  map-get($solutionPage, isSeparated);
     $sectionSeparated: map-get($solutionPage, sectionSeparated);
-    $chapterSeparated: map-get($solutionPage, chapterSeparated);
-    @each $page in $compositePages {
-      $hasSolutions: map-get($page, hasSolutions);
-      $source: map-get($page, source);
-      @if ($hasSolutions) {
-        [data-type="chapter"] {
-          .eoc.#{$source}-container {
-            [data-type="solution"] {
+    $chapterSeparated: map-deep-get($solutionPage,chapterSeparated);
+
+   // $sectionSeparated: map-get($isSeparated, sectionSeparated);
+
+   @each $page in $compositePages {
+    $hasSolutions: map-get($page, hasSolutions);
+    $source: map-get($page, source);
+    @if ($hasSolutions) {
+      [data-type="chapter"] {
+        .eoc.#{$source}-container {
+          [data-type="solution"] {
+            @if ($isSeparated) {
               @if ($sectionSeparated) {
                 move-to: #{$solutionSource}-GETSECTION;
               }
               @if ($chapterSeparated and not $sectionSeparated ) {
                 move-to: #{$solutionSource}-GETCHAPTER;
               }
-              @else {
-                move-to: #{$solutionSource}-TOCOMPOSITE;;
+            }
+            @else {
+              move-to: #{$solutionSource}-TOCOMPOSITE;;
+            }
+          }
+          @if ($sectionSeparated) {
+           .#{$source} {
+            &::after {
+              container: section;
+              class: #{$solutionSource};
+              content: pending(EOBsectionHeader) pending(#{$solutionSource}-GETSECTION);
+              @if ($chapterSeparated) {
+                move-to: #{$solutionSource}-GETCHAPTER;
               }
             }
           }
         }
       }
     }
+  }
+}
+
+
+
+  @if ($chapterSeparated) {
     [data-type="chapter"] {
-     > h1[data-type="document-title"] .number {
-      string-set: ChapNum content();
-    }
-    @if ($sectionSeparated) {
-      &::after {
-        container: section;
-        class: #{$solutionSource};
-        content: nodes($sectionHeaderNode) pending(#{$solutionSource}-GETSECTION);
-        @if ($chapterSeparated) {
-          move-to: #{$solutionSource}-GETCHAPTER;
+       > h1[data-type="document-title"] .number {
+         string-set: ChapNum content();
         }
-        @else {
-          move-to: #{$solutionSource}-TOCOMPOSITE;
-        }
-      }
-    }
-    @if ($chapterSeparated) {
       &::after {
           //FIXME wrap the in spans
           container: h2;
--- a/books/rulesets/common/_compose.scss
+++ b/books/rulesets/common/_compose.scss
@@ -224,13 +224,9 @@
   }
   @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
     $solutionSource: map-get($solutionPage, source);
-    //$isSeparated: map-has-key($solutionPage, separated);
     $isSeparated:  map-get($solutionPage, isSeparated);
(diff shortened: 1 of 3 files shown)


books/rulesets/common/_compose.scss
```diff
@@ -191,116 +191,94 @@
...
-          }
-          @if ($sectionSeparated) {
-            &::after {
-              container: section;
-              class: #{$solutionSource};
-              content: nodes($sectionHeaderNode) pending(#{$solutionSource}-GETSECTION);
-              move-to: #{$solutionSource}-TOCOMPOSITE;
+  @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
+    $solutionSource: map-get($solutionPage, source);
+    $sectionSeparated: map-get($solutionPage, sectionSeparated);
```
philschatz: Is there a way in SASS to have `$solutionPage` store an enum (ie `separated = section` or `separated = chapter`) rather than having what appear to be 2 boolean flags? Someone _could_ set `sectionSeparated` _and_ `chapterSeparated` which would cause both `@if` statements to be true but I do not think there is a valid reason for that happening, no?

philschatz:
Can you include a tiny HTML snippet in the Pull Request description that illustrates the changes? It would be helpful in reviewing to see the changes 😄

books/rulesets/common/_compose.scss
helenemccarron: There _is_ support for both to be enabled or not  in the current mixin:
```@if ($chapterSeparated and not $sectionSeparated ) {
                move-to: #{$solutionSource}-GETCHAPTER;
              }
```
and 
```
    @if ($sectionSeparated) {
...
        @if ($chapterSeparated) {
          move-to: #{$solutionSource}-GETCHAPTER;
        }
```

However, looking at it, I think there is a bug in the firing order when using both.

philschatz: I guess I was wondering if there were any books (or use cases) that would need to have both `sectionSeparated=true` and `chapterSeparated=true`. In that case, there would be 2 `move-to` declarations? 

If that happens, then it is unclear to me where the element would be moved. But maybe I'm misunderstanding?

helenemccarron: Ah, indeed, I missed that part :) Thanks for punching holes in my logic! I'm looking into it.

helenemccarron: My updated code should handle both situations better. It's not perfect. As far as I could tell, one way of achieving the same result would be with nested maps but it turned out to be a cumbersome and heavily repetitive solution. I have settled for something a bit less elegant but much cleaner. Maybe I'm missing something obvious but I couldn't find anything else in SASS that would resolve this particular issue.

helenemccarron:
I have added the cooked HTML applying both EOB  chapter separated sections and section separated section. This currently doesn't exist anywhere in Statistics or Econ and is purely for testing purposes.

philschatz:
Thanks! I was thinking of something like (raw HTML):

```
<div data-type="chapter">
  <div data-type="page">
    NON_COLLATED_PAGE_CONTENT
    <section class="self-check-questions">
      <h3 data-type="title">Self-Check Questions</h3>
      <div data-type="exercise">
        <div data-type="problem">
          <p>QUESTION?</p>
        </div>
        <div data-type="solution">THE_SOLUTION</div>
      </div>
    </section>
    <!-- Maybe add other collated sections here -->
  </div>
</div>
```

I'll re-review in an hour or 2

reedstrm:
we all can't type pseudo-html as fast as you can, phil. :)

books/rulesets/common/_compose.scss
```diff
@@ -111,59 +111,59 @@
...
-        }
-        @if ($chapterSeparated) {
-          &::after {
-            class: "#{$source}-chapter-area";
-            content: pending(#{$source}-GETCHAPTER);
-            @if ($chapterPages) {
-              data-type: "composite-page";
+            @if ($sectionSeparated) {
+              content: nodes(sectionHeaderNode) content();
+              @if ($chapterSeparated) {
```
philschatz: Is this dead code? When would both `$sectionSeparated` and `$chapterSeparated` be true?

books/rulesets/common/_compose.scss
```diff
@@ -191,116 +191,110 @@
...
+              node-set: eobsectionHeader;
+              move-to: EOBsectionHeader;
+            }
             &::after {
               container: section;
               class: #{$solutionSource};
-              content: nodes($sectionHeaderNode) pending(#{$solutionSource}-GETSECTION);
-              move-to: #{$solutionSource}-TOCOMPOSITE;
+              content: pending(EOBsectionHeader) pending(#{$solutionSource}-GETSECTION);
+              @if ($chapterSeparated) {
```
philschatz: Is this also dead code? When would `$sectionSeparated` and `$chapterSeparated` both be true? (this line is inside a `@if ($sectionSeparated)` block)

helenemccarron: Maybe I am mistaken but wasn't that exactly what you were asking in a previous comment, what if the end of book material is both section separated and chapter separated?

philschatz: Ah, what I meant with the question was "Do we need to handle the case of both `sectionSeparated:true` and `chapterSeparated:true`?" I thought the answer would be "no" (hence #51) and I hadn't heard anything about the content team wanting that feature. If it works then I feel like there should be code that "exercises" that feature so it does not bitrot or confuse people when they eventually dod try to set both to be true

philschatz:
I think now the only thing left on this PR is indentation of the CSS. Here are the lines I found:

image

image

Before this PR there were no indentation errors in that file so I _think_ they are all valid; thoughts?

books/rulesets/common/_compose.scss
helenemccarron: I agree but the philosophy of CTE has been to be as inclusive as possible, this includes the rulesets. We added many features that aren't implemented in the two books we currently have a ruleset for but we know would be needed for other books as well as the possibility for functionalities that do not currently exist but are possible and should be supported.

helenemccarron:
Bah, auto reindent in Sublime Text is not as neat as it looked :( If you have time, it would be a greatly appreciated.

PR #48, opened 2016-09-09
Generate HTML guide for books

philschatz:
This adds a new script `./scripts/generate-guide ${BOOK_NAME}` which uses KSS to create a set of HTML files in `/styleguide/` that show cooked snippets of HTML and are specific to the rulesets of a book.

The source for the documentation is in CSS comments and is split into `common` and book-specific.

Also, may depend on https://github.com/Connexions/cnx-easybake/pull/32 being merged first.

# Directory Organizational Map

- `rulesets/common/examples/*.html` : The Raw HTML snippets for "common" elements in a Page
- `rulesets/common/examples/_all.scss`: The documentation for "common" elmeents in a Page
- `rulesets/examples/*.html` : The Raw HTML snippets for book-specific collated pages and any other customizations
- `rulesets/${BOOK_NAME}.scss`: Added CSS docs that are slurped in when generating the guide

# Book-Agnostic Documentation

The common documentation is in `rulesets/common/examples/_all.scss` only because I didn't know of a better place to put them but I wanted them "near" the common code. That directory also contains the Raw HTML snippets

# Book-Specific Documentation

The book-specific documentation is done as comments in the `${BOOK_NAME}.scss` and `${BOOK_NAME}-config.scss` files and the Raw HTML snippets are in `rulesets/examples/` because each book does not have a directory (yet?).

# TODO

- [x] Move the PR description into a `README.md` somewhere inside `/books/rulesets/`

# Deferred to another PR

- Simplify the names of the cooked HTML files so writing CSS comments that link to the files is easier
- render the raw HTML in the guide
- generate a separate directory for the guide for every book


books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+
+  Style guide: page.equation
+*/
+[data-type='equation'] {}
+
+
+/*
+  Glossary
+
+  This element occurs as a separate Page at the end of a Chapter or a Book. It is collected from individual glossary elements in each raw Page.
```
helenemccarron: It is often referred to as Key Terms.

books/rulesets/_statistics-config.scss
```diff
@@ -212,3 +221,13 @@ $targetLabels: (
...
+
+/*
+  Chapter Objectives
+
+  This is a Note that appears in the Introduction page of every chapter.
+
+  Markup: ./examples/page.note.chapter-objectives.html-cooked.html
+
+  Style guide: page.note.chapter-objectives
+*/
```
helenemccarron: Note for later, there are various levels of objectives, historically, we had some in the metadata but we moved to creating them in the content using a note. The ruleset should always handle those and not the metadata and they should be in the content, not the in the metadata or as an automatic field in webview.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+    Style guide: page.0-processing
+  */
+  > cnx-pi { display: none; }
+
+}
+
+/*
+  Term
+
+  A term which is defined in a glossary
```
helenemccarron: We refer to that as the Index - there's a difference between the glossary, which is typically several terms gathered within a block with processing instructions (collated material) gathered at the end of chapter or at the end of book, key terms are spread throughout the book and listed at the end of book.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+  - optional `[data-type='title']`
+  - `figure` element
+    - `[data-type='media']` : a wrapper element which may disappear in the future
+      - `img[src][alt]` : The image for the Figure
+  - `.caption-container` : contains the figure number and caption (may be renamed to figcaption in the future and moved into figure)
+    - `.title-label` : The word `Figure`
+    - `.number` : The number of the figure (ie `3.1`)
+    - `.caption`
+  - `figcaption`
+
```
helenemccarron: The caption for the figure and the table are the same format. It follows the 4 spans concepts but uses a special .caption span instead of a .text span. We could use exactly the same structure instead of using a special caption class.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+      > img[src] {}
+    }
+  }
+  > .caption-container {}
+}
+
+
+/*
+  Paragraph
+
```
helenemccarron: corner case: the issue of a para having a title, which is not proper HTML and how that is handled

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+  Style guide: page.example
+*/
+[data-type='example'] {}
+
+
+/*
+  Exercise
+
+  An exercise contains a problem, optional title, and optional solution(s).
```
helenemccarron: are usually numbered, their title need to contain the same span structure as titles

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+
+  Markup: ./page.exercise.html-cooked.html
+
+  Style guide: page.exercise
```
helenemccarron: Exercises are numbered, contain a link to the solution if it exists.
Solutions are collated (at the end of section, end of chapter or end of book), they are numbered (the number is the same as the exercise they are associated to) and link back to their exercise.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+  Style guide: book.composite
+*/
+[data-type='composite-page'] {}
+
+
+/*
+  Chapter
+
+  A Chapter contains a title and multiple Pages, one of which may be a `.introduction` Page.
+
```
helenemccarron: The introduction page typically contains a splash image, a chapter outline and an introductory paragraph.

books/rulesets/common/examples/page.equation.html
```diff
@@ -0,0 +1,3 @@
+<div data-type="equation">
+  <math><mn>2</mn><mo>*</mo><mn>2</mn><mo>=</mo><mn>4</mn></math>
+</div>
```
helenemccarron: Equations often need to be numbered.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+*/
+[data-type='exercise'] {
+  > [data-type='problem'] { }
+  > [data-type='solution'] { }
+}
+
+/*
+  Equation
+
+  An equation wraps a MathML formula so it can be referenced by a number.
```
helenemccarron: Equations often need to be numbered.

books/rulesets/common/examples/_all.scss
```diff
@@ -0,0 +1,258 @@
...
+
+  Style guide: page.table
+*/
+> table {
+  caption {}
+}
+
+/*
+  Composite Page
+
```
helenemccarron: End of book/End of chapter pages: 
They must use `data-type="composite-page"`. They must have a unique identifier (class) for styling purposes. 
They usually contain content separated by section(pages) or by chapter.
They now appear in the TOC (currently not the case in the PDF).
They aren't numbered in the TOC (the chapters and sections are)
They follow the same rules as a regular `data-type="page"` (header hierarchy, metadata, etc.)

books/rulesets/_statistics-config.scss
philschatz: Has this been decided? The reason for putting the objectives in the metadata was that if a teacher removed one of the sections from their custom book, objectives that relate to that removed module would also be removed in their book (similar to how exercises just "disappear" if you remove one of the modules).

books/rulesets/common/examples/_all.scss
philschatz: Hmm, maybe there should be a separate section for Intro pages? Are they book-specific? I added your notes in here so they don't get lost though 😄

books/rulesets/common/examples/_all.scss
philschatz: I tried to create an exercise (in the example HTML) which generated the numbering, no? Ah, maybe we could chat Monday about how to write up the "common" docs so they apply to all the books but describe the elements that are "typically" found in them when cooked

books/rulesets/common/examples/_all.scss
philschatz: I included these comments in the file but the original comment was on a blank line so GitHub did not recognize it was addressed.

philschatz:
I believe the concerns that were brought up should be resolved.

Also, this repository is now checked by Travis! (at least the guide for statistics)

philschatz:
To get this Pull Request to work either:

- https://github.com/Connexions/cnx-easybake/pull/32 needs to be merged first (cnx-easybake is choking on CSS comments)
- documentation comments need to use `//` instead of `/*`

books/rulesets/_statistics-config.scss
helenemccarron: This has gone back and forth as part of our team, using the metadata doesn't give us enough flexibility and isn't enough for styling or additional classes, especially for Tutor books. Several books already moved away from using it. Cc'ing @InconceivableVizzini and @openstaxalina

books/rulesets/common/examples/_all.scss
helenemccarron: Yes, maybe a separate section would be useful. The splash image is common but sometimes the title for the intro chapter is overwritten, sometimes not so there are differences between books that are worth listing.

books/rulesets/common/examples/_all.scss
helenemccarron: I'll make sure I don't comment on empty lines :P

books/rulesets/_statistics-config.scss
philschatz: but, wouldn't easybake fix the "not enough flexibility" concern? But I agree the documentation should reflect how each book currently works 😄

helenemccarron:
Would it make sense to rename the example folder to guide or something of that nature? The HTML data within the folder structure is the raw version and not representative of the ruleset itself outside of the html version of the guide.

philschatz:
Yep, I have no big preference for `./example/` or `./guide/` . Can it be changed in a separate PR because #50 also uses `/example/`

helenemccarron:
Agreed!

PR #49, opened 2016-09-09
capture a few debugging tips

reedstrm:
first cut. Place to land solutions to problems we hit, perhaps as staging for incorp. into other docs.


philschatz:
Thanks for adding! I made each bullet linkable (via URL) and made the commandline options and code snippets more distinguishable from the reset of the descriptions.

PR #50, opened 2016-09-10
Guide: Show raw and cooked HTML in the guide

philschatz:
This builds on #48 and adds the ability to see the Raw HTML and the Cooked HTML, and gives the user a dropdown to pick which format they want to look at. Eventually, this could also include the source CNXML.

Since this "piggy-backs" on #48 the list of changes in the PR also contains those commits. To see **only** these changes, **[click here](https://github.com/Connexions/cnx-recipes/compare/993cedc...use-custom-kss)**

# Screenshots

image

# How it Works

- `./scripts/generate-guide` combines the raw and cooked HTML into 1 HTML file named `*-cooked.html` for each snippet

- There is a custom kss-generator in `/js/openstax-custom-builder` which contains:
  - an `index.html` template file (handlebars):
    - it has a custom `<style>` tag (I did it this way because laziness)
    - it has a `<select>` dropdown button for choosing the format
  - a `kss.js` file:
    - parses the "annotated" output HTML snippets from the `generate-guide` (these contain both the Raw and Cooked HTML markup)
    - removes the extra `<html><body>` tags which are created by `cnx-easybake`
    - listens to changes in the `<select>` dropdown and hides/shows the formats

# TODO

- [x] style the example code blocks better
- ~maybe pull in the source CNXML instead of Raw HTML~
- [x] move the "How it Works" into a README


books/rulesets/common/examples/page.table.html
```diff
@@ -1,10 +1,15 @@
...
-<div data-type="page">
+  <div data-type="page">
 
-<table>
-  <caption>and a caption</caption>
-  <tr><td>A table cell</td></tr>
-</table>
+    <!-- Markup starts here -->
+    <table>
+      <caption>and a caption</caption>
```
helenemccarron: Similarly to figures, tables may also contain `<div data-type="title" class="title">content</div>`

books/rulesets/common/examples/page.term.html
```diff
@@ -1,2 +1,9 @@
-<p>The <span data-type="term">velocity</span> of an object
-   represents the speed and the direction</p>
+<div data-type="chapter">
```
helenemccarron: This applies specifically to the end of book Index. I am not sure how to make that distinction clear but key terms/glossary/index can be confusing, I feel we should try our best to define each concepts as clearly as possible

books/rulesets/examples/book.composite.critical-thinking.html
```diff
@@ -0,0 +1,38 @@
+<div data-type="chapter">
```
helenemccarron: In the current file structure, there is no way of knowing which book this applies to. These are what we commonly refer to as "collated elements" and are extremely books specific, even if in some cases, one collated element may be used in more than one book. They are always only relative to their own book, however, as far as the ruleset is concerned.

books/rulesets/examples/book.composite.references.html
```diff
@@ -0,0 +1,28 @@
+<div data-type="chapter">
```
helenemccarron: These have a slightly unique nature. Even if they are treated the same way as other collated element are sharing an identical structure and name regardless as a book. They need to be defined at the book level but should be consistent for all books.

books/rulesets/guide-homepage.md
```diff
@@ -0,0 +1,29 @@
...
+
+- Use `data-*` whenever possible. This applies to anything we use internally.
+- Classes are used when several attributes are needed on one element: `class="eoc practice-container"` `data-type="composite-page"`.
+- No numbering or titling should be accomplished via display css, it needs to be present in the unstyled cooked HTML
+- All book specific strings are handled via the ruleset, this includes titles for notes/features and “generic” strings like “Chapter” or “Figure”. Instead of using labels in the content, we create all the titles via the ruleset, this provides one location to look for errors if titles are incorrect and prevents localised error when creating content. This is needed for internationalization. The ruleset handles them via variables, none of those should be hard coded.
+- Address the risk of CSS input by author interacting with our base CSS. `data-*` vs `.class`. Need for class prefix? `.cnx-eoc `
+    **Questions:** how do we handle our internal markup
```
helenemccarron: Could we add a question about table/figure caption and a final definition of how they should be handled? (ie using captions/figcaptions, divs, how they are handled etc). 
Idem for lists, question: can we establish which route we would like to go, either ul or divs?
This may not be the place to add it but I'd like to record it for future reference1

README.md
```diff
@@ -19,5 +19,13 @@ There are 2 major parts to cooking a book (_listed above_). You will first need
 
 # Documentation
 
-1. run `./scripts/generate-docs`
-2. open the generated file in your browser
+1. run `./scripts/generate-docs` to generate the SASS Docs
+1. run `./scripts/generate-guide statistics` to generate the HTML Guide for a book
+1. open the generated file in your browser
+
+To update the Documentation in the gh-pages branch:
```
philschatz: Yep, this script is a little dangerous and not well tested yet; I could remove it from the `README.md` for now. It:

1. switches to the `gh-pages` branch
1. pulls master in (via `git rebase master`)
1. runs `./scripts/generate-docs` and `./scripts/generate-guide ${BOOK_NAME}` to create the HTML pages
1. checks in the HTML pages and pushes them into github so they are available at https://connexions.github.io/cnx-recipes/styleguide/

README.md
```diff
@@ -19,5 +19,13 @@ There are 2 major parts to cooking a book (_listed above_). You will first need
...
+
+1. `git checkout master`
+1. `./scripts/generate-and-update-gh-pages`
+
+**Note:** This will push changes if successful
```
philschatz: They would become available at https://connexions.github.io/cnx-recipes/styleguide/ . It's the link at the top of the repo:

image

books/rulesets/common/examples/page.exercise.html
```diff
@@ -1,10 +1,10 @@
 <div class="eoc">
```
philschatz: The cooked HTML is visible at https://connexions.github.io/cnx-recipes/styleguide/section-page.html#kssref-page-exercise

I added the `eoc` wrapper element because I noticed exercises were only being changed (numbered) if they occurred inside a `.eoc` element

books/rulesets/common/examples/page.table.html
philschatz: I don't recall seeing any examples of what that markup looked like; could you paste an example? (or just push to this branch 😄 )

books/rulesets/common/examples/page.term.html
philschatz: Sure! Any suggestions? And could they be in a separate PR(s) since I think we'll be editing these files as a group?

README.md
helenemccarron: Understood, should we leave it off for now until thoroughly tested?

README.md
helenemccarron: Got it, knowing what it does help, see above :)

books/rulesets/examples/book.composite.critical-thinking.html
philschatz: Yep, the guide _should_ only collect collated elements that are specific to the book. There's a "bug/feature" in kss that will require me to put each book into a separate directory (so `/rulesets/${BOOK_NAME}/book.scss` instead of the current `/rulesets/${BOOK_NAME}.scss`

books/rulesets/common/examples/page.exercise.html
helenemccarron: That is part of the cooking process but technically, they could be at the end of a section or at the end of book. They also can be in an example.

books/rulesets/examples/book.composite.references.html
philschatz: I only saw them in statistics (not economics). Are you suggesting changing the comment? If so, feel free to push updates to this PR

README.md
philschatz: 💭  It's difficult to thoroughly test if the code is not available. What if I put it in an "Experimental" section in the `README.md`? That way the code and instructions are available for testing but it's clear that the instructions are "use at your own risk"? Otherwise I can just remove it.

books/rulesets/guide-homepage.md
philschatz: ok. I updated the documentation for `table` and `figure` to have TODO items  that we can go over.

books/rulesets/common/examples/page.exercise.html
philschatz: ok, makes sense. Since none of the books so far seem to have exercises inside examples, I addressed it by removing the `.eoc` container and replaced it with a chapter and page container.

books/rulesets/common/examples/page.term.html
philschatz: I _think_ I tried to clarify the description a bit, noting that the Glossary does not use this `term`

books/rulesets/examples/book.composite.critical-thinking.html
philschatz: Would it be ok to fix this bug in a separate PR? That way the scope of moving files around is limited to that change?

README.md
helenemccarron: I like the idea of "use at your own risk", as long as it clearly labeled as such. 
💭 Big warning sings right by a little shark-infested pond visions 💭

books/rulesets/common/examples/page.table.html
helenemccarron: This is in Economics:
```
<caption>
<span data-type="title">Economic Freedoms, 2015</span>
(Source: The Heritage Foundation, 2015 Index of Economic
 Freedom, Country Rankings, http://www.heritage.org/index/ranking)
</caption>
```
The output is the same as a figure.

books/rulesets/common/examples/page.term.html
helenemccarron: That works. However, based on issues we are currently experiencing in OER.exports, it might not be enough for some of our books, there might be a need for an automated glossary. Can we add a TODO just so we don't forget and look into it later?

books/rulesets/guide-homepage.md
```diff
@@ -0,0 +1,29 @@
+# Output HTML conventions
+
+The purpose of this part is to document and centralize the output HTML decisions and conventions. These are the cooked formats that are delivered to the cnx.org website, the PDF generation system, and external consumers like TEA.
+
+- Use `data-*` whenever possible. This applies to anything we use internally.
+- Classes are used when several attributes are needed on one element: `class="eoc practice-container"` `data-type="composite-page"`.
+- No numbering or titling should be accomplished via display css, it needs to be present in the unstyled cooked HTML
+- All book specific strings are handled via the ruleset, this includes titles for notes/features and “generic” strings like “Chapter” or “Figure”. Instead of using labels in the content, we create all the titles via the ruleset, this provides one location to look for errors if titles are incorrect and prevents localised error when creating content. This is needed for internationalization. The ruleset handles them via variables, none of those should be hard coded.
+- Address the risk of CSS input by author interacting with our base CSS. `data-*` vs `.class`. Need for class prefix? `.cnx-eoc `
```
philschatz: hmm, I think I addressed this by changing the line to describing what you mentioned

books/rulesets/common/examples/page.table.html
philschatz: Hmm, I was wondering what the Raw HTML for the table looks like when it contains a title and caption. But I did add a `TODO:` to the HTML documentation noting that we need to decide what figures and tables look like

books/rulesets/common/examples/page.term.html
philschatz: Yep, comment added.

books/rulesets/common/examples/page.table.html
helenemccarron: I didn't paste the entire table here but the snippet above is the raw caption inside a table.

books/rulesets/examples/book.composite.references.html
philschatz: Yep, I think this could be a separate PR (which goes into more detail about book-specific features/documentation)

PR #51, opened 2016-09-14
Refactor EOB separators (section or chapter) to be an enum

philschatz:
(**Note:** This is a PR to #47 rather than `master` /cc @helenemccarron )

... instead of 2 booleans (`sectionSeparated` and `chapterSeparated`). Valid values are: `'SECTION'`, `'CHAPTER'`, and nothing (`null`?)

Now, in the config file you would write:

```sass
$pageSummary: (
  name: "Key Concepts and Summary",
  source: "summary",
  separateOn: 'SECTION',
);
```

instead of the more error-prone:

```sass
$pageSummary: (
  name: "Key Concepts and Summary",
  source: "summary",
  sectionSeparated: true,
  chapterSeparated: false
);
```

It also:

- [x] removes some dead code
- [x] fixes the indentation level in a few places (because I was confused as to what was going on) 
  - Maybe we could use [scss-lint](https://github.com/brigade/scss-lint) in the future
- [x] now uses the `@error "The Error Message";` in several places when `separateOn` is set to something other than `'SECTION'` or `'CHAPTER'`

# Notes

Since this is a refactor, the generated CSS files are bit-for-bit unchanged.

### DIFF
--- a/books/rulesets/_economics-config.scss
+++ b/books/rulesets/_economics-config.scss
@@ -57,7 +57,6 @@ $pageIndex: (
 $pageSolutions: (
   name: "Answer Key",
   source: "solutions",
-  isSeparated: true,
   chapterSeparated: true,
   compoundComposite: true,
   isAnswerKey: true,
--- a/books/rulesets/_economics-config.scss
+++ b/books/rulesets/_economics-config.scss
@@ -16,7 +16,7 @@ $pageKeyTerms: (
 $pageSummary: (
   name: "Key Concepts and Summary",
   source: "summary",
-  sectionSeparated: true
+  separateOn: 'SECTION',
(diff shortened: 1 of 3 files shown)


helenemccarron:
I think this would cause issues if the End of Book was both section separated and chapter separated, I have tested extensively using a map and nested maps to achieve both possibilities in the EOB and it wasn't very efficient. There might be another way following the path you are trying to use but I am currently running out of time trying to wrap up phase one. Is that something we can revisit when time allows?

philschatz:
sure, we can revisit next sprint 😄 . I created the PR because I didn't see any elements that were using both `sectionSeparated:true` and `chapterSeparated:true`

PR #52, opened 2016-09-14
adds sass-lint

philschatz:
To run it: `./scripts/test`

Currently, everything is marked as a warning but once the indentations are fixed these would be changed into errors.

# Screenshot

image

# TODO (future PR)

- clean up the current linting errors
- move the `:match("...")` SASS into a separate file so it can be ignored (because sass-lint does not parse the file)
- [x] add the linting to `./scripts/test` so Travis runs it (if #48 is merged first)

/cc @helenemccarron

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -1,0 +1,51 @@
+# Linter Options
+options:
+  # Don't merge default rules
+  merge-default-rules: false
+  # Set the formatter to 'stylish' From https://github.com/eslint/eslint/tree/master/lib/formatters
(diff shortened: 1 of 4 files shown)


philschatz:
If #50 is merged first, I was planning on adding it to https://github.com/Connexions/cnx-recipes/pull/50/files#diff-bd0b996d66f03a10364809b532be8f0f . That way: 

1. Travis runs it
2. you can run all the tests via `./scripts/test` (and that would be in the README)

PR #53, opened 2016-09-14
adds a CSS coverage report

philschatz:
The report can be seen here: https://connexions.github.io/cnx-recipes/coverage/

This uses the Raw HTML for the books plus the CSS files for each book and generates a report showing which selectors were unused (could be removed from the CSS) as well as the ones that were used (and how often they matched)

# Screenshot

image

**Note:** the coverage correctly handles LESS files but not SASS so I pointed the coverage tool to the `rulesets/output/${BOOK_NAME}.css` file instead of the `rulesets/${BOOK_NAME}.scss` file

# TODO

- [ ] plug it into https://coveralls.io by running all the example HTML snippets through `css-coverage`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@
 /books/data/
 /data/*-raw.html
 /data/*-cooked.html
+# CSS Code coverage
+/data/*.lcov
 
 # SASS stuff
 /.sass-cache/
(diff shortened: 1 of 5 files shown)


helenemccarron:
Is there a reason why this PR doesn't pass Travis tests?

philschatz:
ah yep. I pushed a fix 😄

PR #54, opened 2016-09-14
add in some documentation (and validation) for mixins

philschatz:
I started adding documentation for the mixins and thought to add type-validation for the arguments. Is that something useful? or do you think the builtin sass errors are enough? They would not catch cases like:

- developer thinks they should pass in a list but the actual mixin that uses the arg expected it to be a bool
- developer thinks they should pass in a bool but the actual mixin that uses the arg expected it to be a string

If it is not useful then I will just stick with writing sass docs.

/cc @helenemccarron 

# View the Rendered Overview of Types: [README.md](https://github.com/Connexions/cnx-recipes/blob/add-mixin-docs/books/rulesets/README.md#mixin-types)

# TODO (in another PR)

- rename vars so they match the convention listed in this PR (ie use `isSectionSeparated`)
  - change `source` to `sourceType` and use the `[data-type="..."]` selector
- combine several of the mixins into 1 (ie `number_*` and `count_*`)
  - basically anything that is grouped together in the scss doc by virtue of lack-of-vertical-whitespace
- move the `:pass(#)` prefix into the mixin so people do not have to remember when it runs
  - This will require creating an enum of pass names that are used instead of just a number (for search-and-replace and sanity)

### DIFF
--- a/books/rulesets/README.md
+++ b/books/rulesets/README.md
@@ -47,6 +47,41 @@ These are the Types used to configure the numbering an collation of a book.
 - `compoundComposite` - bool
 - `isAnswerKey` - bool
 
+These are collected into a list as `$chapterCompositePages` and `$bookCompositePages` in the config file.
+
+
+### Mixins
+
+```sass
+@include reference_refSectionHeaderNodeAs(sectionHeaderNode);
+@include compose_createChapterComposites($chapterCompositePages, sectionHeaderNode);
+
+//After: ???
+@include reference_refSectionHeaderNodeAs(sectionHeaderNode);
+@include reference_refSectionHeaderStringAs(sectionHeaderString);
+@include compose_prepBookComposites($bookCompositePages, sectionHeaderNode, sectionHeaderString);
+
+//Only move solutions after exercises/solutions are numbered
+@include reference_refSectionHeaderNodeAs(sectionHeaderNode);
+@include compose_createEOBSolutions($chapterCompositePages, $pageSolutions, sectionHeaderNode);
+
+//After: prepBookComposites
+@include reference_refChapterHeaderNodeAs(chapterHeaderNode);
+@include compose_prepChapterAreas($bookCompositePages, chapterHeaderNode);
+
+//After: prepBookComposites, prepChapterAreas
+@include compose_createBookComposites($bookCompositePages);
+
+//After: createChapterComposites, createEOCSolutions, createBookComposites
+@include compose_titleEOCComposites($chapterCompositePages);
+@include compose_titleEOBComposites($bookCompositePages);
+
+//After: createChapterComposites, createEOCSolutions, createBookComposites
+@include modify_compositeAutoID();
+@include modify_chapterAutoID();
+
+@include modify_retitleCompositeMetadata();
+```
 
 # TitleContent
 
@@ -56,20 +91,103 @@ Describes numbering and labeling of elements. Also known as FigNumber, TableNumb
 - `number` - set of counters
 - `divider` - string
 
+### Mixins
 
-# setFigure(Table)Caption
+```sass
+//Come up with mixins for common numbering schemes as per Phil's suggestion
+@include count_countChapters(chapter);
+@include number_numberChapters($chapterTitleContent);
+@include count_countAppendices(appendix);
+@include number_numberAppendices($appendixTitleContent);
+@include count_countSections(section);
+@include number_numberSections($sectionTitleContent);
+@include count_countExamples(example);
+@include number_numberExamples($exampleTitleContent, $exampleSolutionTitleContent);
+@include count_countNote(Try, "try");
+@include number_numberNote("try", $tryTitleContent);
+@include number_numberNote("calculator", $calcTitleContent);
+
+
+//After: createChapterComposites
+@include count_countEOCExercises(exercise);
+@include number_numberEOCExercises($exerciseTitleContent, $exerciseTitleContent);
+```
+
+# setFigure(Table)Caption(Number)
 
 - `type` token (can be `table` or `figure`)
 - `defaultContainer` token (can be `caption`)
 - `hasCaption` bool
 - `hasTitle` bool
 
+### Mixins
 
-# $targetLabels
+```sass
+@include count_countTables(table);
+@include count_countFigures(figure);
+@include number_setCaptions($setTableCaption, $captionTableNumber, $captionTableNumberAp);
+@include number_setCaptions($setFigureCaption, $captionFigNumber, $captionFigNumberAp);
+```
+
+# TargetLabel
 
 These are used when creating the text for a link (ie `See Figure 4.3`).
 
-## TargetLabel
-
 - `selector` string
 - `label` nodes? containing things like `"Figure" counter(chapter) "." counter(figure)`
+
+These are collected in `$targetLabels`.
+
+```sass
+//After: composite page creation
+@include count_countChapters(chapter);
+@include count_countAppendices(appendix);
+@include count_countEOCExercises(exercise);
+@include count_countExamples(example);
+@include link_setTargetLabels($targetLabels);
+@include link_setLinkLabels();
+```
+
+
+# ToC
+
+Generating to Table of Contents requires the following mixins (each done in a separate pass).
+
+```sass
+@include toc_prepTOCElements();
+@include toc_putTOCElements();
+@include toc_navTOCUnwrap();
+```
+
+
+# Misc mixins
+
+These are a grab-bag of the rest of mixins that are used. They should eventually be grouped into categories.
+
+```sass
+//Some modifications to be done to the book before any collation
+@include modify_titlePreface();
+@include modify_spanWrapTitles();
+@include modify_simLinkTarget();
+@include modify_trash('.try [data-type="solution"]');
+@include utils_hasSolution();
+
+@include count_countTerms(term);
+@include reference_refPageIDStringAs(pageID);
+@include modify_prepIndexTerms(term, pageID);
+
+//After: createChapterComposites, numberEOCExercises
+@include count_countExercises(exerciseAll);
+@include link_linkToProblemsFromSolutionsEOC("number", "number", exerciseAll);
+
+//After: prepIndexTerms, createBookComposites
+@include modify_addIndexSymbolGroup("index", $symbolsSectionTitle);
+
+@include reference_refBookMetadataNodeAs(bookMetadata);
+@include modify_compositeMetadata(bookMetadata);
+
+//After: All metadata tasks are completed
+@include modify_suppressURI();
+
+@include utils_clearTrash();
+```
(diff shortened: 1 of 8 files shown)


books/rulesets/common/_compose.scss
```diff
@@ -25,34 +39,57 @@
...
           }
         }
       }
       #{$sourceSelector} {
         > h3[data-type="title"] {
+          // TODO: Explain why is this discarded to the trash
           move-to: trash;
         }
         @if ($sectionSeparated) {
+          // TODO: Is this used? because it looks like there is a typo here. `sectionHeaderNode` is not defined
```
tomjw64: It should read **$**sectionHeaderNode and use the string variable with the node name passed into the mixin

tomjw64:
> developer thinks they should pass in a list but the actual mixin that uses the arg expected it to be a bool
> developer thinks they should pass in a bool but the actual mixin that uses the arg expected it to be a string

Why would these cases not be caught?

philschatz:
@tomjw64 thanks for taking a look!

Since many of the `@if` statements just check for "truthiness", if a developer thought `sectionSeparated` was a list of sections to separate (instead of a bool) they might unknowingly make it true.

---

I liked that you tried to include the name of the type (like Page or TitleContent) in the variable name (that may also reduce the need for type-check code), any thoughts about the following renames?

1. from `sectionSeparated` to `isSectionSeparated`
1. from `compoundComposite` to `isCompoundComposite`
1. from `title-label` to `titleLabel` (to match camelCasing)
1. from `groupBy` to `groupBySel`
1. from `source` to `sourceSel` (or `sourceSelector`) 
  - and change the code so it does `#{$sourceSel}` instead of `@if $source == 'glossary' then "[data-type='glossary']" else ".#{$source}"`
1. from `label` to `labelNodes` (to clarify more than a string can be passed in)

Also, it's still a little confusing to me when reading `statistics.scss` to know which vars were defined in locally in the file and which were defined in `statistics-config.scss`; any thoughts on somehow casing the ones defined in `statistics-config.scss` differently?

Anywho, thanks for hopping in!

books/rulesets/common/_compose.scss
philschatz: Yep, while reviewing I noticed that variable was not being used ( sass-lint should check for unused variables _soon_ https://github.com/sasstools/sass-lint/issues/721 ) and was going to go back and replace `sectionHeaderNode` below with `$sectionHeaderNode`. I just wanted to make sure folks saw it 😄

tomjw64:
Ah! A misunderstanding. I assumed you were talking about the changes in this PR (and not the original code) when you said that those cases would not be caught! So I guess we agree that the code in this PR will catch those cases 😃 

To respond to your suggestions: I'll use some of the pre-edit suggestions because it covers a little more
> from title-label to titleLabel (to match camelCasing)

I would not make that change because title-label ends up being used as a string and set as a class/data-type name in the book's HTML, where camel case is more frowned upon, but I assume you found this out and that's why you removed this one. Maybe there could be some documentation/code change to make that more evident.
> prefixing fields like `sectionSeparated` and `compoundComposite` with `is`

Heck yeah, 11/10 that's more specific about what type they need to be
> suffixing fields like `groupBy` or `source` with `Sel` and change the code so it does `#{$sourceSel}` instead of `@if $source == 'glossary' then "[data-type='glossary']" else ".#{$source}"`

I think adding the suffix to 'groupBy' makes sense, but makes less sense with 'source'. Since glossary sections have 'data-type=$source' and all other sections use 'class=$source', that change would involve writing out a ton of long, but really similar selectors to define pages when they really shouldn't have to be that long just to account for the glossary. I think it would be a lot nicer if glossaries section containers were just labeled and selected like every other section, but that's not my choice to make 😆 
>from label to labelNodes (to clarify more than a string can be passed in)

Not the biggest fan of this either, because the label (I assume this is target-labels we are talking about here?) does not necessary need to contain nodes. For example, iirc links to Try It problems in Statistics simply appear with the label of "Try It" with no reference or node to the source problem in the label.

tomjw64:
> Also, it's still a little confusing to me when reading statistics.scss to know which vars were defined in locally in the file and which were defined in statistics-config.scss; any thoughts on somehow casing the ones defined in statistics-config.scss differently?

I was not aware any variables were declared in 'statistics.scss', I was under the impression that all variables declared for a book were done so in the config file. Can you give an example?

tomjw64:
Also I have a documentation related question: Now that the composite page object structure has been removed, is there any documentation anymore that states all valid map-keys and their descriptions/options that you can choose from when you create a page.
For example:

```
/*
Options for Pages:

name - desc
source - desc
isGlossary - desc
sortBy - desc
sectionSeparated - desc
chapterSeparated - desc
hasSolutions - desc
isIndex - desc
compoundComposite - desc
isAnswerKey - desc
*/
```

philschatz:
Unfortunately, the types are not documented (yet). From looking at the mixins, I tried to recreate a list (I pasted it in the body of this PR). I've been trying to get it into sassdoc but haven't seen a way to document types; it seems sassdoc only supports docs on mixins ( [File-Level Annotations](http://sassdoc.com/file-level-annotations/) seemed promising but ended up being a dead-end.

I was thinking of just including the types in a `/books/rulesets/README.md` but if someone knows of a better place I'm all 👂 's!

philschatz:
Oops, forgot to respond to your answers. They make sense, I'll do the renames you mentioned in a separate PR (to keep this one documentation-only).

The only one that is a bit confusing to me is the `source` suggestion. Most of the selectors are classes (but likely to change in  the very-near-future from `.exercise` to `[data-type="exercise"]`) so at that point maybe the name could become `sourceType` to clarify where it is coming from. 

Currently, when I read the config file, it's not clear to me if `source` is just a string (ie for a pending-bucket), or something used in a selector.

books/rulesets/common/_compose.scss
```diff
@@ -218,14 +315,34 @@
...
 @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
   $solutionSource: map-get($solutionPage, source);
-  $isSeparated:  map-get($solutionPage, isSeparated);
+  $isSeparated:  map-get($solutionPage, isSeparated); // TODO: Could this just be $sectionSeparated or $chapterSeparated ?
   $sectionSeparated: map-get($solutionPage, sectionSeparated);
   $chapterSeparated: map-get($solutionPage,chapterSeparated);
+
+  // Validate
+  @include utils_checkType($solutionSource, string);
+  @include utils_checkType($isSeparated, bool);
```
philschatz: but it's not in master yet, right?

philschatz:
💭  The README (linked in the PR body) is currently pretty painful to write; so far I have just been able to organize the mixins into which "Types" they affect. I think they're useful because they act as "primitives" for doing things but I was thinking of pushing them down under another layer. 

(even if not, I think I'll use this organization in the README to help folks understand how the moving parts are related)

I jotted notes in #60 for more mulling/discussion 😄

books/rulesets/common/_compose.scss
```diff
@@ -25,34 +39,57 @@
...
   }
 }
 
-@mixin createChapterComposite($page, $sectionHeaderNode) {
+/// Collects elements defined by `$page.source` (a selector-ish thing) and
+/// collates them at the end of a chapter.
+/// @arg {Map} $page - contains `name`, `source`, and optionally `sortBy` and `compoundComposite` and `isGlossary`
+/// @arg {string} $sectionHeaderNode - Curently Unused (because of typo)
+@mixin _createChapterComposite($page, $sectionHeaderNode) {
+  $name: map-get($page, name);
```
philschatz: Yep, I tried to move them up as much as possible. The impact is that variables are defined at the start of `@mixin`, `@if`, `@else`, and `@each`.

Also, validation occurs immediately after variables are declared (so you can visually verify the variables are validated)

books/rulesets/common/_compose.scss
```diff
@@ -218,14 +315,34 @@
 }
 
 
+/// Create solutions at the end of the book (by creating a new collated Page).
+/// These can be separated by section, chapter, both, or none.
+///
+/// When they are separated by chapter/section a new element is created with the Chapter/Section title.
+///
+/// @arg {list} $compositePages - eah element contains `source` and `hasSolutions`
+/// @arg {Map} $solutionPage - contains `source`, `isSeparated`, `sectionSeparated`, and `chapterSeparated`
```
philschatz: Depending on which PR is merged first, I think this would need to be fixed in the other PR since one adds documentation to existing code and the other changes existing code (and updates documentation)

books/rulesets/README.md
```diff
@@ -28,3 +28,48 @@ The common documentation is in `./common/examples/_all.scss` only because phil d
...
+- `chapterSeparated` - bool
+- `hasSolutions` - bool
+- `isIndex` - bool
+- `compoundComposite` - bool
+- `isAnswerKey` - bool
+
+
+# TitleContent
+
+Describes numbering and labeling of elements. Also known as FigNumber, TableNumber
```
philschatz: ✅  It's poorly worded ; ( I extracted those strings from the variables `$captionFigNumber` and `$captionTableNumber`. Updated the doc

books/rulesets/common/_compose.scss
```diff
@@ -218,14 +315,34 @@
 }
 
 
+/// Create solutions at the end of the book (by creating a new collated Page).
+/// These can be separated by section, chapter, both, or none.
+///
+/// When they are separated by chapter/section a new element is created with the Chapter/Section title.
+///
+/// @arg {list} $compositePages - eah element contains `source` and `hasSolutions`
```
philschatz: ✅  pushed fix

books/rulesets/common/_compose.scss
```diff
@@ -218,14 +315,34 @@
 }
 
 
+/// Create solutions at the end of the book (by creating a new collated Page).
+/// These can be separated by section, chapter, both, or none.
+///
+/// When they are separated by chapter/section a new element is created with the Chapter/Section title.
+///
```
philschatz: ✅  I added your note to the sassdoc comment.  

ooh, is there a way to validate that?

tomjw64:
@philschatz 
>The only one that is a bit confusing to me is the source suggestion. Most of the selectors are classes (but likely to change in the very-near-future from .exercise to [data-type="exercise"]) so at that point maybe the name could become sourceType to clarify where it is coming from.
>
>Currently, when I read the config file, it's not clear to me if source is just a string (ie for a pending-bucket), or something used in a selector.

If the selectors of all `<section>`s were changed to use data-type rather than class, then I agree that the logic could be more simplified!
The `source` string is used mainly to be put into the selector for the targeted section. It was also used as a bucket name as a matter of convenience and being able to tell what kind of section a bucket represents while debugging.

philschatz:
Is there anything left for this PR? (Travis is now happy)

books/rulesets/common/_compose.scss
helenemccarron: I am not sure, since this would be in a different map but I like the idea :)

PR #56, opened 2016-09-17
fix the indentation warnings

philschatz:
small fix to error when the indentation is not correct

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -33,7 +33,7 @@ rules:
   #   -
   #     style: uppercase
   indentation:
-    - 1
+    - 2
     -
       size: 2
   property-sort-order:
--- a/books/rulesets/_economics-config.scss
+++ b/books/rulesets/_economics-config.scss
@@ -246,27 +246,27 @@ $targetLabels: (
   (
     selector: '.eoc [data-type="exercise"]',
     label: "Exercise " counter(chapter) "." counter(exercise)
-  )
+  ),
   (
     selector: 'div[data-type="chapter"] .example',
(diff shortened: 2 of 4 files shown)


books/rulesets/_economics-config.scss
```diff
@@ -246,27 +246,27 @@ $targetLabels: (
   (
     selector: '.eoc [data-type="exercise"]',
     label: "Exercise " counter(chapter) "." counter(exercise)
-  )
```
reedstrm: Hmm, this doesn't look like indentation - does adding commas here actually have no syntactic difference?

.sass-lint.yml
```diff
@@ -33,7 +33,7 @@ rules:
   #   -
   #     style: uppercase
   indentation:
-    - 1
```
reedstrm: I'm not a big fan of 1 space indenting, but ... has everyone agreed that 2 is the right number, or is it the tool driving this?

books/rulesets/_economics-config.scss
philschatz: Yep, it seems SASS allows you to drop commas in this case and the linter does not respect the lack-of-commas. I think maybe the reason SASS allows optional commas in lists is because CSS sometimes uses commas, and sometimes doesn't http://sass-lang.com/documentation/file.SASS_REFERENCE.html#lists

.sass-lint.yml
philschatz: oh, we're using 2 spaces. the `1` here is the warning level (0=ignore, 1=warn, 2=error). The next line below defines the indentation level to be 2 spaces 😄

PR #57, opened 2016-09-20
Clean up EOB chapter/section separated

helenemccarron:


books/rulesets/common/_compose.scss
```diff
@@ -217,35 +217,27 @@
...
-    $map: map-get($map, $key);
-  }
-  @return $map;
-}
 
 @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
   $solutionSource: map-get($solutionPage, source);
   $isSeparated:  map-get($solutionPage, isSeparated);
   $sectionSeparated: map-get($solutionPage, sectionSeparated);
-  $chapterSeparated: map-deep-get($solutionPage,chapterSeparated);
```
philschatz: 💭  Why was `map-deep-get` used here previously?

books/rulesets/common/_compose.scss
```diff
@@ -217,35 +217,27 @@
...
-@function map-deep-get($map, $keys...) {
-  @each $key in $keys {
-    $map: map-get($map, $key);
-  }
-  @return $map;
-}
 
 @mixin compose_createEOBSolutions($compositePages, $solutionPage, $sectionHeaderNode) {
   $solutionSource: map-get($solutionPage, source);
   $isSeparated:  map-get($solutionPage, isSeparated);
```
philschatz: Doh, I forgot to check if isSeparated is still used... if not, could it be removed here and in the *-config.scss files?

books/rulesets/common/_compose.scss
helenemccarron: That was left-over from my attempt to use nested maps to achieve something similar to the `enum` you mentioned.

books/rulesets/common/_compose.scss
helenemccarron: Ah, that one slipped, removing.

PR #58, opened 2016-09-20
fix a couple bugs

philschatz:
Small PR that does a little cleanup. I forgot to regenerate the output CSS files, fixed a backticks bug.


scripts/generate-guide
```diff
@@ -21,21 +21,14 @@ if [ -z "${BOOK_NAME}" ]
 then
   echo 'ERROR: Argument Missing. You must specify the name of the book as the 1st argument. For example: physics'
   exit 1
-else
-  # Check that the CSS file exists
```
philschatz: This test is no longer needed because the output CSS file is always generated (a couple of lines later)

PR #61, opened 2016-09-21
Add config file for book UUIDs and which ruleset they map to

philschatz:
... to keep everyone in-sync, make it easier to fetch updates, and make it
easier to bake a book.

So now, to do all the stuff with a book you would do:

```sh
./scripts/fetch-book microeconap
./scripts/bake-book microeconap
./scripts/report-book-coverage microeconap
```

Previously, you:

- had to specify a UUID in order to fetch a book (it now should always get the latest version)
- could not bake a book other than one that matched the ruleset name
- could only have 1 econ or stats book at a time

The names are not the greatest, but they can be changed _really_ easily (just edit `/books.txt`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@
 
 # Cook a New Book
 
-1. run `./scripts/fetch-book statistics ${ID}`
-  - **Example:** `./scripts/fetch-book statistics 69619d2b-68f0-44b0-b074-a9b2bf90b2c6@11.330`
-  - **Note:** This will require you to log in via ssh. Also, make sure the **ID** contains the version.
+1. run `./scripts/fetch-book statistics`
+  - **Note:** To see the list of books available see `./books.txt`
+  - **Note:** This will require you to log in via ssh.
   - **Note:** your can set the ssh username by running `USER=myusername ./scripts/fetch-book ...`
 1. run `./scripts/bake-book statistics`
 
@@ -17,7 +17,7 @@ There are 2 major parts to cooking a book (_listed above_). You will first need 
 
 ## Add a New Book to the config
 
-Sometimes you need to add a new book (like "dark-matter") into this repo. Here are the steps to add it:
+Sometimes you need to add a new book (like "dark-matter-for-dummies") into this repo. Here are the steps to add it:
 
 1. Find the UUID of the book you want
   - point your browser to `cte-cnx-dev.cnx.org` and find a book you want.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -17,6 +14,17 @@
 
 There are 2 major parts to cooking a book (_listed above_). You will first need to get the single-file HTML from the server (`fetch-book`) and then convert the single-file HTML locally into the "baked" book via `bake-book`. Once you have done the first part, you can run `./scripts/bake-book statistics` to your :heart:'s content!
 
+
+## Add a New Book to the config
```
helenemccarron: Should there be a note to update the version of the book if it changes? Or a method to specify it when passing the book title? This might lead to confusion if changes are made to the collection and are not reflected in the fetched data.

philschatz: The `./books.txt` file does not contain the version number, so when you fetch a book it _should_ fetch the latest version of the book.

/cc @reedstrm to make sure

philschatz:
@dtwilliamson mentioned these issues:

`BOOK_LIST` could be an array instead of a string: `BOOK_LIST=(foo bar baz)`. You could use an associative array instead of per-book vars and `eval` (AKA evil). Speaking of that instead of `eval $(cat ${BOOK_CONFIG_FILE})` <shudder>, use `. "$BOOK_CONFIG_FILE"` (or use `source` instead of `.` if you find it more readable). Quotes, quotes, always use quotes. I just looked at books.txt and the script that immediately followed it.

Using the array reference? You can do that or you can use `read` like this: `read -r foo bar < <(make_foo_and_bar)` then you don't have to do a second assignment

note that using `read` will result in any excess being included in the last variable. To protect against that, you can include a throw away variable to catch the excess if any `read -r foo bar _ < <(make_foo_and_bar)``

README.md
```diff
@@ -6,9 +6,6 @@
 
 # Cook a New Book
 
-1. Find the UUID of the book you want
-  - point your browser to `cte-cnx-dev.cnx.org` and find a book you want.
-  - click "More Information" and copy the **ID**
 1. run `./scripts/fetch-book statistics ${ID}`
```
helenemccarron: This is still referencing an ${ID}, should this be updated?

helenemccarron: Also, how do I know what the book is called, you mention microeconap for example but my tendancy would be to call it apmicroecon. Could you mention here where to look to find that information?

philschatz: Thanks, updated both!

philschatz:
Yep, it does in fact fetch the latest version of the book 😄  I checked the browser version of stats and microecon against the `/data/*-raw.html` version text inside the file and they matched!

PR #63, opened 2016-09-21
Note Titles

helenemccarron:
Create the note titles via the ruleset, create a body container for existing content and preserving original title


philschatz:
ooh, can you add those notes to the guide? It would be as comments just above where they are defined in `_economics-config.scss` (with a little HTML snippet for each)

books/rulesets/common/_modify.scss
```diff
@@ -94,3 +94,24 @@
     }
   }
 }
+
+
+@mixin modify_notes($type, $noteContent) {
+  [data-type="note"].#{$type} {
+    @include utils_title($noteContent, bNoteLabel-#{$type});
```
philschatz: Is the type actually needed here? If not, what do you think about renaming `$type` to `$selector` (and including a dot)? That way it's clear that `bringhome` below (in the main sass file) is a class name.

Also, what if the type/selector was on the `$linkupNote` variable inside the `-config.scss` file? That way the title, selector, and documentation are in one place.

books/rulesets/common/_modify.scss
```diff
@@ -94,3 +94,24 @@
...
+    &::before {
+      container: h3;
+      content: pending(bNoteLabel-#{$type});
+      class: "title";
+      move-to: bNoteHeader;
+    }
+    &::after {
+        class: "body";
```
philschatz: Any chance this could be something like `note-title`? Having 2 words makes it easier to search for when looking to see where/how an element/attribute is created. Similar to bootstrap's `class="modal-body"`

books/rulesets/common/_modify.scss
```diff
@@ -94,3 +94,24 @@
     }
   }
 }
+
+
+@mixin modify_notes($type, $noteContent) {
```
philschatz: it seems this does something like "modify_noteWrapBody", rather than generically modify a note. Also, it modifies a single note, not a list of notes.

The reason I noticed was because I was wondering why in `economics.scss`:

- notes we `modify_notes` except the linkup one 😄 
- each note needed a separate "modify_note _s_" call

books/rulesets/common/_modify.scss
helenemccarron: `$selector` would work, but not the dot, that selector is used within the move-to as well.  I can move it to the config file, that makes sense.

books/rulesets/common/_modify.scss
helenemccarron: I don't mind the new classes. However, does this open the door to multiple somewhat unique arbitrary classes created on the fly? If that is a direction we want to go, should we establish conventions on how those are constructed?

books/rulesets/common/_modify.scss
helenemccarron: Yes on note, not notes :) 

Regarding the mixin itself, yes, it resembles number_numberNote (not count_countNote) but does slightly more (or less, depending). It doesn't trash the existing title, creates a new title and wraps the original title and the contents. The reason why I used number_numberNote for `linkup` is admittedly in the interest of time, because it does "just" what I needed it to do, create a title without creating the body. I prefer modify_note in both cases, because there is no numbering involved with those notes. The new mixin is handling/creating the titles and modifying the note accordingly. I think the `iswrapped` logic could then be created here instead of using `number_numberNote`.

Concerning the name, the mixin creates a title and wraps the body, that seems a bit long for a mixin name, so I simplified to "modify". Maybe it could be `modify_titleWrapNotes`? It's a bit lenghty but would be more descriptive.

books/rulesets/common/_modify.scss
philschatz: Couldn't the `move-to:` also just be `bNoteLabel` (without the type)?

books/rulesets/common/_modify.scss
philschatz: yeah, I guess it would open that door. Initially I thought at least the classes could be book-specific but that is probably not the case.

Maybe there should be a larger discussion (rather than in comments here). We could do something that's easily greppable here and ugly (so it "sticks out like a sore thumb"), discuss, and then incorporate the decision here.

Maybe something like `class="-os-note-contents"` or `data-baked="note-contents"` for now 😄 ?

books/rulesets/common/_modify.scss
philschatz: Thanks! I read and could we discuss over voice?

After noticing the `countNote` and `numberNote` mixins I rethought my suggestion of naming it something like `modifyNoteWrapBody`. To keep some level of consistency, it seems like either:

- the mixins should do one small atomic task and be clearly named with what that task does) (like `modifyNoteWrapBody`)
- or, the mixins should be named a little more generically and have conditional logic in them (like `modifyNote` which checks an `isBodyWrapped` field)

Thoughts?

philschatz:
I don't think the structure of the comments you added will be parsed and included in the style guide. Example: https://github.com/Connexions/cnx-recipes/blob/master/books/rulesets/_economics-config.scss#L13-L22

philschatz:
Also, when you generate the guide elements items will currently be duplicated/triplicated (because kss parses the comments in the SASS files as well as the comments in `output/*.css`. The fix for that is part of #64

PR #65, opened 2016-09-26
Titles - refactor, various bug fixes

helenemccarron:


books/rulesets/common/_number.scss
```diff
@@ -126,10 +129,19 @@
     }
   }
   @if ($hasTitle) {
-    #{$defaultContainer} [data-type="title"] {
-      container: span;
-      class: "title";
-      move-to: bCaption;
+    @if $type == "table" {
```
philschatz: Is there a convention for wrapping the `@if` expression in parentheses? I think this is the first time I've seen an `@if` without parentheses.

If so/not could it be added to `.sass-lint.yml`?

PR #71, opened 2016-09-30
Add $pageTestPrep

helenemccarron:
Add the end of chapter mixin for Test Prep For AP® Courses  and documentations

### DIFF
--- a/books/rulesets/output/economics.css
+++ b/books/rulesets/output/economics.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /*
   Book Page
 
@@ -266,6 +267,16 @@
   Markup: ./examples/book.composite.problems.html-cooked.html
 
   Style guide: book.composite.problems
+*/
+/*
+  TEST PREP FOR AP® COURSES
+
+  These should be collated at the end of a chapter
+  and have answers collated at the end of the chapter.
+
+  Markup: ./examples/book.composite.testPrep.html-cooked.html
+
+  Style guide: book.composite.testPrep
 */
 /*
   References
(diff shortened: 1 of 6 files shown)


books/rulesets/output/economics.css
```diff
@@ -740,6 +763,15 @@
   container: h1;
   data-type: "document-title";
   content: pending(-TITLECONTAINER); }
+:pass(5) [data-type="composite-page"].os-eoc.os-ap-test-prep-container::before {
+  container: span;
+  content: "Test Prep For AP® Courses";
+  class: os-text;
+  move-to: -TITLECONTAINER; }
```
reedstrm: bucket name starts with `-` ?

reedstrm: did some map-key get typoed?

helenemccarron: This appears to be a bug in how the $bucket is used. See PR #72

books/rulesets/output/economics.css
```diff
@@ -1,3 +1,4 @@
+@charset "UTF-8";
```
philschatz: ooh, what caused this to show up in this file and not the other CSS files?

helenemccarron: I believe the use of the (r) character

philschatz: ah yep, Thanks!

philschatz:
@helenemccarron do you mind if I add `TEAmacroeconap` to `./scripts/test`? That way those files get regenerated?

helenemccarron:
Go for it!

PR #74, opened 2016-10-03
Testing: verify CSS files are regenerated when pushing commits

philschatz:
This makes sure that things like https://github.com/Connexions/cnx-recipes/pull/71/commits/1a008d0f657db31739cc9c95cae352d948eb96a3 do not get overlooked.

I created a separate script for Travis that:

1. runs the tests
1. verifies that there were no changes to files stored in git

When CSS files were forgotten, Travis will fail and print out something like:

image

(except that the diff would point to the CSS files, not `./scripts/test` or `./scripts/ci` 😄 )

### DIFF
--- a/scripts/ci
+++ b/scripts/ci
@@ -10,7 +10,8 @@ hasChanges=$(git diff)
 
 if [[ ! -z "${hasChanges}" ]]
 then
-  echo "ERROR: The CSS files were not regenerated. Be sure to run ./scripts/test before pushing"
+  echo "ERROR: Running the tests caused unexpected changes in committed files."
+  echo "Were the CSS files regenerated? Be sure to run ./scripts/test before pushing"
   git diff
   exit 1
 fi
--- a/scripts/ci
+++ b/scripts/ci
@@ -1,0 +1,16 @@
+#!/bin/sh
+set -e
+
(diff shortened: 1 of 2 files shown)


reedstrm:
so since this is content based, it will find things like "the merge did not do the right thing w/ the CSS" (which is what actually happened) but not trigger if the merge _does_ create the same result as would have happened w/ rebuilding the files. I think. Right?

scripts/ci
```diff
@@ -0,0 +1,16 @@
...
+# Run the tests (which happens to re-generate the CSS files)
+./scripts/test
+
+# Verify that the CSS files were regenerated so they never become out-of-sync
+
+hasChanges=$(git diff)
+
+if [[ ! -z "${hasChanges}" ]]
+then
+  echo "ERROR: The CSS files were not regenerated. Be sure to run ./scripts/test before pushing"
```
reedstrm: Perhaps: "Running the  tests make unexpected changes. Were the CSS files regenerated? Be sure <yada yada>"

philschatz:
I _think_ the `continuous-integration/travis-ci/pr` test would fail which would trigger the original dev to `git rebase master` which would catch the issue that actually happened, right?

PR #76, opened 2016-10-04
create book-specific directories for the rulesets

philschatz:
# Book-specific Snippets in Separate Directory

These were moved from `books/rulesets/examples/` into `books/rulesets/${BOOK_NAME}/examples/`

# A Real Book-specific Guide

When you run `./scripts/generate-guide statistics` it will _really_ only generate the statistics guide. To accomplish this, `generate-guide` creates a temporary directory and moves all the `common/` and `${BOOK_NAME}/` SASS/HTML files into that directory.

# Better Travis Failures

Also, if an HTML snippet is missing then Travis will fail and the output will look something like this:

```txt
Step 4/4: Deleting the cooked example HTML files
ERROR: Some HTML files were missing:
 - book.composite.practice: ./examples/book.composite.practice.html-cooked.html NOT FOUND!
```

(to verify, just rename one of the snippet files and run `./scripts/test` 😄 )

# TODO

- [x] ensure the guide only contains stuff for the book (not all books)
- [x] move the snippets into the book-specific directory


helenemccarron:
Would it make sense to move the rulesets to the root and to sort books within the rulesets? Something like
data/
rulesets/
    output/
    examples/
    books/ (or some other name)
        econ/
        stats/

philschatz:
Could the alternative directory structure be a separate PR?

PR #77, opened 2016-10-05
add docs for regression-testing books

philschatz:
This adds documentation for how to regression-test the rulesets by:

1. baking the book with the old rulesets
2. baking the book with the new rulesets
  - and comparing the 2 versions of the baked book

You can skip the first step if the "old" version has not changed.

**Note:** The baked HTML files are stored in `./data/`

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,32 @@ You can pass 3 additional arguments to `report-book-coverage` to change how it r
 
 For more details on the commandline options see the [css-coverage](https://www.npmjs.com/package/css-coverage#commandline-options) documentation.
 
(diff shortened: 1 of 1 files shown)


helenemccarron:
It might be useful to include a method to output the results to an external file. Reading the results in Terminal is a bit awkward, especially when the change is leading to many differences.

philschatz:
Hmm, would adding `> output.txt` to the end of the command address the problem?

helenemccarron:
Yup, that's what I was thinking. Can we add it to the README?

PR #79, opened 2016-10-10
adds commas between links in the index

philschatz:
Previously, links in the index were smushed together; this adds a space between each link.

image

/cc @reedstrm : Is there a way to add a space character to a bucket without creating an element (like via `container: none;`)?

Also, I have a feeling people would prefer each link to be separated by a comma instead of a space. That would require prefixing a comma to each link _except_ the first link; I'm not sure how to do that within `cnx-recipes`

### DIFF
--- a/books/rulesets/output/statistics.css
+++ b/books/rulesets/output/statistics.css
@@ -161,6 +161,15 @@
   Style guide: book.glossary
 */
 /*
+  Book Index
+
+  The book Index contains terms and links to their occurences in the book.
+
+  Markup: ./book.index.html-cooked.html
+
+  Style guide: book.index
+*/
+/*
   Table
 
   The `table` element may contain a `caption`, `tbody`, `col`, and other _usual_ elements.
@@ -527,6 +536,11 @@
   container: span;
   class: "os-term-section";
   move-to: index-section; }
+:pass(2) div[data-type="page"] span[data-type="term"]::after, :pass(2) div[data-type="composite-page"] span[data-type="term"]::after {
+  container: span;
+  content: ' ';
+  class: "os-index-link-separator";
+  move-to: index-section-link; }
 :pass(2) div[data-type="page"] span[data-type="term"]::after, :pass(2) div[data-type="composite-page"] span[data-type="term"]::after {
   container: a;
   content: pending(index-section);
--- a/books/rulesets/output/statistics.css
+++ b/books/rulesets/output/statistics.css
@@ -537,15 +537,15 @@
   class: "os-term-section";
   move-to: index-section; }
 :pass(2) div[data-type="page"] span[data-type="term"]::after, :pass(2) div[data-type="composite-page"] span[data-type="term"]::after {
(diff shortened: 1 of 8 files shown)


helenemccarron:
Oh, I really don't think this kind of element should be in the rulesets, I have left those out on purpose since we can add padding via css. I believe adding this kind of spacing element would lead to confusion down the road as to where to locate it when trying to troubleshoot.

books/rulesets/statistics/book.scss
```diff
@@ -101,6 +101,7 @@
   @include number_setCaptions($setFigureCaption, $captionFigNumber, $captionFigNumberAp);
 }
 :pass(6) {
+  @include modify_trash('.os-index-link-separator:last-child');
```
reedstrm: Why is this once-per-book, rather than in a common/ file?

books/rulesets/output/statistics.css
```diff
@@ -827,6 +841,8 @@
   container: div;
   content: content() pending(bCaptionContainer); }
 
+:pass(6) .os-index-link-separator:last-child {
```
reedstrm: Hmm, per book?

books/rulesets/statistics/book.scss
philschatz: Could/Should this be a named mixin? Like `modify_indexCommaCleanup` or something verbose so it's clear why this mixin is being run?

books/rulesets/output/statistics.css
reedstrm: o.k, good to know it's a design choice (re: passes in the common).  I agree that the comment in `_compose.css` and the actual call (it says `osx-link-separator` right there in the line after all) is sufficient. By using the generic mix-in, it's more clear that it's not doing any special magic, just deleting it. Want me to merge, or will you @helenemccarron ?

books/rulesets/statistics/book.scss
philschatz: 💭  I was thinking that we do not currently expose selectors that are generated within the mixins (`.os-index-link-separator`) up in the main `book.scss`. If I was looking at `book.scss` for the first time I would be confused as to where that class name came from; that's why I was suggesting that it be pushed down into the mixins (similar to the `_toc.scss` file)

helenemccarron: Mmh, it is true that is the first time it happens. However, I wonder if pushing it to an obscure part of the ruleset would make it easier to find/ understand? Either way, it would have to be heavily documented since it couldn't live along the in index files. The toc is a complete block for all of it that lives on its own when the index is living among the other parts of the book. Having a single index related mixin living in a strange place might be even more confusing.

reedstrm: Convinces me. :-)

helenemccarron: I agree that the index could be its own set of mixins. Can we do that in a separate PR, though? That is going to require some shuffling around.

PR #80, opened 2016-10-17
remove unused python packages

philschatz:
It seems these packages are no longer used in any of the scripts. I think this is because the cnx-epub work is done on the deployed dev machine rather than locally but I'm not 💯 % sure.

/cc @reedstrm

### DIFF
--- a/scripts/ci
+++ b/scripts/ci
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -e
 
 # Run the tests (which happens to re-generate the CSS files)
--- a/scripts/setup
+++ b/scripts/setup
@@ -12,10 +12,7 @@ virtualenv .
 echo "Step 3/4: Installing python packages"
 
 pip install --quiet git+https://github.com/Connexions/cssselect2.git
-pip install --quiet git+https://github.com/Connexions/rhaptos.cnxmlutils.git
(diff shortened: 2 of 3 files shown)


reedstrm:
Hmm, does the use-case for `cnx-recipes` not include `cnx-epub-single_html` and `cnx-epub-validate-collated`? That would need cnx-epub, which should pull in cnx-archive via dependencies, doesn't need an explicit install.

philschatz:
Currently `scripts/fetch-book` runs those commands on the remote dev machine so I do not think those packages are used directly in `cnx-recipes` (see [scripts/fetch-book](https://github.com/Connexions/cnx-recipes/blob/master/scripts/fetch-book#L51-L56) )

philschatz:
soo, I think I addressed @reedstrm 's questions; is there anything else left?

reedstrm:
you didn't so much as address my comment as dismiss it, saying your workflow, as embodied in the scripts, is the only supported workflow. For `cnx-epub-validate-collated` in particular:

helenemccarron:
Oops, I think I merged it too soon :( Sorry about that

philschatz:
oh cool, I didn't know we had a way to validate! an example wasn't in the docs dir so I never knew you could do that. I'll continue discussion in #80 😄

PR #83, opened 2016-10-19
Remove the abstract from metadata

helenemccarron:
(refs https://github.com/Connexions/webview/pull/1460 )


philschatz:
Would it make sense to make this a mixin so there can be documentation as to why the abstract is being discarded (because the abstract is being moved into the content for these books)?

helenemccarron:
I'm not in favor of creating a special mixin for each specific action, especially when using something as basic as trashing an element. If we were to follow that path, each time we trash something, move it or modify it, we would need a specific mixin. This said, I am thinking about combining all those actions into a base file, which could be documented once and applied to all the books that would require it.

PR #84, opened 2016-10-19
Add unnumbered support to tables and figures

helenemccarron:
Tables and figures with the class `.unnumbered` should not have a caption or be numbered

### DIFF
--- a/books/rulesets/common/examples/_all.scss
+++ b/books/rulesets/common/examples/_all.scss
@@ -254,13 +254,14 @@
 /*
   Unnumbered Content
 
-  Content that has been marked with the class .unnumbered by the content team.
+  Content that has been marked with the class `.unnumbered` by the content team.
 
-  This may apply to a note, a table, an equation, a figure. Currently, the ruleset supports only tables and figures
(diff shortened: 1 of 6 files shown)


philschatz:
Can you add an example with an unnumbered table to https://github.com/Connexions/cnx-recipes/blob/master/books/rulesets/common/examples/_all.scss#L202-L218 ? 

That way folks can see/know that it's a _special_ class that we handle properly

books/rulesets/common/examples/_all.scss
```diff
@@ -251,6 +251,18 @@
   Style guide: book.chapter
 */
 
+/*
+  Unnumbered Content
+
+  Content that has been marked with the class .unnumbered by the content team.
+  T
```
philschatz: Is this a typo?

helenemccarron: I would appear so :-D

philschatz:
Mind if I:

1. move this to be `page.unnumbered` (since the class isn't used on book-level elements)
2. include a Figure example

helenemccarron:
sure, go for it

PR #85, opened 2016-10-19
exercise/solution links use the exercise id instead of a new id

philschatz:
(refs https://github.com/Connexions/webview/pull/1460)

It also helps with transclusion and reduces the number of new id’s generated.

<details>
<summary>

Example Raw HTML:</summary>

``` html
<div data-type="exercise" class="exercise">
  <div data-type="problem" class="problem">
    <p>What is 2+2?</p>
  </div>
  <div data-type="solution" class="solution">
    4
  </div>
</div>
```

</details>

<details><summary>

Old cooked HTML:</summary>

``` html
<div data-type="exercise" class="exercise os-hasSolution " id="ex123">
  <div data-type="problem"><a class="os-number" id="problem-1" href="#solution-1">1</a>
    <p>What is 2+2?</p>
  </div>
</div>

<!-- ... and the EOB solutions area: -->

<div class="os-eob os-solutions-container" data-type="composite-page">

  <div data-type="solution" class="solution">
    <a id="solution-1" href="#problem-1">
      <span class="os-number">1</span>4</a>
    4
  </div>
</div>
```

</details>

New Cooked HTML:

``` html
<div data-type="exercise" class="exercise os-hasSolution " id="ex123">
  <div data-type="problem" class="problem">
    <a class="os-number" href="#ex123-solution">1</a>
    <p>What is 2+2?</p>
  </div>
</div>

<!-- ... and the EOB solutions area: -->

<div class="os-eob os-solutions-container" data-type="composite-page">

  <div data-type="solution" id="ex123-solution">
    <a href="#ex123">
      <span class="os-number">1</span>
    </a>
    4
  </div>
</div>
```

Also, this fixes a couple other issues:
1. selectors now use `[data-type="exercise/problem/solution"]` instead of `.exercise/problem/solution`
2. the "4" in the answer is no longer duplicated in the link text

### DIFF
--- a/books/rulesets/common/_link.scss
+++ b/books/rulesets/common/_link.scss
@@ -1,14 +1,24 @@
-@import "./utils";
-
 //EOC composite pages must be created before this runs
-@mixin link_linkToProblemsFromSolutionsEOC($toElement, $fromElement, $problemCounter) {
-  @include utils_linkToFromForward(
-    ".os-hasSolution [data-type="problem"] .os-#{$toElement}",
-    "[data-type="solution"] .os-#{$fromElement}",
-    problem,
-    solution,
-    $problemCounter
-  );
+/// Creates a link to the solution and a link back to the problem for an exercise
+/// Note: This replaces the solution id with a new id whose value is `${EXERCISEID}-solution`
+@mixin link_linkToProblemsFromSolutionsEOC() {
+
+  [data-type="exercise"].os-hasSolution {
+    string-set: exerciseId attr(id);
+
+    [data-type="problem"] .os-number {
+      container: a;
+      attr-href: "#" string(exerciseId) '-solution';
+    }
+
+    [data-type="solution"] {
+      attr-id: string(exerciseId) '-solution';
+      .os-number {
+        container: a;
+        attr-href: "#" string(exerciseId);
+      }
+    }
+  }
 }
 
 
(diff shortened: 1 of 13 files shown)


reedstrm:
Ah, at first I was concerned that an exercise could contain more than one problem, but that is not the case. We allow more than one solution, though.

philschatz:
Yep, that's a problem with the existing implementation as well. Maybe I should add a `:first` to the `[data-type="solution"]` selector. But we'd also have to:

1. decide what to do with the 2nd/3rd/4th solutions
  - should they be labeled 23a, 23b, 23c ? Should they be discarded?
2. (depending on ^^^ ) make sure to only move the solutions that we want to move

helenemccarron:
CC @Stackblocks

philschatz:
This also allows webview to have CSS like `:target { background-color: yellow; }` which shows the student which exercise/figure/table was linked to:

image

reedstrm:
@helenemccarron any issues with this? Seems to make sense to me.

helenemccarron:
Alan is currently testing it to make sure that nothing was modified by this change.

Stackblocks:
Able to navigate between the problem sets and answer keys correctly in multiple books. Encountered issue with webview not directly taking you to the answer but per Helene that is known.

PR #89, opened 2016-10-21
Physics ruleset

helenemccarron:
## TODO
- [x] add guide files
- [x] change `@mixin count_countEOCExercises($name)` to support different numbering schemes (see TODO)
- [ ] _maybe?_ Create a common file, at least for Physics and HS Physics
## Note:

I had to separate AP physics and Physics, AP physics has an answer key while Physics doesn't. 
This raises the question of templating
- base common base to all books
- how to extend an existing ruleset to include features present only in the child ruleset
- new folder structure to isolate the books from the common folders and other guide folders (for easier navigation)


philschatz:
Is this related to #89 ?

helenemccarron:
They are related, I'll merge them and rebase master.

philschatz:
Given that this is a third book being added, are there plans to reorganize code to reduce common tasks done in the books (instead of copy/pasting from an existing book)?

helenemccarron:
Yes, I will be working on it after these books

books.txt
```diff
@@ -30,6 +30,22 @@ TEAmacroeconap_ruleset_name=economics
...
+physics_uuid=7f6fbc5b-6f3b-40a8-87ac-b14adc941e5e
+physics_ruleset_name=physics
+
+apphysics_uuid=8d04a686-d5e8-4798-a27d-c608e4d0e187
+physics_ruleset_name=physics
+
+hsphysics_uuid=051ff00c-34b9-4547-ab03-fe5b1b95cf62
+hsphysics_ruleset_name=hs-physics
+
+physics_uuid=7f6fbc5b-6f3b-40a8-87ac-b14adc941e5e
```
philschatz: It seems like these are duplicate entries and can be removed, no?

books/rulesets/ap-physics/book.scss
```diff
@@ -0,0 +1,133 @@
...
+// $chapter_de: "Kapitel";
+// $appendix_de: "Anhang";
+//TODO: Add warnings/flags for warnings/documentation when bundles are called in an unexpected order (dependencies for mixin calls?)
+
+:pass(0) {
+  //Some modifications to be done to the book before any collation
+  @include modify_titlePreface();
+  @include modify_spanWrapTitles();
+  @include modify_simLinkTarget();
+  @include utils_hasSolution();
```
philschatz: It seems like this file differs from `physics/book.scss` by 5 lines. Just noting; I don't mind merging as long as the physics refactor gets priority & time 😄

books.txt
```diff
@@ -30,6 +30,22 @@ TEAmacroeconap_ruleset_name=economics
 TEAmicroeconap_uuid=7f6fbc5b-6f3b-40a8-87ac-b14adc941e5e
 TEAmicroeconap_ruleset_name=economics
 
+physics_uuid=7f6fbc5b-6f3b-40a8-87ac-b14adc941e5e
+physics_ruleset_name=physics
+
+apphysics_uuid=8d04a686-d5e8-4798-a27d-c608e4d0e187
```
philschatz: Probably worth discussing a naming strategy between the ruleset name and the book name. `dash-case`, `snake_case` and `camelCase` are all allowed and it might be nice to have the book name match the ruleset name for books (then again, for greppability it may not be desired). Just fodder for discussion 😄

books/rulesets/ap-physics/book.scss
helenemccarron: I just want to make sure we are on the same page, Ap-physics and physics are very close but not identical, @Stackblocks is testing both of them and they other than for the appendices, they should be working properly. HS physics is very different, currently incomplete and not fully tested yet.

books.txt
helenemccarron: Agreed, for the record, I tried using `ap-physics` for consistency but it wasn't working when I tested it. Maybe I missed something?

helenemccarron:
Statistics and Econ were tested for regressions and found working as intended. HS Physics is currently incomplete and untested, it will be addressed in a separate PR

PR #90, opened 2016-10-21
rename "cook" to "bake"

philschatz:
also had to rename “cooked” to “baked” and “cooking” to “baking” while
preserving case
# TODO

I might also clean up the scripts so:
- [x] virtualenv is installed in `/venv/` instead of `.` (makes it easier to remove it since the `/bin`, `/lib/`, `/pip-selfcheck.json`, and `/src/` dirs will all be in the `./venv` dir.
- [ ] clean up the bash scripts so they follow dennis' conventions in #61

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,7 @@
 # working files
 /books/data/
 /data/*-raw.html
-/data/*-cooked.html
+/data/*-baked.html
 # The following file is used for diffing 2 versions of a book
 /data/*-prepared.html
 # CSS Code coverage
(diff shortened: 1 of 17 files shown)


reedstrm:
While I'm usually all for consistency in nomenclature, how far do we want to take this? I note that internally, cnx-epub, cnx-archive, and cnx-publishing refer to this as `collated`. Changing that will take a schema change (it's in table and column names). If were going to do it, best do it before any of that is rolled out.

reedstrm:
These scripts are mainly used by textbook recipe authors, so I'd like y'all to chime in on if you care one way or the other. :-)

helenemccarron:
I don't care either way :)

PR #91, opened 2016-10-24
Remove dividers from the rulesets

helenemccarron:
Also removes extra chapter labels in titles

### DIFF
--- a/books/rulesets/economics/_config.scss
+++ b/books/rulesets/economics/_config.scss
@@ -170,18 +170,15 @@ $bookCompositePages: (
 //Usage Note: key: class of the containing span, value: text that will go inside span
 $chapterTitleContent: (
   number: counter(chapter),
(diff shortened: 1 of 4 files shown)


philschatz:
💭  Should the divider code get removed/commented or we create a catch-all book called `books/rulesets/unused-features.scss` with code that exercises those unused features so they do not bit-rot / accidentally stop working?

helenemccarron:
For dividers, they are just part of the map, they can be there or not along with other parts of the titles, so I don't think there is any risk of bit-rot. Maybe we should update the guide to mention that the new title structure is `<title><span class="os-label></span><span class="os-number></span><span class="tos-ext"></span></title>`

philschatz:
I merged this without deploying first and noticed that we probably need to switch the divider to be a space instead:

image

PR #93, opened 2016-10-27
Add header treatment to all notes and examples in Statitics

helenemccarron:
All notes and example should have proper headers


philschatz:
@helenemccarron It seems like all the CSS files were not regenerated (econ) 😄

helenemccarron:
ooh that's what that was!

PR #94, opened 2016-10-27
add script to retrigger baking a book remotely

philschatz:
This adds `./scripts/bake-book-remote ${BOOK_NAME} ${VERSION}` which retriggers a bake using the css file in `./books/rulesets/output/`

# Rebake

To update a development instance you need to rebake a book.
To do that, you will need to upload a CSS ruleset file, trigger a rebake, and then clear the cache.

1. run `./scripts/compile-books` to generate the rulesets CSS file
1. run `./scripts/bake-book-remote ${BOOK_NAME} ${VERSION}` to upload, trigger rebake, and clear the cache

# TODO
- [x] update `README.md`

/cc @Stackblocks for feedback

PR #95, opened 2016-10-29
add space divider for chapters and sections

philschatz:
This partially undoes the changes in #91 to add better spacing between the chapter number and the title.

image

# Test Criteria

- re-bake the economics book and the statistics books on the dev machine and verify the spacing in the screenshot is correct

**Note:** This depends on #94 being merged (so @Stackblocks can rebake the dev machine and verify)

### DIFF
--- a/books/rulesets/economics/_config.scss
+++ b/books/rulesets/economics/_config.scss
@@ -170,15 +170,18 @@ $bookCompositePages: (
 //Usage Note: key: class of the containing span, value: text that will go inside span
 $chapterTitleContent: (
   number: counter(chapter),
+  divider: ' '
 );
 
 $appendixTitleContent: (
   title-label: "Appendix ",
   number: counter(appendix, upper-alpha),
+  divider: ' '
 );
 
 $sectionTitleContent: (
   number: counter(chapter) "." counter(section),
+  divider: ' '
 );
 
 $exerciseTitleContent: (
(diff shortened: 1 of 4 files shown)


helenemccarron:
This goes against what was discussed during last meeting. This type of separation can be done via styling rather than via the ruleset, we discussed using ` ::before` and `::after` to recreate those dividers (this would include extra padding) instead of adding them in the ruleset, shouldn't we use that approach here?

philschatz:
There are some cases where styling is not possible:

- Archive URLs
- ePub 2 files
- external user wants to use the baked book (ie TEA)

Also, I do not think it is possible to detect whether an element `:has(.os-divider)` in CSS which means that if a book wants to have a `|` divider between the chapter/section then there is no way to do something like:

```css
.title.section:not(:has(.os-divider)) > .os-text { padding-right: 1em; }
```

Also, that padding code would need to be duplicated in every place where the title could appear (even in places outside of the content) which could be easily missed:

- Blue title on the Page
- ToC
- Index links
- Search Results
- Page Metadata

helenemccarron:
I think the idea was simply to simply use something like `.os-label::after {padding: 10px;}`, which would also be doable for external users but I didn't think about Archive URLs, it does make sense in that case. I also agree with you, I'd rather have it in one place rather than several rules scattered across platforms.

PR #96, opened 2016-10-31
Clean up dependencies

philschatz:
@Stackblocks had fewer globally-installed commands and was already running in a python virtualenv so I took that as an opportunity to install dependencies (sass and python virtualenv) using the language-specific ways to do it:

- specify the os-level packages to install (using homebrew for OSX folks)
- use the same version of ruby
- same version of sass

Note: I did not specify a specific version of node

# Test Criteria

- [ ] ensure `./scripts/setup && ./scripts/test` works (Travis should already do this too)
- [ ] ensure `./scripts/bake-book ${BOOK_NAME}` still generates a book


scripts/setup
```diff
@@ -1,23 +1,37 @@
...
+then
 
-deactivate
+  echo "==> Installing python packages"
 
-echo "Step 4/4: Installing sassdoc (depends on NodeJS)"
+  pip install --quiet requests # Used for rebaking a book
+  pip install --quiet git+https://github.com/Connexions/cssselect2.git
+  pip install --quiet git+https://github.com/Connexions/cnx-easybake.git
+  pip install --quiet git+https://github.com/Connexions/cnx-epub.git
```
reedstrm: shouldn't this be a `requirements.txt` file and a `pip install -r requirements.txt`

philschatz: Sure! What format should the `requirements.txt` have? I tried searching for it in cnx-epub, cnx-archive, and rhaptos.cnxmlutils. `pip freeze > requirements.txt` (from [StackOverflow](https://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source) results in the following (no git repo):

```
cnx-easybake==0.7.0
cnx-epub==0.12.0
cssselect==1.0.0
cssselect2==0.2a0
Jinja2==2.8
lxml==3.6.4
MarkupSafe==0.23
requests==2.11.1
tinycss2==0.5
webencodings==0.5
wheel==0.24.0
```

Also, how would the `scripts/update` script look in that case? `pip update -r requirements.txt`?

PR #97, opened 2016-11-02
Script cleanup

philschatz:
Finally getting around to @dtwilliamson 's comments in #61 😄 

**Note:** This contains the "rename cooked to baked" PR only because I did not want to redo this renaming after the PR was merged

This also uses https://github.com/Connexions/cnx-easybake/pull/45 to generate precise CSS coverage files for coveralls (see the end-result at: https://coveralls.io/github/Connexions/cnx-recipes )

# TODO

- [x] output `==>` instead of `Step 1/4:` to make scripts easier to maintain
- [x] use `>&2 echo "foo"` instead of `echo "foo"`
- [x] rename vars to be camelCase
- [x] activate virtualenv using `source ./venv/bin/activate` instead of `. ./venv/bin/activate`
- [ ] use double-quotes when calling echo
- [x] change how `/books.txt` is formatted and loaded
- [ ] remove `set -e` from the scripts by adding `|| exit 1` to every command
- [ ] change if statements to use `[[ ... ]]` instead of `[ ... ]`


philschatz:
Coverage now shows the coverage for individual selectors (run `./scripts/test`). The red lines were not matched at all while the 2nd selectors were matched. At the bottom is an example showing that both selectors were matched:

image

PR #98, opened 2016-11-03
make dependencies more pythonic

philschatz:
This moves the python dependencies into a separate file (`requirements.txt`) to be more pythonic.

So now we have the following files:

- `/requirements.txt` : python dependencies
- `/Gemfile` : ruby dependencies
- `/package.json` : JavaScript dependencies
- `/Brewfile` : OSX system dependencies
- `/.ruby-version` : Version of ruby
- `/.nvmrc` : Version of node

/cc @reedstrm Thanks!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+requests
+git+https://github.com/Connexions/cssselect2.git#egg=cssselect2
+git+https://github.com/Connexions/cnx-easybake.git
+git+https://github.com/Connexions/cnx-epub.git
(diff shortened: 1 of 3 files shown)


reedstrm:
unified build system, eh?

requirements.txt
```diff
@@ -0,0 +1,4 @@
+requests
+git+https://github.com/Connexions/cssselect2.git#egg=cssselect2
+git+https://github.com/Connexions/cnx-easybake.git
```
reedstrm: for dependency resolution cases, these should also have `#egg=packagename`, so `#egg=cnxeasybake` and `#egg=cnxepub`, respectively. It doesn't matter in this case, since these installs are just to get the cli scripts. This is not a python package context.

PR #103, opened 2016-11-10
Fix ruby bundler

philschatz:
People used to get the following error:

```
$: ./scripts/setup

==> Updating Homebrew…
Already up-to-date.
==> Installing Ruby Bundler gem…
2.3.1 (set by /path/to/cnx-recipes/.ruby-version)
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.

$:
```

Now they get a message with how to fix the problem:

```
$: ./scripts/setup

==> Updating Homebrew…
Already up-to-date.
==> Installing Ruby Bundler gem…
2.3.1 (set by /Users/phil/github/cnx/cnx-recipes/.ruby-version)
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.


--------------- Error -------------
Unable to install the bundler.
You may need to add the following to ~/.bash_profile :

eval "$(rbenv init -)"

Doing this will require restarting your terminal
or running "source ~/.bash_profile"

$:
```

/cc @kerwinso @Stackblocks

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -21,11 +21,17 @@ fi
 if [ ! "${CI}" = "true" ] && [ -f ".ruby-version" ] && [ -z "$(rbenv version-name 2>/dev/null)" ]; then
   >&2 echo "==> Installing Ruby…"
   rbenv install --skip-existing
-  which bundle >/dev/null 2>&1  || {
-    gem install bundler
-    rbenv rehash
-  }
 fi
+
+>&2 echo "Using ruby via rbenv:"
+rbenv version
+
+which bundle >/dev/null 2>&1  || {
+  >&2 echo "==> Installing Ruby Bundler gem…"
+  rbenv version
+  gem install bundler
+  rbenv rehash
+}
 
 if [ -f "Gemfile" ]; then
   >&2 echo "==> Installing gem dependencies…"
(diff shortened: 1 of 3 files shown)


kerwinso:
worked on my machine

scripts/bootstrap
```diff
@@ -23,18 +23,28 @@ if [ ! "${CI}" = "true" ] && [ -f ".ruby-version" ] && [ -z "$(rbenv version-nam
...
   rbenv version
-  gem install bundler
+  gem install bundler || {
+    >&2 echo ''
+    >&2 echo ''
+    >&2 echo '--------------- Error -------------'
+    >&2 echo 'Unable to install the bundler.'
+    >&2 echo 'You may need to add the following to ~/.bash_profile :'
+    >&2 echo ''
+    >&2 echo 'eval "$(rbenv init -)"'
```
reedstrm: What's wrong with initializing rbenv in this script? Polluting every single shell with the ruby config. It seems to not be quite as invasive as rvm was back-in-the day, but it still executes things in my shell based on whats in a directory. And when did recipes gain a dependency on ruby, anyway? This is starting to feel like oer.exports all over again …

philschatz:
@reedstrm I removed the requirement of changing `~/.bash_profile` . `rbenv` is now initialized before each call to execute ruby code (similar to how `virtualenv` is activated before each call to execute python code). Hope that address your concern 😄

PR #104, opened 2016-11-11
Diff multiple books

philschatz:
Adds support for `./scripts/diff-book-prepare --all` and `./scripts/diff-book --all` to make regression-testing easier.

Unfortunately, autogenerated `id=""` attributes are still different so not sure how useful the diffs are but you can run the files through a GUI and visually ignore the mismatched uuid's

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,12 @@ Here are the steps to run it:
 ./scripts/diff-book statistics
 
 # Any differences would pop up here
+
+# Optional: Create a diff and give it to a GUI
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -102,6 +102,8 @@ Here are the steps to run it:
 # Any differences would pop up here
```
helenemccarron: Can we add a reminder that it might be useful to save the output in a file (maybe using `&> someFile.txt `) , some non-technical folks may find it useful, especially when using an external tool.

philschatz: ok. I think I addressed this issue. Is there anything else remaining on this?

PR #107, opened 2016-11-23
experiment to get 1 common book.scss file

philschatz:
I tried to pull out all the differences in the `book.scss` files (by diffing them with each other to ignore the common parts) to try and come up with configs for each book with the goal of having ~~1 `book.scss` file for all the books~~ generating rulesets only via _config options_ (distilling the copy/pasta code before it gets unwieldy).

I did not remove the existing stuff in the `_config.scss` files but significant parts of the file could probably be removed.

If there is interest in this approach then I will continue down this rabbit hole.

# [Click here to View Example config and documentation](https://github.com/Connexions/cnx-recipes/blob/experiment-book-configs/rulesets/books/_example/_config.scss)

# Notes

The baked books are equivalent to what is on `master` (I used `./scripts/diff-book` to make sure)

- The idea is that the configs would eventually be something that is generated by the book editor GUI (hence the need to have 1 `book.scss` file)
- Use the "Split Diff" view to see the changes side-by-side
- some commits like https://github.com/Connexions/cnx-recipes/pull/107/commits/ff6a05c7dbe98f1206e75f0d3b6f94d3cbec2af0 show the old `book.scss` file with "hardcoded instructions" compared to loops and conditionals that turn on/off those features (defined in the config file)

As more recipes are customized hopefully this `_book.scss` file will have more `@if` and `@each` loops to customize the book from the config file (I just wrote them quickly, we can revisit and clean up the file)

# Bugs found while combining the book.scss files

- [x] The styleguide was not being generated because `Markup               :` contained spaces and was missing vertical line breaks
- [x] physics books forgot to call `@include modify_retitleHeaders();`
- [ ] `$targetLabels` (which is just copy/pasta'd between all the files) had a bug in it when labeling figures in the appendix
- [ ] `$targetLabels` contained references to `Try It` even though the book did not have any "Try It" notes
- [x] adding in commas for the index was missed in several books (like hsphysics)
- [ ] Trashing solutions was done at the end of Physics instead of the beginning, any reason why?
- [ ] physics books no longer have the divider for Chapter/Appendix titleContent, why?
- should the `$PartType_*` be an object (with all the `map-get`), or should each field be a separate var? For example `$PartType_Exercise_moveTo: $AREA_EOC;` (crude var naming scheme) or `$PartExerciseMoveTo: $AREA_EOC;` (to be more consistent)
- should local vars begin with `_`? I left both options in the PR to provide some examples (note: mixins might need to follow this rule as well)

/cc @helenemccarron & @zroehr for thoughts/feedback


rulesets/books/statistics/_config.scss
```diff
@@ -89,34 +80,14 @@ $setFigureCaption           : ( type: figure, defaultContainer: figcaption, hasC
...
-    selector    : '[data-type="appendix"] :not(figure) > figure',
-    label       : "Table " counter(appendix, upper-alpha) "." counter(figure)
-  )
+  ( selector: '.os-eoc [data-type="exercise"]',               label: "Exercise " counter(chapter) "." counter(exercise) ),
+  ( selector: 'div[data-type="chapter"] .example',            label: "Example " counter(chapter) "." counter(example) ),
+  ( selector: 'div[data-type="chapter"] .try',                label: "Try It" ),
+  ( selector: '[data-type="chapter"] :not(table) > table',    label: "Table " counter(chapter) "." counter(table) ),
+  ( selector: '[data-type="chapter"] :not(figure) > figure',  label: "Figure " counter(chapter) "." counter(figure) ),
+  ( selector: '[data-type="appendix"] :not(table) > table',   label: "Table " counter(appendix, upper-alpha) "." counter(table) ),
+  ( selector: '[data-type="appendix"] :not(figure) > figure', label: "Table " counter(appendix, upper-alpha) "." counter(figure) ),
```
philschatz: @helenemccarron I _think_ this line has a bug in it. It wasn't easily visible until I reformatted the target-labels to be more table-like 😄

philschatz:
[[CTE.04.01] Create Documentation for HTML conventions](https://trello.com/c/lxg5Vbie/9-cte-04-01-create-documentation-for-html-conventions)

reedstrm:
I like the idea of trying to make everything a config diff. The problem you may run into is will this give fine-enough tuning for new features in new books. Will take some thought and playing around with trying to add a new "thing"

philschatz:
I'm not sure what you meant with "fine-enough tuning" (everyone can fine-tune any program if they know assembly but people still choose to live with restrictions imposed by programming languages 😄 ). I think that, in general, config files describe the "higher-order" features in a book which makes the code more maintainable and approachable.

The underlying motivation in this PR was not to make config files per-say, but to remove copy-pasta before too many "really similar but slightly different" files were created.

I think copy/pasta is the reason the template books are fragile and scary to edit (ie the physics PDF templates) and without this "elbow-grease" we end up with more scary & fragile code.

philschatz:
closing this in favor of #164

PR #108, opened 2016-11-23
Fix equation numbering

helenemccarron:
Omit section summary equations
Count equations after collations occurred

TODO:
- [x] Change the section-summary hard coded value to a variable for the book, 
- [x] Create a book wide variable for the section summary since the class varies from book to book and might be used in several places?


philschatz:
Ooh, the coverage decreased; I wonder if it is because one of the snippets needs to be updated (those are the files used to generate the coverage)

helenemccarron:
Which snippets do you think should be updated? Alan tested the results successfully.

philschatz:
Looks like that did the trick 😄  (coverage no longer decreased)

philschatz:
hmm, I wonder if this should be additive (specify that equations should be numbered for each collated page) rather than subtractive (specify which section should not be numbered)

rulesets/books/physics/book.scss
```diff
@@ -37,7 +37,7 @@
   @include count_countExamples(example);
   @include number_numberExamples($exampleTitleContent, $exampleSolutionTitleContent);
   @include count_countEquations(equation);
-  @include number_equations ($equationNumber);
+  @include number_equations ($equationNumber, $sectionSummaryClass);
```
philschatz: @helenemccarron : Should this line have moved down to pass 5 like the AP Physics one did? I noticed while trying to keep #107 in-sync with `master`. 

I'm assuming it was moved to pass 5 because it needed to be numbered _after_ the move, is that correct?

(I added a config piece to #107 named `numberAt:` that could be set to `$NUMBER_NONE`, `$NUMBER_BEFORE_MOVE` and `$NUMBER_AFTER_MOVE`)

To find out if any equations were skipped I opened up `./data/physics-baked.html` and ran the following JS in a browser console.

```js
// Print out all the equation numbers 
// (I wonder if it would be worth making a script that does this for testing. it could do equations, figures, tables, exercises, etc)
document.querySelectorAll('[data-type="equation"] .os-number').forEach((n) => console.log(n.innerText))
```

And here are the gaps (I added `(...)` to skip when numbers were consecutive)

```txt
1.1
1.2
1.3
(...)
1.13
2.1
2.2
2.3
<== MISSING 2.4
2.5
2.6
2.7
2.8
2.9
2.10
<== MISSING 2.11 - 2.14
2.15
(...)
2.25
2.26
<== MISSING 2.27
2.28
(...)
2.75
<== MISSING 2.76 - 2.81
2.82
(...)
2.98
<== MISSING 2.99 - 2.100
2.101
(...)
2.113
```

and when `numberAt:` is set to `$NUMBER_AFTER_MOVE` there is only 1 gap in the numbering (2.10 and 2.11 are missing, not sure why):

```txt
1.1
(...)
1.13
2.1
(...)
2.9 #auto_15a9679c-0422-4712-9a6e-71bad4472a0c@4_import-auto-id1750086
<== MISSING 2.10, 2.11
2.12 #auto_6023b87d-5a28-4910-9e51-ee7fd11c98e1@4_import-auto-id2400875
(...)
2.101
3.1
```

/cc @Stackblocks in case these types of scripts (show all the equation/table/exercise numbers for testing) is useful. Also, I think we could quickly make a script that prints an error if there are gaps in the numbering or if the numbering ever decrements

PR #110, opened 2016-11-30
Biology Ruleset

helenemccarron:
TODO:
units currently do not have an id?

AP biology and HS Physics both need support to units.


philschatz:
Which `_config.scss` and `book.scss` is this based off of?

helenemccarron:
As stated in my message to Zach, this is based on Econ (still mentioned in the config files). I haven't tested the other TOCs yet, that is the next step after I see Zach off so it does not delay his work.

books.txt
```diff
@@ -14,6 +14,7 @@ BOOK_CONFIGS=(
   "physics            physics           031da8d3-b525-429c-80cf-6c8ed997733a"
   "apphysics          ap-physics        8d04a686-d5e8-4798-a27d-c608e4d0e187"
   "hsphysics          hs-physics        051ff00c-34b9-4547-ab03-fe5b1b95cf62"
+  "biology            biology           051ff00c-34b9-4547-ab03-fe5b1b95cf62"
```
philschatz: It looks like this uuid still needs to be updated 😄

philschatz:
I noticed that the style guides on `master` no longer work (because the `Markup:` and `Style guide:` need to have vertical whitespace and the colon needs to be next to the word). Since this is a new books would it make sense to fix it up here?

I went through and fixed the existing books in #107

philschatz:
[[CTEC.01.08] Generate Cooked HTML for AP Biology](https://trello.com/c/rf7j5iqG/27-ctec-01-08-generate-cooked-html-for-ap-biology)

philschatz:
Coverage decreased only because of the extra selectors in the ToC mixins to handle units

PR #111, opened 2016-12-01
Books.txt and server change

helenemccarron:
- added biology from production to `/books.txt`
- added proper collection to HS Physics. 
- added field in `/books.txt` to denote which hostname to use for downloading the book

# Questions

- [ ]  now that each book can specify a hostname, should we remove the option of specifying an environment-var-override option?

### DIFF
--- a/scripts/bake-book-remote
+++ b/scripts/bake-book-remote
@@ -12,8 +12,6 @@ set -e
 # - Invoke the collation via a the python interpreter `import requests; resp = requests.post('http://dev00.cnx.org.cnx.org:6543/contents/031da8d3-b525-429c-80cf-6c8ed997733a/collate-content', headers={'x-api-key': 'b07'}); print(resp)`
 # Originally from: https://gist.github.com/pumazi/93f5ed32cb9e094e5a97415ced480a16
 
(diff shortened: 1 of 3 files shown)


scripts/bake-book-remote
```diff
@@ -99,7 +99,7 @@ do
   ssh ${USER}@${HOST} ${buildBookCmd} || exit 1
 
   >&2 echo "==> Re-baking the book (may take up to 12 minutes)"
-  python -c "import requests; resp = requests.post('http://${HOST}:6543/contents/${uuid}@${bookVersion}/collate-content', headers={'x-api-key': 'cte-dev--sprint-6.7'}); print(resp)" || exit 1
+  python -c "import requests; resp = requests.post('http://${HOST}:6543/contents/${uuid}@${bookVersion}/collate-content',  headers={'x-api-key': 'b07'}); print(resp)" || exit 1
```
philschatz: Why did this header change? Is this something that would need to change for each server?

helenemccarron: The API key is different on the new server. I don't know if that is the case for the next server cc @pumazi and @reedstrm will know better than me.

PR #112, opened 2016-12-02
WiP - HS Physics

helenemccarron:
TODO 
- [x] PZS - create a mixin to allow the creation of a title for a table (used to title the Section Key Terms, `key-terms`  - this might include a generic mixin to title things (see lists in Snap Labs) (almost done, it's a temp fix for this book)

- [x] HM - verify EOC collation structure, numbering scheme from college physics (ask Alana if necessary)
Per Alana: the collated elements do not have to be divided by section, if the teacher’s request this functionality, we can add it later.
Should follow scheme below: 
```- Glossary
- Section Summary
- Section Key Equations
- Chapter Review
    - Concept Items for all modules
    - Critical Thinking for all modules
    - Problems for all modules
    - Performance Task (this happens only once a chapter, should come after concept items, critical thinking, problems etc. )
- Test Prep
    -  Test Prep Multiple Choice
    -  Test Prep Short Answers
    -  Test Prep Extended Response
```
 
- [x] PZS - Snap Labs: ignore number of participants, safety warning etc., that info contained in class is present in content
    -  [x] materials list, titles: Materials (It seems like all we have to do is move the `ul[data-label]` to be a real label on the list - Phil)
    -  [x] warnings list, title: should be Safety Warning
    -  [x] grasp checks title

- [ ] `_target: blank` on videos as well as on simulations
Note: the videos in `ost-video` are currently embedded in an iframe. Sometimes, there is a link to the video before the iFrame, sometimes not. 
For the PDF: how is the iFrame handled? Is it showing or not displaying it at all? We probably can hide it via css but in that case, not all the links will be available.
For TEG: do we need to remove the Iframe? If so, we need to ensure that the link is present in the content, otherwise we will loose the video the same way we would in the PDF.
Possible solution: ensure that all links are mentioned before the iFrame. Ideally, we would have a specific element that would allow to target those links so we can manipulate them.


philschatz:
[[CTEC.01.07] Generate Cooked HTML for High School Physics](https://trello.com/c/tbmiGOH5/16-ctec-01-07-generate-cooked-html-for-high-school-physics)

philschatz:
It seems there already is a "title" for Key Terms, unless I'm misunderstanding something?

image

helenemccarron:
Oooh I missed that. I think the book may be set not to use titles for tables. Is that true for all tables or should they be set to use it? Or is Key Terms an exception?

helenemccarron:
The current application for excluding classes doesn't work, the `trashUnnumberedCaption($type);` mixin is still applying and discarding the caption because it's not within the excluding selector.

I am concerned as well that the excluded tables aren't wrapped in the 'os-table` div anymore, thus possibly  causing more styling to target them down the line.
It may also be difficult to track down why those tables do not follow other tables structure seemingly arbitrarily a few months after the ruleset has been completed or if a different developer were to take over.

I do think an excluding mechanism would be useful, but maybe based on a class or indication from the content rather than an ad hoc decision made within the ruleset.

 I am currently investigating a way to preserve the title and display it above the table. We have may use cases for titles above tables that are currently handled in a hacky way in the content.

philschatz:
ok. I was wondering why `trashUnnumberedCaption($type);` wasn't written to be inside the selector, if there is not a good reason than I can just move it into the selector.

RE `.os-table`: I thought that wrapper element was only being introduced when a table needed to have both a title (like on the top of the table) _and_ a caption because the `<table>` element does not allow both. 

But maybe my understanding is wrong and instead, _all_ tables are always wrapped in an `.os-table`? If so, then should all the classes on the `<table>` be moved up to the `.os-table` wrapper? That way we can select/style it?

philschatz:
**Note:** the only regressions in the commits up to `properly wrap tables, even if they are exceptions (like key-terms)` are in `physics`, `apphysics` and `hsphysics`.

In `physics` and `apphysics`: The only difference is that `Table A_01` is now wrapped in a `.os-table` (soon to be renamed to `data-type="table"`)

philschatz:
I'm not sure if this was introduced with @helenemccarron 's commit but it seems the end of chapter material now lists all the section titles and then for each exercise, it repeats the section title instead of the exercise number

image

helenemccarron:
Yes, that is the point of the commit I made, and it's currently broken so not what the final desired output should be.

rulesets/common/_compose.scss
```diff
@@ -130,10 +130,14 @@
...
 
+//TODO: verify that the empty chapter areas are not displayed, currently working better than previous commit but still displays empty sections. The exercises are empty, why?
+//TODO: add/fix numbering
+//TODO: Optional, try to use isCompoundCompsite instad of loop inside a loop
+
 @mixin processComposites ($compounds) {
   [data-type="chapter"] {
     [data-type="page"] {
-      > [data-type="document-title"] {
+      > h2[data-type="document-title"] {
```
philschatz: Is the `h2` necessary or just descriptive?

helenemccarron:
With the latest commit, the tree structure is cleaner but I haven't found a way to prevent the empty `chapter-area` containers not display if they do not contain any collated elements. Currently, the titles is inserted before the container is discarded as being empty. It seems like it would be achievable by using a different pass but seems convoluted

philschatz:
hmm, could we leave the empty elements in for now? In my quick testing, it seems like easybake does not support the `:empty` selector (even though cssselect2 parses it) /cc @reedstrm 

But maybe a more declarative thing to do would be to only create the end-of-chapter item when there is stuff in the bucket to fill it with. Maybe introducing `::has-pending(bucketName)` which would look something like:

```css
.chapter::after::has-pending(bucketName) {
  /* do the work of creating the EOC element if the bucket is not empty */
}
```

philschatz:
Did I hear that we can remove the Compound Composites from this book? Or maybe it was just wishful thinking?

helenemccarron:
Not compoundComposites, just the sectionSeparated part. I committed that change yesterday, Which, incidentally, you removed when merging the branch :D

rulesets/books/hs-physics/book.scss
```diff
@@ -95,19 +92,22 @@
...
+  @include number_numberEOCExercises($exerciseTitleContent, $exerciseTitleContent);
+  
   @include count_countTables(table);
   @include count_countFigures(figure);
   @include number_setCaptions($setTableCaption, $captionTableNumber, $captionTableNumberAp);
   @include number_setCaptions($setFigureCaption, $captionFigNumber, $captionFigNumberAp);
   @include modify_retitleHeaders();
 }
 :pass(6) {
-  @include modify_trash('.os-index-link-separator:last-child');
```
philschatz: Why was this removed?

rulesets/common/_compose.scss
```diff
@@ -126,7 +126,82 @@
...
+      &::after {
+        class: "os-#{$compositeSource}-container";
+        content: pending(#{$compositeSource}-title) pending(#{$compositeSource}-DESTINATION);
+        move-to: composite-DESTINATION;
+      }
+    }
+  }
+}
+
+@mixin EOCcompoundCompositesPages ($getAllComposites) {
```
philschatz: Should this mixin be prefixed with `compose_` since it is not a local mixin?

rulesets/books/hs-physics/book.scss
helenemccarron: A mistake on my part, restoring.

rulesets/common/_compose.scss
```diff
@@ -126,7 +126,82 @@
 /// @arg {string} $sectionHeaderNode - Curently Unused (because of typo)
 /// @arg {string} $sectionHeaderString - Optional, used when `isIndex` is true
 /// @group compose
-@mixin _prepBookComposite($page, $sectionHeaderNode, $sectionHeaderString) {
+
+
+//TODO: verify that the empty chapter areas are not displayed, currently working better than previous commit but still displays empty sections. The exercises are empty, why?
+//TODO: Optional, try to use isCompoundCompsite instad of loop inside a loop
+
+@mixin processComposites ($compounds) {
```
philschatz: What do you think of renaming `$compounds` to `$allCompoundCompositePages` to clarify what the argument is? I can go through and do the renames 😄

helenemccarron: On it, since I'm in there :)

rulesets/common/_compose.scss
helenemccarron: Yes, fixing. Also renaming `processComposites` to `_processComposites` as well

philschatz:
Btw, there are at least a couple classes that I created that we should probably discuss before merging: `.os-label-title-thingy` and `[data-type="list-wrapper"]`

rulesets/common/_modify.scss
```diff
@@ -178,3 +178,64 @@
...
+      // }
+      // GAAAH the previous does not work. Instead, we have to do the following
+      &::after {
+        container: div;
+        class: "os-label-title-thingy";
+        content: attr(data-label);
+        move-to: bListLabel;
+      }
+      &::outside {
+        attr-data-type: "list-wrapper";
```
helenemccarron: Is there a reason for this to be different from the other `data-type` tags used?

rulesets/common/_number.scss
```diff
@@ -99,40 +99,100 @@
...
-      @include utils_title($labeling, bCaption);
-      @include setCaptionsHelper($type, $defaultContainer, $hasCaption, $hasTitle);
+  // $noTitleContent: (); // For the exceptions, this is the "titleContent" that will be used
+
+  @include utils_checkTypeOptional($exceptionClassNames, list);
+  // These selectors are of the form:
+  // `table:not(.unnumbered):not(.custom1):not(.custom2)` (everything that is not an exception)
+  // `table.unnumbered, table.custom1, table.custom2`     (everything that is an exception)
+  $notAnExceptionSel: "#{$type}";
+  $exceptionsSel: unquote("#{$type}.HACK_DOES_NOT_EXIST");
```
helenemccarron: Can we rename this to something that doesn't look like a huge warning yelling at you? Maybe `doesNotExist`?

rulesets/common/_number.scss
```diff
@@ -99,40 +99,100 @@
...
 
-  .appendix {
-    :not(#{$type}) > #{$type}:not(.unnumbered) {
-      @include utils_title($labelingAp, bCaption);
-      @include setCaptionsHelper($type, $defaultContainer, $hasCaption, $hasTitle);
+  // /* The :not(#{$type}) is here because of subfigures (I think) */
+  [data-type="chapter"],
+  [data-type="chapter"] :not(#{$type}) {
+    > {
+      #{$notAnExceptionSel} {
```
helenemccarron: Can we rename this to `notAnExceptionSelector`, while being a bit longer, it is more consistent with other `selector` variables used.

rulesets/common/_number.scss
```diff
@@ -99,40 +99,100 @@
...
-    }
-    @if ($type == "figure") {
-      figcaption {
-        move-to: trash;
+      // A bug in easybake does not allow 2 selectors with `::outside` to match an element.
+      // in hs-physics the `.key-terms` table is also marked as `.unnumbered`
+      // see https://github.com/Connexions/cnx-easybake/issues/48
+      //
+      // TODO: Maybe these selectors can be simplified to just always call setCaptionsHelper
+      #{$exceptionsSel} {
```
helenemccarron: Can we rename this to `exceptionsSelector`?

rulesets/common/_modify.scss
```diff
@@ -178,3 +178,64 @@
...
+      //
+      //   &::before {
+      //     content: attr(data-label);
+      //     class-add: os-label-title-thingy;
+      //   }
+      // }
+      // GAAAH the previous does not work. Instead, we have to do the following
+      &::after {
+        container: div;
+        class: "os-label-title-thingy";
```
helenemccarron: Maybe just use ` os-label-title`? :D

rulesets/common/_modify.scss
```diff
@@ -178,3 +178,64 @@
...
+        content: attr(data-label);
+        move-to: bTitleLabel;
+      }
+      &::outside {
+        attr-data-type: "exercise-wrapper";
```
helenemccarron: Is there a reason for this not to be a `data-type`?

rulesets/common/_modify.scss
```diff
@@ -178,3 +178,64 @@
     }
   }
 }
+
+/// This moves the data-label attribute off of a list and into a new element so it is visible
+@mixin modify_listMakeLabelVisible() {
```
helenemccarron: Maybe this mixin could be written with a variable for the element type, something similar to the table/figure mixin?

rulesets/common/_modify.scss
philschatz: so far, it seems like tables and figures and things that end up having a wrapper element sometimes end up gaining the `data-type` of the thing they wrap, and sometimes not. I think it would be ok to use `data-type="*-wrapper"` (or `*-container`?) to clarify.

What do you think?

rulesets/common/_number.scss
helenemccarron: I like more time to think about it :) Let's keep it like this for now, maybe with a TODO or FIXME so we do not leave it there forever.

rulesets/common/_modify.scss
philschatz: Yep, it seemed both similar to the exercise wrapping-to-add-a-label mixin as well as different from the `modify_note` mixin which also does this form of wrapping

rulesets/common/_modify.scss
philschatz: In this case it is only the label, there is no title so would it be `os-label`? 💭

rulesets/common/_modify.scss
philschatz: I just want the wrapper/container elements to be marked up the same way; `[data-type="exercise-container] > [data-type="exercise"] { ... }` or `.os-exercise-container > [data-type="exercise"] { ... }` seem like two options

rulesets/common/_modify.scss
helenemccarron: Or it could simply be `os-title`... since, well it is a title. Or is there a reason to keep traces of the label?

philschatz: Lists can contain both a label and a title: http://legacy.cnx.org/eip-help/list but since we do not convert that to HTML nicely yet I do not know what it would look like ; (  

`os-label-title` is more consistent with other places of the code (but requires that we create an additional child element named `class="os-label"`), while `class="os-label"` is more precise about the current conversion format.

rulesets/common/_modify.scss
helenemccarron: We have been using `data-type: myDataType`, I believe Easybake takes your attribute name and converts into a tag, so writing `attr-data-type` produces `<div attr-data-type="">`

rulesets/common/_modify.scss
helenemccarron: I guess we can use `os-label` for now

rulesets/common/_modify.scss
helenemccarron: Yes, that would make sense,  I believe you would need to use `data-type: myDataType` (see note below). Also, container would be more consistent with what already exists in other parts of the book

rulesets/common/_modify.scss
philschatz: oh my, that sounds scary & a bit too "magic"... What if I wanted to create an attribute named `content` ... anywho, updating 😄

reedstrm: There are two cases: data-* and attr-*. For data-*, the declarations are passed throught:

helenemccarron:
So much prettier than HACK_DOES_NOT_EXIST :D

rulesets/common/_modify.scss
philschatz: @reedstrm 💭  when I did `attr-data-type: "foo";` it created an attribute named `attr-data-type="foo"` ... the `attr-*` was not stripped and yet the declaration did not begin with `data-` so I was confused as to what was happening...

PR #115, opened 2016-12-06
update servers to point to tea

philschatz:
This should help keep everyone up-to-date on which server to talk to for getting/rebaking books 😄 

Also, the title could be renamed to something like "move dev00 books to tea" instead of waiting until all books are moved to tea


helenemccarron:
I have renamed books to specifically mention wether they are TEA books or not.

PR #118, opened 2016-12-07
Add AP physics 2 for TEA to books.txt

helenemccarron:
And use the TEA data for TEA AP Physics 1


helenemccarron:
@philschatz Can you please test `bake-book-remotely`? I think it might not work with the `tea.cnx.org`

philschatz:
`./scripts/bake-book-remote` should now work again for all books

PR #119, opened 2016-12-08
fix misc bugs found in #107

philschatz:
These are fixes for items found in #107

- [x] The styleguide was not being generated because `Markup               :` contained spaces and was missing vertical line breaks
- [x] physics books forgot to call `@include modify_retitleHeaders();`
- [x] `$targetLabels` (which is just copy/pasta'd between all the files) had a bug in it when labeling figures in the appendix
- [x] `$targetLabels` contained references to `Try It` even though the book did not have any "Try It" notes
- [x] adding in commas for the index was missed in several books (like hsphysics)

# Maybe included?

These are inconsistencies I found but may not need to be fixed in this PR

- [ ] Trashing solutions was done at the end of Physics instead of the beginning, any reason why?
- [ ] physics books no longer have the `divider:` element for Chapter/Appendix `titleContent`, why?


helenemccarron:
I don't remember why trashing the solutions are happening last but I can't think of any valid reason why it would be at the end. It might be a random act of randomness :)

Re: dividers, those where removed when the decision to handle purely visual elements such as dividers with CSS rather than doing in the rulesets. I have removed them everywhere.

PR #120, opened 2016-12-09
removes the lingering glossary title when they are collated

philschatz:
In books the glossary items are collated at the end of the Chapter/Book. This problem was noticed while working on the biology book.

# New Screenshot:

image

---

# Old Screenshot:
image

---
/cc @zroehr


philschatz:
closing in favor of #121

PR #121, opened 2016-12-09
fix trashing of duplicate glossary title

zroehr:
There was an extra glossary title that wasn't getting trashed due to the glossary title selector changing. Fixed selector in `rulesets/common/_compose.scss` to trash the extra title.


philschatz:
@helenemccarron do you know why the old code removed only the title but left the `<section>` in the HTML?

helenemccarron:
No, I reviewed Zach's fix and it seemed right but I cannot remember why we left the section in there, there seems to be no need for it now.

PR #125, opened 2016-12-16
add alternatives for iframe

philschatz:
This wraps an `<iframe>` with something like:

```html
<div data-type="alternatives" 
     class="os-has-iframe os-has-link">

  <a class="os-is-link" 
     href="https://www.youtube.com/embed/rStL7niR7gs"
     target="_window">

     link
  </a>

  <iframe class="os-is-iframe"
          src="https://www.youtube.com/embed/rStL7niR7gs"
          width="560" height="315">
  </iframe>
</div>
```

# Rationale

I wanted to make sure it was possible to use CSS to display the best alternative. Here is an example:

```less
[data-type="alternatives"] {
  // By default, use the browser-friendliest alternative
  // Note: This should always be included in the CSS, and we should test that various overrides work
  > :not(:last-child) { display: none; }

  // Add this if you prefer iframes
  &.os-has-iframe > :not(.os-is-iframe) { display: none; }

  // Add this if you prefer links (ie PDF)
  &.os-has-link > :not(.os-is-link) { display: none; }

  /// Add this if you prefer an image when both an iframe and a link are available
  &.os-has-iframe.os-has-picture > :not(.os-is-picture) { display: none; }
}
```

and example HTML:

```html
<div data-type="alternatives" class="os-has-iframe os-has-link os-has-picture">
  <a class="os-is-link">link</a>
  <a class="os-is-picture"><img src="..."/></a>
  <iframe class="os-is-iframe"></iframe>
</div>
```

# TODO

- [x] regression-test with all the other books
- [ ] decide if the link needs a wrapping `<p>` (and move the class to the `<p>`)
- [ ] decide if custom link text is necessary or nice-to-have (it will be more annoying to maintain)


rulesets/common/_modify.scss
```diff
@@ -239,3 +241,48 @@
...
+    //     }
+    //   }
+    // }
+    //
+    // The previous does not work. Instead, we have to do the following (& use a bucket)
+    > iframe {
+      class: attr(class) "os-is-iframe";
+      &::after {
+        container: a;
+        class: "os-is-link";
```
helenemccarron: Does it need to be `os-is-link`? Could it be `os-link`?

rulesets/common/_modify.scss
```diff
@@ -239,3 +241,48 @@
...
+    //       tag-name-set: "a";
+    //       attrs-set: href attr(src), target "_window";
+    //       class-add: "os-is-link";
+    //       content: "link";
+    //     }
+    //   }
+    // }
+    //
+    // The previous does not work. Instead, we have to do the following (& use a bucket)
+    > iframe {
```
helenemccarron: You could guard it here, maybe? Using `> iframe:not([data-type="alternatives"])`

rulesets/common/_modify.scss
philschatz: 🤔  I wanted to make the relationship between `.os-has-link` and this element clear so people could write CSS like:

```less
[data-type="alternatives"] {
  // By default, use the browser-friendliest alternative
  // Note: This should always be included in the CSS, and we should test that various overrides work
  > :not(:last-child) { display: none; }

  // Add this if you prefer iframes
  &.os-has-iframe > :not(.os-is-iframe) { display: none; }

  // Add this if you prefer links (ie PDF)
  &.os-has-link > :not(.os-is-link) { display: none; }

  /// Add this if you prefer an image when both an iframe and a link are available
  &.os-has-iframe.os-has-picture > :not(.os-is-picture) { display: none; }
}
```

and example HTML:

```html
<div data-type="alternatives" class="os-has-iframe os-has-link os-has-picture">
  <a class="os-is-link">link</a>
  <a class="os-is-picture"><img src="..."/></a>
  <iframe class="os-is-iframe"></iframe>
</div>
```

🤔  maybe the PR should use `*-embedded` instead of `*-iframe` to better-reflect the alternative, not the DOM node?

rulesets/common/_modify.scss
philschatz: Unfortunately the `[data-type="alternatives"]` element is the parent of the iFrame so I do not think that would work ; (

rulesets/mixins/_modify.scss
```diff
@@ -239,3 +241,48 @@
...
+    //
+    // The previous does not work. Instead, we have to do the following (& use a bucket)
+    > iframe {
+      class: attr(class) "os-is-iframe";
+      &::after {
+        container: a;
+        class: "os-is-link";
+        attr-target: "_window";
+        attr-href: attr(src);
+        content: "link";
```
helenemccarron: Maybe we can use something a bit more descriptive yet global enough, something like: "Click to view content"

rulesets/mixins/_modify.scss
```diff
@@ -180,20 +180,21 @@
 }
 
 /// This moves the data-label attribute off of a list and into a new element so it is visible
+/// @group modify
 @mixin modify_wrapLists() {
   ol,
   ul {
     &[data-label]:not([data-label=""]) {
       // &::outside {
```
helenemccarron: Can we get rid of the commented code since it doesn't work?

rulesets/mixins/_modify.scss
```diff
@@ -239,3 +241,48 @@
...
 }
+
+/// This wraps every iframe with an element that has classes on it
+/// saying which alternatives are present,
+/// and adds a link that opens in a new window.
+/// @group modify
+@mixin modify_wrapIframesWithAlternatives() {
+  // just a guard to make sure we did not already convert the iframe
+  :not([data-type="alternatives"]) {
+    // > iframe {
```
helenemccarron: Same here, can we get rid of comments?

helenemccarron:
I think wrapping the link in something isn't a bad idea.  In my opinion, it should look good without css, but I don't think we need custom text, as stated in my of review comments. I also remember Alina mentioning the use of a custom image for the pdf to replace the embedded content, so it might be a non issue for the PDF.

rulesets/mixins/_modify.scss
philschatz: Any thoughts on the "readability" of the commented code? I included it as an example to discuss changes to easybake and as a little bit of documentation since I felt that the actual code was not as easy to grok what was going on.

Related Links: 

- [Holiday work: Order-independent attribute modifiers](https://github.com/openstax/napkin-notes/blob/master/phil/2016-easybake-holiday.md#order-independent-attribute-modifiers)
- https://github.com/Connexions/cnx-easybake/pull/60 (PR to implement part of ^^^)

philschatz:
I _think_ the `.os-has-picture` in the **Rationale** section at the top of this PR should address the case of PDFs wanting to show an image instead of just a link to the sim, no?

helenemccarron:
Yes, it does.

PR #135, opened 2016-12-22
add ap biology rulesets

zroehr:


philschatz:
Does this still work for biology? It seems like this introduces several other pages ("Review Questions" and "Critical Thinking Questions", and "Test Prep for AP Courses").

philschatz:
What still needs to be done for AP Biology? Can you add it to the PR body or link to the Trello ticket?

philschatz:
[[CTEC.01.08] Generate Cooked HTML for AP Biology](https://trello.com/c/rf7j5iqG/27-ctec-01-08-generate-cooked-html-for-ap-biology)

philschatz:
This removes the solutions for biology as well. Is that desired behavior?

zroehr:
@philschatz Alana said we need the ruleset to show in biology. I will need to separate out the rulesets to achieve this.

philschatz:
It seems the styleguide still needs updates (as noted in the coverage misses)

helenemccarron:
As well as remove the missing exercises, I believe? @reedstrm did the code that marks the missing exercises in the raw html make it to tea.cnx.org?

rulesets/books/ap-biology/_config.scss
```diff
@@ -239,19 +239,19 @@ $targetLabels: (
     label       : "Example " counter(chapter) "." counter(example)
   ),
   (
-    selector    : '[data-type="chapter"] :not(table) > table',
+    selector    : '[data-type="chapter"] :not(table) > table:not(.unnumbered)',
```
philschatz: Just making sure... the plan is to wait on merging this PR until the `.unnumbered` code is fixed properly, right? Or, is it OK if the other books may have regressions as a result of this code change?

PR #138, opened 2017-01-03
add python3 as a dependency

philschatz:
It seems that over the holidays something changed that ended up with this message when running `./scripts/setup` on OSX:

```
==> Installing Homebrew dependencies…
Error: No such file or directory - /usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions
```

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,2 +1,3 @@
+brew "python3"
 brew "rbenv"
 brew "lcov"


reedstrm:
Hmm, not sure we're python3 compatible w/ easybake.

philschatz:
Hmm, so if we wanted to ensure easybake ran with python2 would we need to pass an argument to https://github.com/Connexions/cnx-recipes/blob/master/scripts/setup#L19 or to the `pip install virtualenv` line?

It _seems_ that python3 may be a new dependency for Homebrew but I'm not sure. @Stackblocks and I have both seen the following when using homebrew (not just when installing new packages):

```
Error: No such file or directory - /usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions
```

It also occurred when running `brew bundle check` which is used to see if the packages listed in `Brewfile` are already installed

philschatz:
This seems to no longer be an issue. Waiting a bit to see if anyone responds

PR #140, opened 2017-01-05
Fix simulation links to open in a new tab

### LINKED ISSUE #139: `target="_blank"` does not seem to work anymore
This is added to some links (I think external ones like simulations)

philschatz:
fixes #139 

I tried to see how the code was broken but it did not seem to be broken (see the styleguide test). Maybe I am missing something?

### DIFF
--- a/rulesets/mixins/styleguide/page.link.html
+++ b/rulesets/mixins/styleguide/page.link.html
@@ -4,7 +4,7 @@
     <!-- Markup starts here -->
     <p id="id-para">
       Link to a simulation:
-      <a href="foo-archive.cnx.org/specials/bam">link</a>
+      <a class="os-interactive-link" href="foo-archive.cnx.org/specials/bam">link</a>
     </p>
 
     <p>
@@ -24,6 +24,10 @@
 
     <figure id="id-figure">
       <div data-type="title">Figure Title</div>
+    </figure>
+
+    <figure>
+      <a class="os-interactive-link" href="foo-archive.cnx.org/specials/bam">link inside a figure</a>
     </figure>
 
   </div>
--- a/rulesets/mixins/styleguide/page.link.html
+++ b/rulesets/mixins/styleguide/page.link.html
@@ -27,7 +27,7 @@
     </figure>
 
     <figure>
-      <a class="os-interactive-link" href="foo-archive.cnx.org/specials/bam">link inside a figure</a>
+      <a class="os-interactive-link" href="https://phet.colorado.edu/sims/html/forces-and-motion-basics/latest/forces-and-motion-basics_en.html">link inside a figure</a>
     </figure>
 
   </div>
--- a/rulesets/mixins/styleguide/page.link.html
+++ b/rulesets/mixins/styleguide/page.link.html
@@ -1,0 +1,30 @@
+<div data-type="chapter">
+  <div data-type="page">
+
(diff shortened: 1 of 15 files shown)


rulesets/mixins/styleguide/page.link.html
```diff
@@ -0,0 +1,30 @@
...
+      Link to a figure:
+      <a href="#id-figure">link</a>
+    </p>
+
+    <figure id="id-figure">
+      <div data-type="title">Figure Title</div>
+    </figure>
+
+  </div>
+</div>
```
philschatz: The baked HTML looks like this:

```html
<div data-type="chapter">
  <div data-type="page">
    <!-- Markup starts here -->
    <p id="id-para">
      Link to a simulation:
      <a href="foo-archive.cnx.org/specials/bam" target="_blank">link</a>
    </p>
    <p>
      Link to a website:
      <a href="https://cnx.org">link</a>
    </p>
    <p>
      Link to a paragraph:
      <a href="#id-para">link</a>
    </p>
    <p>
      Link to a figure:
      <a href="#id-figure">link</a>
    </p>
    <div class="os-figure">
  <figure id="id-figure">
      <div data-type="title">Figure Title</div>
    </figure><div class="os-caption-container"><span class="os-title-label">Figure </span><span class="os-number">1.1</span></div></div>
  </div>
</div>
```

helenemccarron:
When I bake HS physics from the tip of masters, this is what I get

`<a class="os-embed os-interactive-link" href="https://archive.cnx.org/specials/2c7acb3c-2fbd-11e5-b2d9-e7f92291703c/collision-lab/">Click here for the simulation</a>` 

This link should have `target=_blank` based on the ruleset (`:pass(0) a[href*="archive.cnx.org/specials/"] {
  attr-target: "_blank"; }`) for HS physics

philschatz:
(grumble, easybake working unexpectedly) ... I thought I could write a little test to check whether it works but apparently not ; ( . I found the problem @helenemccarron  found and am trying to create a minimal test case so I can:

1. include it in this PR
2. report a bug to `cnx-easybake`

Also, it seems the current strategy for detecting links will not be sufficient. I found links like this:

```html
<a class="os-embed os-interactive-link" href="https://phet.colorado.edu/sims/html/forces-and-motion-basics/latest/forces-and-motion-basics_en.html">Click here for the simulation</a>
```

Why does the baked HTML need to have the `target="_blank"` in it? Is that something specific to TEA? Because I think

1. webview would prefer to show the sim as a bootstrap modal
2. PDFs do not need that extra attribute
3. I thought TEA preferred that the sim is embedded in the page (as long as the URL is `https://`)

helenemccarron:
Yes, this is for TEA. We had discussed a long time ago that they couldn't embed the simulations because they couldn't be resized and didn't fit their body's dimensions, so we decided to use `target=_blank` as an alternative. 

Also, during our last call, we agreed that we didn't have an efficient way to provide them with something other than a link to the content. They agreed to use only the links, as long as they were all `https://`, which we confirmed.

philschatz:
I found out why the links were not being matched... it's because something in the `modify_note` was causing the links (which are inside a Note) to not be matched & updated. 

This was a problem for other things as well (wrapping figures and lists) so I moved `modify_note` into a separate pass.

Gonna try to make a smaller test case for `cnx-easybake` & create an Issue

PR #142, opened 2017-01-13
update hs-physics to match PDF

philschatz:
- [x] Exercises in "Check Your Understanding" need to be numbered
  - "Check your Understanding" does not have a special class name on it so it is not selectable
  - It seems all exercises in the main content that are not inside an Example need to be numbered
- [x] remove "Table 2.5" from the HTML (item in Key Equations), each equation has this table label
- [x] rename "Glossary" to "Key Terms"
- [x] add/remove section separators as needed
- [x] remove `#.1 Introduction` from end-of-chapter section titles (see screenshot below)
- [x] remove empty section titles from end-of-chapter collated material (see screenshot below)
- [x] Find out why some exercises are missing in the HTML
  - "Rhianna and Logan start at the same point and walk due north" (Ch2.12 in PDF)
  - "A position vs. time graph of a frog swimming across a pond has two distinct straight-line sections" (Ch2.6)

/cc @Stackblocks for updating w/ more details that I missed 😄


philschatz:
I added logic to number exercises except when either of these are true:

- they are a **child** of an example (cannot do descendants easily in css to match the PDF XSLT precisely)
- they have the class `.unnumbered`

This caused the "Check Your Understanding" exercises to render properly but I noticed 2 other problems:

1. the `titleContent` items need to be specified backwards, not sure why
1. Adding this code into the other books causes regressions; is that desired?
  - For example, in `statistics`: "Try It" notes with an exercise are now numbered

reedstrm:
re: the last point - missing exercises. The PDF was built from content that has been updated on staging2. Is that where the recipe is getting its content?

philschatz:
@reedstrm HSPhysics is being pulled from [tea.cnx.org](https://github.com/Connexions/cnx-recipes/blob/master/books.txt#L16) and the build-single-html-file script is being [run on the same server](https://github.com/Connexions/cnx-recipes/blob/master/scripts/fetch-html#L62-L67); not sure where that script is configured to pull from though

philschatz:
By enabling `sectionSeparated: true` I get the following:

image

It seems there are 2 problems:

1. the Introduction Page is being counted as a Book Section
2. empty sections are still rendering the title

RE the 2nd point, is there a way in easybake to not render a section (and its title) if no elements are moved to it? /cc @reedstrm

philschatz:
The tables in "Key Equations" do not have an `.unnumbered` class on them so we would need to suppress numbering some other way /cc @helenemccarron

helenemccarron:
Per Alina, numbered tables will never occur within the Key Equation section so simply excluding tables from being numbered in that section would be an option. The table in that section is structural, so it might be possible to select it very specifically, which theoretically would allow for tables inside that particular table to be numbered should the need arise.

philschatz:
I think that the numbering exclusion for tables in "Key Equations" depends on #147 being fixed first. But even that may not be enough because that only allows for `$exceptionClassNames`, not "selectors that this table is in".

helenemccarron:
The introduction being numbered was a content issue. With the proper class `.introduction`, introductions are no longer numbered

rulesets/books/hs-physics/_config.scss
```diff
@@ -359,11 +363,11 @@ $targetLabels: (
     label       : "Example " counter(chapter) "." counter(example)
   ),
   (
-    selector    : '[data-type="chapter"] :not(table) > table',
+    selector    : '[data-type="chapter"] :not(table) > table:not(.unnumbered)',
```
philschatz: Just making sure... the plan is to wait on merging this PR until the `.unnumbered` code is fixed properly, right?

philschatz:
Also, I noticed that this change creates a regression in statistics (& probably more books).

Each `[data-type="problem"]` now has the following children in it:

```
<span class="os-title-label">Exercise</span><span class="os-number">1</span>
```

Also in statistics: the numbering of figures has decreased by 1. I noticed it in (`./scripts/diff-book --all`)

```
(...) <span class="os-title-label">Table </span><span class="os-number">1.2</span><span data-type="title" class="os-title">Fall Term 2007 (Census day)</span></div></div>
(...) Look at <a 
  class="autogenerated-content" 
  href="#auto_098e1a26-e612-4449-a45e-80fa23feba02@12_ch01_mod02_fig003">
  Figure 1.4
</a> and <a 
  class="autogenerated-content" 
  href="#auto_098e1a26-e612-4449-a45e-80fa23feba02@12_ch01_mod02_fig004">
  Figure 1.5
</a> (...)
```
Previously they were `Figure 1.5` and `Figure 1.6`

---

Are those changes desired for all the books?

rulesets/books/hs-physics/_config.scss
helenemccarron: Sure, we can wait.

helenemccarron:
I haven't got around fixing the numbering for the other books, I just pushed the css so Travis would stop yelling at me. I am investigating a better solution to ` selector    : '[data-type="chapter"] :not(table) > table:not(.unnumbered)'` so that each book doesn't have to be updated

helenemccarron:
As for the `[data-type="problem"]` containing an exercise numbering, I think it's because `number_numberContentExercises` targets everything that is not in an example.

scripts/compile-books
```diff
@@ -23,5 +23,6 @@ do
 
   >&2 echo "==> Generating ${cssFile}"
 
-  bundle exec sass ${sassFile} ${cssFile} || exit 1
+  # bundle exec sass ${sassFile} ${cssFile} || exit 1
```
philschatz: Hmm, was this not working for you? It seems like commenting this out causes Travis to fail. 

If it does not work could you try running `gem install sass` and then trying again? If that works then maybe `gem install sass` could be added to `./scripts/setup` explicitly (in another PR)

helenemccarron: As a note, I am able to `./scripts/compile-books` works

rulesets/books/statistics/_config.scss
```diff
@@ -65,6 +69,9 @@ $tryTitleContent            : ( title-label: "Try It ", number: counter(chapter)
 
 $calcTitleContent           : ( title-label: "Using the TI-83, 83+, 84, 84+ Calculator" );
 
+
+$collabNote   : ( label : "Collaborative Excerise", selector : collab, );
```
philschatz: Is there an HTML snippet for this in the styleguide?

rulesets/books/hs-physics/book.scss
```diff
@@ -37,6 +37,15 @@
...
   @include modify_note($watchPhysicsNote);
   @include modify_note($virtualPhysicsNote);
+
+  //the metadata has been moved to the content, remove it from the metadata
+  @include modify_trash('[data-type="metadata"] > [data-type="description"]');
+  @include modify_trash('div.missing-exercise');
+  @include modify_trash('[data-type="solution"]');
+  @include modify_addUnnumbered ('.key-equations', 'table');
+  @include modify_addUnnumbered ('.worked-example', '[data-type="exercise"]');
+  @include modify_addUnnumbered ('.grasp-check', '&[data-type="exercise"]');
```
philschatz: Why does this one have an `&` while the rest do not?

rulesets/books/hs-physics/book.scss
```diff
@@ -37,6 +37,15 @@
   @include modify_note($linksPhysicsNote);
   @include modify_note($watchPhysicsNote);
   @include modify_note($virtualPhysicsNote);
+
+  //the metadata has been moved to the content, remove it from the metadata
+  @include modify_trash('[data-type="metadata"] > [data-type="description"]');
+  @include modify_trash('div.missing-exercise');
+  @include modify_trash('[data-type="solution"]');
```
philschatz: Is this here because all solutions are discarded in the baked file?

rulesets/books/hs-physics/book.scss
```diff
@@ -37,6 +37,15 @@
   @include modify_note($linksPhysicsNote);
   @include modify_note($watchPhysicsNote);
   @include modify_note($virtualPhysicsNote);
+
+  //the metadata has been moved to the content, remove it from the metadata
+  @include modify_trash('[data-type="metadata"] > [data-type="description"]');
+  @include modify_trash('div.missing-exercise');
```
philschatz: Is this specific to hs-physics or would it also work if it was included in all `books.scss` files?

rulesets/books/ap-physics/book.scss
```diff
@@ -60,6 +55,10 @@
   @include compose_createChapterComposites($chapterCompositePages, sectionHeaderNode);
   //create a chapter outline
   @include compose_chapterOutline($objectivesTitleContent);
+
+  @include modify_addUnnumbered ('[data-type="example"]', '[data-type="exercise"]');
+  @include modify_addUnnumbered ('[data-element-type="check-understanding"]', '&[data-type="exercise"]');
```
philschatz: Do you know why this attribute is `data-element-type` and not `class`? Is that something from the CNXML? (like maybe the CNXML is `<example type="check-understanding">`?)

rulesets/books/statistics/book.scss
```diff
@@ -24,6 +25,7 @@
   @include modify_wrapLists(); // no dependencies
   @include modify_wrapIframesWithAlternatives(); // no dependencies
   @include modify_trash('.try [data-type="solution"]');
+  @include modify_GenericNote();
```
philschatz: Would all the books still bake well if this was included in all the `book.scss` files? Or is it something that needs to be enabled just for a couple books? (I noticed it was not included in all the `book.scss` files)

helenemccarron: I am handling those individually as I go through the books. I _think_ all of them will need it but I haven't confirmed yet.

helenemccarron: Also, if you want to roll that back, that's fine, I have them implemented in the branch I am currently working.

philschatz: Nah, I was just curious. I'm just trying to keep the `book.scss` files consistent 😄  (except for book-specific settings of course)

helenemccarron: 👌

rulesets/books/ap-physics/book.scss
helenemccarron: Yes, this is an exercise with `[data-element-type="check-understanding"]` and no class.

rulesets/books/hs-physics/book.scss
helenemccarron: It works in hs physics and ap bio, which are both using it. Other books do not need this mixin because they aren't fetching exercises

rulesets/books/hs-physics/book.scss
helenemccarron: yes, most exercises do not have a solution but some do and they were displaying inappropriately

rulesets/books/hs-physics/book.scss
helenemccarron: Because this is an exercise with the class `.grasp check` as opposed to an exercise in a Worked Example section.

rulesets/books/hs-physics/book.scss
helenemccarron: However, I need to check that this rule isn't too broad and doesn't delete inline solutions.

rulesets/books/statistics/_config.scss
helenemccarron: I don't think so, I'll create one.

rulesets/books/hs-physics/book.scss
philschatz: ah ok. Do you mind if I change it to `@include modify_addUnnumbered ('.grasp-check[data-type="exercise"]', null);` (and I'd update the mixin to handle a null).

I get a little worried when `&` 's are in a string rather than in the `.scss` file directly (too much sass magic to follow along with)

helenemccarron: Sure, I didn't like it either!

philschatz: ok pushed these changes.

PR #144, opened 2017-01-18
replace the `os-` prefix with `ox-`

### LINKED ISSUE #123: Change `os-` class prefix to `ox-`
This makes it clearer that these are **O**pensta**X** classes and not OpenSource or OperatingSystem. Also lines up with other cases where we need to abbreviate OpenStax.

philschatz:
Classes in the HTML are currently prefixed with `os-` and should be `ox-` to use the abbreviation used in other places (makes it clear that this is OpenStax, not OpenSource, or OperatingSystem, etc)

Should probably discuss before merging.

I did not replace **all** references to `os-*` because some (like `.os-interactive` are classes in the content)

Fixes #123


philschatz:
hmm, I'm not sure why coverage would decrease. maybe the search-and-replace accidentally caused something to no longer match

helenemccarron:
@phil I seem to remember that we would close this in favor of redoing the work from master to avoid a very gnarly merge. Is that still something we want to do?

philschatz:
Closing in favor of redoing this work as part of #271

PR #146, opened 2017-01-19
provides log file when baking fails

philschatz:
This creates a log file for easier diagnosis and a message letting the user know baking failed.

Depends on https://github.com/Connexions/cnx-easybake/issues/61 before being merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
 /books/data/
 /data/*-raw.html
 /data/*-baked.html
+/data/*-baked.html.log
 # The following file is used for diffing 2 versions of a book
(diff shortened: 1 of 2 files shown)


helenemccarron:
Does Connexions/cnx-easybake#61 need to be merged or can I merge this now?

philschatz:
If the easybake work is not done then the log file will be generated but you won't see the nice `==> ERROR: baking failed. Check ${bakingLogFile}` message. So, if that's still useful then merge, otherwise I'd say leave this open so there's more pressure to get the easybake fix done

helenemccarron:
Let's leave it here for now :)

PR #147, opened 2017-01-24
:bug: fix the `.unnumberedtable` selector to be `table.unnumbered`

philschatz:
The CSS contained selectors containing `... > .unnumberedtable > ...` which was never being matched. This changes the selectors to be `... > table.unnumbered > ...`

It causes easybake to crash so I am looking into that.


helenemccarron:
Where is the `. unnumberedtable` class coming from? It only appears in the css and I can't find it in the `.scss`.

helenemccarron:
Do we still need this pull request? I think we solved the unnumbered issues in #142

philschatz:
yep! closing

PR #151, opened 2017-02-16
Add ordering info into style guide lines

reedstrm:
Not 100% happy with how this is done, but I guess mass-updates is what
sed -i is for, right?

### DIFF
--- a/scripts/generate-guide
+++ b/scripts/generate-guide
@@ -31,8 +31,8 @@ if [ ! -e ${styleguideDir} ]
 then
   mkdir ${styleguideDir} || exit 1
 fi
-sass ${sassFile} ${cssFile} || exit 1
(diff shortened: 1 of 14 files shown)


scripts/generate-guide
```diff
@@ -31,8 +31,8 @@ if [ ! -e ${styleguideDir} ]
 then
   mkdir ${styleguideDir} || exit 1
 fi
-bundle exec sass ${sassFile} ${cssFile} || exit 1
-bundle exec sass ${sassFile} ${bookCssFile} || exit 1
+sass ${sassFile} ${cssFile} || exit 1
```
philschatz: I noticed in a previous PR that you commented this out as well. It seems to work for @helenemccarron, myself, and (most importantly) Travis. I'd like to help see why this does not seem to work for you.

PR #154, opened 2017-02-22
Books various fixes

helenemccarron:
Fixes for TEA HS Stats, TEA AP Macro Econ, TEA AP Micro Econ, TEA AP Physics 1 & 2

Remaining known issues:
**all books:**
- [ ] ❓ remove the cnx-pi and other collation instructions?
- [ ] ❓ intro in the wrong place (needs to be before learning objectives), is that a problem?
- [x] chap-shortid blank string (Ross working on it)

**Stats:**
- [x] two missing appendices classes [Bey working on it]
- [x] example solution has a link to itself. Excluding examples from `@mixin utils_hasSolution() ` is a real pain :-(

**Macro & Micro Econ :**
- [x] missing all appendices classes [Bey working on it]

**AP Physics 1 & 2:**
- [ ] ❓ data-element-type="check-understanding” > on an exercise, title doesn’t match notes

TODO:
- [x] add missing notes to the guide
 - Calculator note
 - Finger Note
 - Stats Lab Note

- [x] clean up css a bit


rulesets/books/statistics/styleguide/book.composite.homework.html
```diff
@@ -0,0 +1,38 @@
+<!-- Markup starts here -->
+<div data-type="composite-page"  class="os-eoc os-homework-container">
```
philschatz: Is this the raw HTML or the baked HTML? 

I think this class is added as part of baking but this file file should contain the raw HTML of a chapter to make sure:

1. content editors know what the raw HTML for a homework section should look like
2. the CSS selectors in `statistics.css` turn from red to green (run `./scripts/test` and then `open ./coverage/index.html`)

scripts/compile-books
```diff
@@ -23,5 +23,6 @@ do
 
   >&2 echo "==> Generating ${cssFile}"
 
-  bundle exec sass ${sassFile} ${cssFile} || exit 1
+  # bundle exec sass ${sassFile} ${cssFile} || exit 1
```
philschatz: This commit is probably causing Travis to fail

helenemccarron: Yup, will fix next push

rulesets/books/statistics/styleguide/book.composite.homework.html
helenemccarron: duh, you're right.

rulesets/books/statistics/book.scss
```diff
@@ -21,11 +21,11 @@
   @include modify_titlePreface();
   @include modify_spanWrapTitles();
   @include modify_simLinkTarget();
+  @include modify_GenericNote();
```
philschatz: ooh, why did this need to move up in the statistics book? Would the other books also work if it was moved to the same location?

philschatz: ooh, why did this need to move up in the statistics book? Would the other books also work if it was moved to the same location?

rulesets/books/statistics/book.scss
```diff
@@ -42,7 +42,9 @@
   @include number_numberExamples($exampleTitleContent, $exampleSolutionTitleContent);
   @include count_countNote(Try, "try");
   @include number_numberNote("try", $tryTitleContent);
-  @include number_numberNote("calculator", $calcTitleContent);
+  @include modify_note($calculatorNote);
+  @include modify_note($fingerNote);
+  @include modify_note($statsLabNote);
```
philschatz: Why are these in `:pass(2)` while other books have the `modify_note` code in `:pass(0)` (or `:pass(1)` in the case of nested notes)?

rulesets/books/statistics/styleguide/book.composite.solutions.html
```diff
@@ -0,0 +1,40 @@
...
+      <p>“Number of times per week” is what type of data?</p>
+      <p>
+        <span data-type="list" data-list-type="labeled-item" data-display="inline">
+          <span data-type="item">a. qualitative</span>
+          <span data-type="item">b. quantitative discrete</span>
+          <span data-type="item">c. quantitative continuous</span>
+        </span>
+      </p>
+    </div>
+    <div data-type="solution">b</div>
```
philschatz: Is `book.composite.solutions` specific to the statistics book or should this snippet be common to all books? The reason I ask is that there do not seem to be any specific class attributes in this snippet

rulesets/books/statistics/styleguide/page.example.html
```diff
@@ -0,0 +1,12 @@
+<div data-type="chapter">
+  <div data-type="page">
+
+    <!-- Markup starts here -->
+    <div data-type="note" class="try">
```
philschatz: I'm a little confused about the name of this file. It seems like the markup is a `<cnxml:note>` with `class="try"` rather than a generic `<cnxml:example>` (as the filename would imply)

Would `page.note.try.html` work? (Although, I could see that might cause the styleguide ToC to be confusing; maybe that was the motivation for the name?)

See `page.note.calculator.html` and `page.note.collab.html` below in the diff

---

Actually, is this just a copy/paste of `page.note.try.html` below in the diff?

rulesets/mixins/_modify.scss
```diff
@@ -122,8 +122,10 @@
 /// Modifies note feature to add a note title and optionally wrap the body content
 /// @arg {map} $noteContent - map containing `$selector`, `$label` and `$isBodyWrapped`
 /// @group modify
+
+//modify GenericNote: in stats, many notes have  different title, the only way to keep those was to pass noteTitle to the title. However, the titles are inconsistent. It might be better to define the title via variables like any other notes and fix the content.
```
philschatz: Thanks! Should this text be included in the sassdoc comment format above? It seems like it might be useful to read in the generated docs

rulesets/mixins/_modify.scss
```diff
@@ -184,10 +197,33 @@
...
+          > .os-title {
+            class: "os-title";
+            container: h4;
+          }
+        }
+      }
+    }
+  }
+ [data-type="composite-page"].os-eob {
+    >[class*=-chapter-area] {
```
philschatz: hmm, could/should we add a better attribute/additional-classname to the elements so this selector would not require a `*=` to match? Also, any reason `-chapter-area` isn't wrapped in quotes like the other attribute electors?

rulesets/mixins/_utils.scss
```diff
@@ -44,7 +44,7 @@
 /// @arg {Text} $toCounter [null] - The counter object used to automatically ID the element to be linked to if it doesn't already have an id. If null, the link href will be to an already existing id of the linked-to element
 /// @group utils
 @mixin utils_hasSolution() {
-  [data-type="exercise"] {
+  :not(.example) > [data-type="exercise"] {
```
philschatz: Could this use `:not([data-type="example"]) > ...` instead of the class name? I'm hoping that the class names will be removed from the raw HTML some time soon (to reduce confusion).

rulesets/mixins/_modify.scss
helenemccarron: Well, each selector is composed of the section name + chapter-area (ie `critical-thinking-chapter-area`). What would be better? Just use `chapter-area`, matching the current class in the PDF?

rulesets/books/statistics/styleguide/page.example.html
helenemccarron: Yup, this was for the examples. Pushed the proper markup for it.

rulesets/books/statistics/styleguide/book.composite.solutions.html
helenemccarron: It is one of the rare books that do not have an EOB answer key but solutions handled like other EOC collations. It would be common to all books that display solutions at the end of a chapter.

rulesets/books/statistics/book.scss
helenemccarron: No particular reason, that's where the notes used to be. I can move it.

rulesets/books/statistics/book.scss
helenemccarron: No, I think I just matched it with hs-physics when comparing them. The order doesn't matter in this case.

rulesets/books/statistics/styleguide/book.composite.solutions.html
philschatz: If this same snippet was included in all books then would each styleguide show where the solutions are collated? For example, this snippet in stats would show the solutions at the end of a chapter while the same snippet for econ would show the solutions in an "Answer Key" at the back of a book?

rulesets/mixins/_modify.scss
philschatz: Yep, can you make an `os-chapter-area` one?

PR #167, opened 2017-03-22
rename `recipes` to `rulesets`

philschatz:
replaces lingering references to the term `recipes` with `rulesets`


philschatz:
scratch that. I think recipes makes more sense because it's in-line with `raw` and `baked`

PR #164, opened 2017-03-15
Refactor Rulesets to use config files

philschatz:
This goes through each book and converts it to a config file so future refactoring can be easier.

# Rulesets to convert

- [x] `ap-biology` os-exercises have additional whitespace at the end of the class attribute
- [x] `ap-physics` spacing after the table/figure number were inconsistent. also, this is the only book that removes *all* solutions
- [x] `biology` (this was missing the cnx-pi-removal code)
- [x] `economics`
- [x] `hs-physics` equations had an extra space in their numbers (ie `8. 30`). Other things still have that extra space (see `TODO:` blocks)
- [x] `physics`
- [x] `statistics`
- [x] `_example` (documentation)

# TODO

- [x] move all the copy/pasta'd `book.scss` files into one `_generator.scss` file
- [x] remove `$Config_ChapterCompoundCompositePages` in favor of just `$Config_ChapterCompositePages`
- [x] bring code coverage back up by having config settings to remove unused CSS in the generated files
- [x] verify whitespace is correct

# TODO In a subsequent PR

Since this PR does not touch the mixins, the following will be done in a separate PR (once this one is merged)

- [ ] change fields like `isGlossary` and `isIndex` to be an enum (ie `pageType: $PAGE_GLOSSARY`) so it is clear that you can only choose one
- [ ] prefix config options with `is` to clarify they are booleans
- [ ] change 1-word config options to 2+ words
- [ ] change `selector:` to `className:`

replaces #107.

PR #172, opened 2017-03-28
Config fields

philschatz:
Followup to #164 which renames fields that are used in mixins.

**Prime Directive:** One goal of this PR is to _not change the CSS files at all_

- [x] change fields like `isGlossary` and `isIndex` to be an enum (ie `pageType: $PAGE_GLOSSARY`) so it is clear that you can only choose one
- [x] convert `sectionSeparated` and `chapterSeparated` to be an enum
- [x] prefix config options with `is` to clarify they are booleans
- [x] Document the config file fields in a file
- [x] remove extra spaces around parentheses (maps) (or always have spaces, we can discuss 😄 )
- [x] replace constructed class names with something easier to grep
  - [x] prefix `divider:` and `number:` with `os-` so they are easier to grep
  - Fixing all of these violates the _Prime Directive_
- [x] Rename the `$Config_ChapterTitleContent` things so documentation is hierarchical
- [ ] either remove `captionType:` or combine `$Config_SetTableCaption` and `$Config_SetFigureCaption` into a list (:+1:)
- [x] change 1-word config options to 2+ words
  - [x] change `source:` to `className:`
  - [ ] change `name:` to something else (This is the only one that still remains)

# [Click Here to see the documentation](https://github.com/Connexions/cnx-rulesets/blob/config-fields/rulesets/README.md)

# Future Work

Here are items that would change the CSS files so they are left for another PR

- [ ] change `os-${className}-container` maybe to `os-container-${className}` or something (discuss?)


rulesets/books/_example/_config.scss
```diff
@@ -88,22 +71,22 @@ $_compoundTestPrep: (
...
-  (source: "summary",         sectionSeparated: true,                               name: "Chapter Review"),
-  (source: "formula-review",  sectionSeparated: true,                               name: "Formula Review"),
-  (source: "practice",        sectionSeparated: true, hasSolutions: true,           name: "Practice"),
-  (source: "bring-together-exercises",                hasSolutions: true,           name: "Bringing It Together: Exercises"),
-  (source: "free-response",   sectionSeparated: true, hasSolutions: true,           name: "Homework"),
-  (source: "bring-together-homework",                 hasSolutions: true,           name: "Bringing It Together: Homework"),
-  (source: "references",      sectionSeparated: true,                               name: "References"),
-  (source: "solutions",       compoundComposite: true, moveSolutionsTo: $AREA_EOC,  name: "Solutions"),
+  (className: "glossary",        specialPageType: $PAGE_GLOSSARY, sortBy: "dl > dt",   name: "Key Terms"),
+  (className: "summary",         groupBy: $GROUP_SECTION,                               name: "Chapter Review"),
```
helenemccarron: I find this a bit confusing, it's very close to `group-by`.

rulesets/books/_example/_config.scss
```diff
@@ -13,70 +13,53 @@
 // They are local vars that are used below.
 // NOTE: The order of each value matters; it is the order in which these
 // elements will occur in the generated text.
-$_exerciseTitleContent       : ( divider: ".", number: counter(exercise) );
-$_exampleTitleContent        : ( title-label: "Example ", number: counter(chapter) "." counter(example) );
-$_exampleSolutionTitleContent: ( title-label: "Solution ", number: counter(chapter) "." counter(example) );
-$_tryTitleContent            : ( title-label: "Try It ", number: counter(chapter) "." counter(example) ); // try it notes are numbered based on the example preceeding them
-$_equationNumber             : ( title-label: "Eqn", number: counter(equation) );
+$_exerciseTitleContent       : (os-divider: ".", os-number: counter(exercise));
```
helenemccarron: Is it necessary to prefix those? What is the rational behind it?

philschatz: `prefix divider: and number: with os- so they are easier to grep` In general, string interpolation (dynamically concatenating strings together) makes it difficult to find in the future. 

As a developer, when I see something like `os-divider` in the CSS that I need to fix, the first thing I do is search the source for `os-divider`; I don't think that "oh, `os-` and `divider` might be combined together so I should search for `divider`". 

And searching for `divider` ends up with many more search results.

rulesets/books/_example/_config.scss
philschatz: Yeah, I didn't like the name of this either. What would be better name for this field? (preferably still concise)

helenemccarron: clusterBy? bunchBy? combinedBy? I don't know, some synonym of group by :-)

philschatz: Renamed! I used `clusterBy` 😄

PR #174, opened 2017-04-05
TEG fixes

helenemccarron:
#### All pages not labeled as data-type="composite-page" need at least one section with an h3 title before the content. Anything before that section will not be imported. 
  - [x] Wrap the learning objectives and add a title via the ruleset.
  - [x] Verify that all books modules start with learning objectives and or chapter outline. 
  - [x] Come up with titles in the content if a module doesn't contain learning objectives

#### Appendices (and Prefaces) do not have the necessary nesting to be imported. They need a binder title (h1), a resource title (h2) and a section title (h3)
  - [x] Repeat the title 3 times with the proper nesting via the ruleset

#### End of book pages need a binder title and a resource title
 - [x] Create a resource title via the ruleset by repeating the binder title 
 - [x] Create composite chapters for appendices

#### Separate TEA rulesets from OSX rulesets
 - [x] Isolate TEA specific ad hoc requirements to avoid confusion


philschatz:
This diff is very large and it is unclear what changed (it looks like the `books/*` directories were just copied and renamed). Is there a more-concise way of describing the difference between `economics` and `TEAeconomics` for example?

rulesets/mixins/_compose.scss
```diff
@@ -550,16 +550,24 @@
...
+        class: "os-chapter-objective";
+        content: content();
+        attr-href: "#" string(outlinePageId);
+        move-to: chapOutlineLink;
+      }
     }
     &::after {
-      class: "os-chapter-objective";
-      content: nodes(outlineSectionTitle) pending(outlineAbstract);
+      class: "os-lo";
```
philschatz: is `os-lo` "Openstax Learning Objective"? If so, could it be `os-learning-objective`? 

Also, is the plan still to rename `os-` to `ox-` (in another PR)?

rulesets/mixins/_modify.scss
```diff
@@ -34,13 +34,21 @@
...
 /// @group modify
 @mixin modify_suppressURI() {
-  div.os-eoc, div.os-eob  {
+  div.os-eoc, div.os-eob {
     [data-type="cnx-archive-uri"] {
       move-to: trash;
     }
   }
+/// Trashes `[data-type="cnx-archive-uri"]` in composite-chapters
+/// @group modify
```
philschatz: Does sassdoc pick this up?

rulesets/mixins/_modify.scss
```diff
@@ -99,15 +107,15 @@
 /// @arg {Text} $bookMetadata - The node set that contains the metadata
 /// @group modify
 @mixin modify_compositeMetadata($bookMetadata) {
-  [data-type="composite-page"] {
+  [data-type="composite-page"], [data-type="composite-chapter"]  {
```
philschatz: can you put these selectors on separate lines?

That makes diffs easier to read (when only 1 selector changes) and the `,` is easy to miss when scanning so it looks like `[data-type="composite-page"] [data-type="composite-chapter"] {`

rulesets/mixins/_modify.scss
```diff
@@ -268,15 +276,35 @@
       }
     }
   }
+  [data-type="composite-chapter"] {
+    > section, > [data-type="page"] {
```
philschatz: same here, at a glance it looks like `> section > [data-type="page"] {`

rulesets/mixins/_toc.scss
```diff
@@ -52,7 +52,7 @@
         container: li;
       }
     }
-    > div[data-type='chapter'] {
+    > div[data-type='chapter'], > div[data-type='composite-chapter'] {
```
philschatz: same here, ` > div[data-type='chapter'] > div[data-type='composite-chapter']`

Do we actually have both, a `chapter` and a `composite-chapter`? (I forget, what is a `composite-chapter`?)

rulesets/mixins/_compose.scss
```diff
@@ -550,16 +550,24 @@
...
-    }
-    [data-type="abstract"] {
-      move-to: outlineAbstract;
+     // node-set: outlineSectionTitle;
+      string-set: outlinePageId attr(id);
+      &::after {
+        container: a;
+        class: "os-chapter-objective";
+        content: content();
+        attr-href: "#" string(outlinePageId);
```
helenemccarron: Yes it does :D

rulesets/mixins/_toc.scss
reedstrm: Yup, that's what I demoed - creating composite-chapters!

rulesets/books/_TEAspecific.scss
```diff
@@ -0,0 +1,90 @@
...
+      }
+      &::outside{
+        data-type: "composite-chapter";
+        content: nodes(eobTitle) pending(referencesTitle) content();
+      }
+    }
+  }
+}
+
+//remove the index
```
helenemccarron: This is really just for TEA, removing the index and solutions _after_ they were created. They need to remain in the base file, I don't have a way to disable them after they have been defined, so I decided to simply remove them in the TEA books.

philschatz: Would something like this work?

https://github.com/Connexions/cnx-rulesets/compare/TEG-fixes...phil-no-index

it removes the Index Page entirely (haven't tested yet) which reduces the CSS file size (since there is no Index that needs to generate)

However, I agree that there is probably no need to spend too much time on getting the TEG-specific stuff to work properly

helenemccarron: If you want, I don't think it really matters how we do it as long as the page isn't there (for TEG). I'm also not too concerned about the css file (once again, for TEG)  but easier is always better.

rulesets/mixins/_compose.scss
helenemccarron: Yes, fixed. For the `ox-` prefix, yes, I thought we discussed starting over with a new PR since  merging would have been really complicated with the existing PR. But maybe I'm wrong :-)

rulesets/books/_TEAspecific.scss
```diff
@@ -0,0 +1,90 @@
...
+        move-to: AppendixTitleSpans;
+      }
+      &::after {
+        container: h1;
+        data-type: "document-title";
+        content: pending(AppendixTitleSpans);
+        move-to: appendicesTitle;
+      }
+      &::after {
+        data-type: composite-chapter;
```
philschatz: Should this be `composite-appendix` rather than `composite-chapter`? Just wondering because it seems like the Appendix is _usually_ a Page but in this case it is multiple Pages

philschatz:
I'm going to try to move each of the `composite-chapter` selectors inside an `@if $hasCompositeAppendixes` in order to not increase the CSS file size (and lower the code coverage) because I think it's less work than adding HTML snippets into the styleguides (to increase the code coverage).

@helenemccarron thoughts?

philschatz:
Also, do all TEG books have this composite-appendix or just some? I don't see anything in the styleguides denoting which books have this structure

philschatz:
Also, when generating `./scripts/bake-book TEAapphysics` I get the following error (not sure why)

```
==> Validating ./data/TEAapphysics-baked.xhtml
cnxepub ERROR Skipping title overrides -mismatched numbers: parent: 17, titles: 18
  File "/Users/phil/github/cnx/cnx-rulesets/venv/bin/cnx-epub-validate-collated", line 9, in <module>
    load_entry_point('cnx-epub==0.12.0', 'console_scripts', 'cnx-epub-validate-collated')()
  File "/Users/phil/github/cnx/cnx-rulesets/venv/lib/python2.7/site-packages/cnxepub/scripts/collated_single_html/main.py", line 49, in main
    binder = reconstitute(args.collated_html)
  File "/Users/phil/github/cnx/cnx-rulesets/venv/lib/python2.7/site-packages/cnxepub/collation.py", line 51, in reconstitute
    return adapt_single_html(xhtml)
  File "/Users/phil/github/cnx/cnx-rulesets/venv/lib/python2.7/site-packages/cnxepub/adapters.py", line 356, in adapt_single_html
    _adapt_single_html_tree(binder, body[0], nav_tree)
  File "/Users/phil/github/cnx/cnx-rulesets/venv/lib/python2.7/site-packages/cnxepub/adapters.py", line 528, in _adapt_single_html_tree
    raise AdaptationError
cnxepub.adapters.AdaptationError
```

helenemccarron:
@philschatz Not all TEG books have appendices. The composite appendices are for TEG only, we do not want to implement those in our books. There is also a possibility that we will need the same treatment for prefaces in TEG only books. 

Composite chapters will be used for EOB sections that need to be split in pages, such as Answer Keys or Index to avoid very long pages.

reedstrm:
The adaptation error indicates that the TOC and the actual XHTML tree disagree on the number of chapters. The license_url not found is from not running the most recent version of cnx-archive, which copies license info down the tree.

philschatz:
@reedstrm hmm, I re-ran `./script/setup` which I thought would re-grab the python dependencies (I did that before posting the error messages). I'll try deleting the `./venv/` dir and running it again.

philschatz:
oh, derp. I forgot to rename `[data-type="composite-chapter"]` to `[data-type="composite-appendix"]`. I will make a separate PR

PR #187, opened 2017-05-09
Dev fixes

philschatz:
- updates `pip` when `scripts/setup` is run
- uses the globally installed `sass` when the rbenv one is not found

Address the stacktrace in https://github.com/Connexions/cnx-rulesets/pull/174#issuecomment-299943157

### DIFF
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -17,13 +17,22 @@ if [ -f "Brewfile" ] && [ "$(uname -s)" = "Darwin" ]; then
   }
 fi
 
+
+>&2 echo "==> Installing Node package dependencies"
+>&2 echo "Using node: $(node --version)"
+>&2 echo "Using npm: $(npm --version)"
+npm install || exit 1
+
 # Note: Skip rbenv for Travis because it already has ruby
 if [ ! "${CI}" = "true" ] && [ -f ".ruby-version" ]; then
   if [[ -n $(which rbenv) ]]; then
 
     if [ -z "$(rbenv version-name 2>/dev/null)" ]; then
       >&2 echo "==> Installing Ruby…"
-      rbenv install --skip-existing
+      rbenv install --skip-existing || {
+        echo "WARNING: 'rbenv install' errored. I am assuming you know what you are doing (and have sass installed locally) so continuing"
+        exit 0
+      }
     fi
   else
     >&2 echo "WARNING: Skipping rbenv. Using local ruby"
@@ -54,9 +63,3 @@ if [ -f "Gemfile" ]; then
     bundle install --path vendor/gems --quiet --without production
   }
 fi
-
-
->&2 echo "==> Installing Node package dependencies"
->&2 echo "Using node: $(node --version)"
->&2 echo "Using npm: $(npm --version)"
-npm install || exit 1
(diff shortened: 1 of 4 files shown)


scripts/bootstrap
```diff
@@ -17,13 +17,22 @@ if [ -f "Brewfile" ] && [ "$(uname -s)" = "Darwin" ]; then
...
   if [[ -n $(which rbenv) ]]; then
 
     if [ -z "$(rbenv version-name 2>/dev/null)" ]; then
       >&2 echo "==> Installing Ruby…"
-      rbenv install --skip-existing
+      rbenv install --skip-existing || {
+        echo "WARNING: 'rbenv install' errored. I am assuming you know what you are doing (and have sass installed locally) so continuing"
+        exit 0
+      }
     fi
```
reedstrm: don't just assume, test for it: `sass --version` , then set `SASS='sass'` (vs `SASS='bundle exec sass'`) and use `$SASS` in the compile script.

scripts/compile-books
```diff
@@ -23,5 +23,17 @@ do
 
   >&2 echo "==> Generating ${cssFile}"
 
-  bundle exec sass ${sassFile} ${cssFile} || exit 1
+  # Try running sass locally (using rbenv). If that does not work then try it globally
```
reedstrm: Oh, you do it here. Hmm. It repeats for every single book. I guess it _works_, it's just not very pretty/nice.

scripts/generate-guide
```diff
@@ -39,8 +39,23 @@ if [ ! -e ${styleguideDir} ]
...
-bundle exec sass ${sassFile} ${bookCssFile} || exit 1
+
+
+# Try running sass locally (using rbenv). If that does not work then try it globally
+if [[ $(bundle exec sass --version) ]]
+then
+  sassBin="bundle exec sass"
+elif [[ $(sass --version) ]]
+then
+  >&2 echo "WARNING: Using global sass"
```
reedstrm: Not a WARNING, I'm doing it on purpose.

philschatz:
@reedstrm I added the log messages so it would be clear when debugging which version of sass was being used (either the local version defined in https://github.com/Connexions/cnx-rulesets/blob/master/Gemfile or a system version which may be different than the one defined in the `Gemfile`).

What if I change the warning to show which version of SASS is running if it is using the system version of sass? Something like `INFO: Using system Sass 3.4.23 (Selective Steve)`

philschatz:
oic, you created a separate script called "sassBin" that "uses the version of sass as configured in ./scripts/setup" ... interesting.

philschatz:
@reedstrm hmm, I don't think your commit address the 1st point in the previous comment. It's useful to know in the console log that the version of SASS is not the one in the gemfile. What do you think of having the console print out that a system version of SASS is being run?

reedstrm:
sure, just do it once per invocation of top-level script, not for every book

philschatz:
ok, I think @71d53eb adds the message once per invocation of the script. I'll revert to that tomorrow 😄

reedstrm:
@phil  I rolled it back for you, and added one tweak - suppress the 'not found' error messages.

PR #189, opened 2017-05-11
rename composite-chapter to composite-appendix

philschatz:
oops, forgot to rename that attribute to clarify that the thing being generated is an appendix, not a chapter.

refs #174

### DIFF
--- a/rulesets/books/_TEAspecific.scss
+++ b/rulesets/books/_TEAspecific.scss
@@ -33,7 +33,7 @@
         move-to: appendicesTitle;
       }
       &::after {
-        data-type: composite-chapter;
+        data-type: composite-appendix;
         data-uuid-key: appendicesChapter;
         // attr-id: uuid() "@1";
         content: pending(appendicesTitle) pending(chapterAppendices);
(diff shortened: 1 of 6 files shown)


reedstrm:
https://github.com/Connexions/cnx-epub/blob/master/cnxepub/adapters.py#L488

reedstrm:
The place for the subtype-specific info is probably the data-uuid-key, anyway.

philschatz:
oh, in that case I should change the name of the config variable that adds these things in. My understanding was that some of the books wanted an Appendix for each Unit (or multiple pages inside an appendix) instead of having 1-page Appendixes at the end of a book.

See https://github.com/Connexions/cnx-rulesets/pull/189/files#diff-153bd716f5fa0585a5cfebf41a4e187bR17 for the variable name and the code that creates `composite-chapter` uses other variables with the word "Appendix" in them.

philschatz:
Closing this because a "chapter" is a structural type (it can only have Page children) while a "Chapter" is a presentational element (like an Appendix or Preface).

So, I think it makes sense to keep the CSS as `composite-chapter` while keeping the config field `$Config_hasCompositeAppendixes`

/cc @helenemccarron

PR #191, opened 2017-05-15
rename scripts dir to script to be more consistent

philschatz:
... with other repositories.

See https://githubengineering.com/scripts-to-rule-them-all/ for more details


reedstrm:
I'm sad the standard is the rather arbitrary singular - the directory contains more than one script.

philschatz:
Yeah, I'm a bit torn too. But it seems like project root directories tend to be singular:

- `bin`
- `lib`
- `test`
- `src`
- `usr`
- `vendor`

so I dunno.

PR #193, opened 2017-05-17
Stats Changes

tomjw64:
Changes to stats books to resolve differences between Webview/PDF


philschatz:
Travis is complaining because [TEAstatistics was not recompiled](https://travis-ci.org/Connexions/cnx-rulesets/builds/233401240#L3202).

You can run `./scripts/compile-books` to regenerate all the CSS files before pushing to make sure all of them are regenerated

PR #197, opened 2017-05-18
Stats lab fixes

helenemccarron:
https://github.com/Connexions/cnx-rulesets/issues/183

### DIFF
--- a/rulesets/mixins/_modify.scss
+++ b/rulesets/mixins/_modify.scss
@@ -201,9 +201,13 @@
   $labelText: map-get($noteContent, labelText);
   $titleContent: map-get($noteContent, titleContent);
   $isBodyWrapped: map-get($noteContent, isBodyWrapped);
+
   @include validate_type($selector, string);
   @include validate_typeOptional($titleContent, map);
   @include validate_typeOptional($labelText, string);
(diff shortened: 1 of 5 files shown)


rulesets/mixins/_modify.scss
```diff
@@ -199,13 +199,20 @@
 @mixin modify_note($noteContent) {
   $selector: map-get($noteContent, selector);
   $labelText: map-get($noteContent, labelText);
+  $titleContent: map-get($noteContent, titleContent);
   $isBodyWrapped: map-get($noteContent, isBodyWrapped);
+  @include validate_type($selector, string);
+  @include validate_typeOptional($titleContent, map);
+  @include validate_typeOptional($labelText, string);
+
```
philschatz: Thanks for adding the validation! If you can pass in _either_ `$titleContent` or `$labelText` (but not both) could you add something like:

```sass
// Check that exactly one option is provided
@include validate_exists($titleContent or $labelText);
@include validate_exists(not($titleContent and $labelText));
```

Not sure if that works but :crossed_fingers:

helenemccarron: It worked!

PR #200, opened 2017-05-19
Biology Sample Fixes for TEA

tomjw64:
Make biology perfect for demonstration


rulesets/books/_TEAspecific.scss
```diff
@@ -65,19 +63,64 @@
...
-//         content: nodes(eobTitle) pending(referencesTitle) content();
-//       }
-//     }
-//   }
-// }
+@if $Config_TEAspecificHasCompositePreface {
+  :pass(1) {
+    body {
+      .preface {
+        move-to: prefaceWaka;
```
philschatz: What does `Waka` mean? I noticed it in several places

rulesets/books/_generator.scss
```diff
@@ -2,6 +2,11 @@
 
 $_symbolsSectionTitle: "Symbols";
 
+//Currently, none of the books have composite chapters
+//TODO: add a switch when our books have composite chapters, define logic/behavior for that switch. This simply could be in the _config files for all books or set based on the presence of at least one type of composite chapter (appendix, multiple page answer key etc.).
```
philschatz: I agree. Is there a way to infer this from some other config options? Like replace `$Config_hasCompositeChapter` with a local var in `_generator.scss` that says something like:

```scss
$_hasCompositeChapter: $Config_hasCompositePreface or $Config_hasCompositeChapter or $Config_hasCompositeAppendix
```

or even something like:

```scss
$Config_Composites_Preface: (title: "Preface");
$Config_Composites_Chapter: (title: "Appendix"); // for unit-level appendixes
```

and then the code that references `$Config_hasCompositeChapter` could be switched to check if☝️ are non-null

rulesets/books/_TEAspecific.scss
helenemccarron: oops, I need to replace it by its final value, it's my test data 🗡

rulesets/books/_generator.scss
helenemccarron: I was thinking about something like that. I don't know if I want to implement it now without a concrete example in the current books. I fear that if we base our approach on the the TEA data, we may skew our thinking rather than plan for what is best for Openstax.

philschatz:
@helenemccarron do you frequently use the word `waka` for testing? I use `PHIL` since the code should never refer to me (especially in all-caps). If so, I'll add a test at the end of `./script/test` that makes sure those strings do not appear in the code anywhere 😄 

it should be as easy as something like:

```
if [[ $(grep -R "waka" . | wc -l) != 0 ]]; then
  echo "Debugging string found!"
  exit 1
fi
```

rulesets/books/_TEAspecific.scss
```diff
@@ -119,15 +118,34 @@
       }
     }
   }
+  :pass(7) {
+    [data-type="composite-chapter"] {
+      //HACK: force the metadata title
```
philschatz: ooh, should all the books use this format (preface is `<h2>` and preface-container is `<h3>`?

rulesets/books/_TEAspecific.scss
```diff
@@ -119,15 +118,34 @@
...
 
-
 //remove the index
 :pass(6) {
     @include modify_trash(".os-index-container");
     @include modify_trash(".os-solutions-container");
     @include utils_clearTrash();
 }
-:pass(7) {
+:pass(8) {
```
philschatz: Can you include a comment above this line describing which mixins in `:pass(7)` affect this? That way we know why `retitleHeaders` needs to run later?

helenemccarron:
@philschatz re: preface headers, I initially thought that no, because only the TEA books need their preface rewritten but thinking about it, it does not hurt to ensure consistency in preface as well.

PR #209, opened 2017-05-22
Fix College Bio Issues

tomjw64:
This PR will address college bio issues, including those that carry over to AP Bio, and the TEG Biology Sample

### DIFF
--- a/books.txt
+++ b/books.txt
@@ -9,8 +9,8 @@
 # - additionalStyleguideDirs : These specify which additional directories should be included when generating the styleguide
 
 BOOK_CONFIGS=(
-#  "_example           _example          example-uuid                            tea.cnx.org          mixins/"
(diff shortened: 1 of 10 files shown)


helenemccarron:
This is a bit of a thorny situation. The "new" note is the same as the AP biology but the class used is different. So there are identical notes with two different classes. We can either mention the class in the html snippet in both places (or in the one that doesn't make sense since the class doesn't match the note name) or create two identical snippets with two different names to match the two classes. I _think_ the content team would like to change the mismatching class when bulk content edits will be available.

philschatz:
@helenemccarron I think that until content changes the HTML the snippets should reflect what the CSS code targets.

If the notes are almost the same (except for class names) then we should have different snippets in each recipe.

PR #216, opened 2017-05-23
fix patch reporting

philschatz:
because it raises false positives.

@tomjw64 noticed this in #209

and it seems to be because SCSS files are included in the diff but are not marked as hits.

See https://docs.codecov.io/docs/commit-status#section-patch-status

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,5 +2,10 @@ comment:
   layout: "header, diff, changes, uncovered, tree"
   behavior: default
 
+coverage:
+  status:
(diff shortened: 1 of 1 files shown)


philschatz:
actually, maybe a better approach is to exclude the SCSS files from codecov. there is an option to exclude paths.... Thanks @RoyEJohnson !

philschatz:
hmm, could we whitelist only the `/rulesets/output` directory in the `.codecov.yml` file instead?

Issue #182, opened 2017-05-01
Introductory Statistics => Note inside TI-83... feature causing title to appear in wrong place.


Stackblocks:
# Scenario

“Using the TI-83…” feature containing a nested note is displaying the feature title above the nested note and below the start of the content.

# Reproducibility Details

* **Impact area**: Content
*  **User**: n/a
*  **URL / Location**: https://dev00.cnx.org/contents/MBiUQmmY@18.13:L3GARsYT@5/47-Poisson-Distribution
* **How many times were you able to reproduce the issue?** Static

*  **Steps to reproduce**:

n/a

* **Images or screenshots Included:** Yes  (_attach below_)

![image](https://cloud.githubusercontent.com/assets/12678235/25588575/7c615150-2e6e-11e7-8967-ba0855f96df3.png)

# Environment:

* **Instance**: Dev00
* **OS**: macOS
* **Browser**: Chrome

# Impact

* **Severity**: High
* **Frequency of occurrence**: Common

tomjw64:
This issue likely appears because of a class="note" occurring within another class="note"

tomjw64:
Addressed in PR #229

Stackblocks:
Appears resolved after re-bake with bio-fixes ruleset.
![image](https://user-images.githubusercontent.com/12678235/26839697-6d4b3076-4aa9-11e7-8e73-64e1ab525bb9.png)

Issue #219, opened 2017-05-25
College Biology => Everyday Connection feature says "Everyday Connection" twice


kerwinso:
Similar to #201. In webview, the Everyday Connection feature has a title "Everyday Connection", and also a subtitle "Everyday Connection", in addition to the actual subtitle. 

Webview URL: 
https://dev00.cnx.org/contents/GFy_h8cu@10.61:MnC6GuJi@7/65-Enzymes

Webview on the left, PDF on the right:
![image](https://cloud.githubusercontent.com/assets/11298165/26457705/bbf94bee-4136-11e7-89cf-f7d06966959f.png)

tomjw64:
Addressed in PR #229

kerwinso:
Verified fix on https://dev.cnx.org/contents/GFy_h8cu@10.61:MnC6GuJi@7/65-Enzymes:

![image](https://cloud.githubusercontent.com/assets/11298165/26792046/b620dfe2-49de-11e7-92c6-67d17f10d6d6.png)

Issue #218, opened 2017-05-25
College Biology => Scientific Connection feature should be Scientific Method Connection


kerwinso:
The feature 'Scientific Connection' should be re-titled 'Scientific Method Connection' to match the PDF.

Webview URL (Ch. 6.3): https://dev00.cnx.org/contents/GFy_h8cu@10.61:HZOoT9Pk@7/63-The-Laws-of-Thermodynamics

Webview on the left, PDF on the right:
![image](https://cloud.githubusercontent.com/assets/11298165/26457372/7c4e0152-4135-11e7-9547-e4f42c7779b8.png)

kerwinso:
Weirdly, this does not seem to affect all Scientific Method Connection features. This one looks correct ([Ch. 10.2](https://dev00.cnx.org/contents/GFy_h8cu@10.61:1tJ55Ot6@7/102-The-Cell-Cycle)):

![image](https://cloud.githubusercontent.com/assets/11298165/26470194/7477305c-4162-11e7-9ec9-3105299f1baa.png)

tomjw64:
Should be addressed in PR #229

kerwinso:
Verified fix on https://dev.cnx.org/contents/GFy_h8cu@10.61:yNlSxj0E@6/192-Population-Genetics:

![image](https://cloud.githubusercontent.com/assets/11298165/26797160/9387c578-49f1-11e7-9ccc-57850947cb3a.png)

Issue #204, opened 2017-05-22
College Biology => Career Connection feature says "Career Connection" twice


kerwinso:
Similar to #201. In Webview, the Career Connection feature has a title "Career Connection Note", and also a subtitle "Career Connection", in addition to the actual subtitle. 

Webview URL: 
https://dev00.cnx.org/contents/GFy_h8cu@10.61:gNLp76vu@13/12-Themes-and-Concepts-of-Biol

Webview on the left, PDF on the right:
![image](https://cloud.githubusercontent.com/assets/11298165/26320512/a3e51416-3ee8-11e7-8481-237c887fee9c.png)

kerwinso:
In section 7.1 (https://dev00.cnx.org/contents/GFy_h8cu@10.61:dfq-SF-8@8/71-Energy-in-Living-Systems), the subtitle actually says "Career Connections" (note the plural as opposed to other instances of the singular). How very odd. 

![image](https://cloud.githubusercontent.com/assets/11298165/26459627/b66a9af6-413c-11e7-99d0-9c601ce97303.png)

tomjw64:
Addressed in PR #229

kerwinso:
Verified fixes:
https://dev00.cnx.org/contents/GFy_h8cu@10.61:gNLp76vu@13/12-Themes-and-Concepts-of-Biol
![image](https://cloud.githubusercontent.com/assets/11298165/26792870/6dbdf714-49e1-11e7-9f43-e895ff675110.png)

https://dev00.cnx.org/contents/GFy_h8cu@10.61:dfq-SF-8@8/71-Energy-in-Living-Systems
![image](https://cloud.githubusercontent.com/assets/11298165/26792862/69300944-49e1-11e7-8a06-aff47f5b627a.png)

PR #229, opened 2017-05-31
Fix Notes

### LINKED ISSUE #163: AP bio notes do not appear to be titled properly
`ap-science-practices` for example does not have a title generated.

tomjw64:
Stop duplicate note titles from occurring. Nest notes titles and bodies properly. Also properly downgrade h3->h4, h4->h5 for notes nested in other notes.

Fixes #163 , #201 , #203 , #204 , #218 , #219


tomjw64:
Code coverage tests now pass. Ready for review.

tomjw64:
Fixes Issues #163 , #201 , #203 , #204 , #218 , #219

tomjw64:
Also fixes Issue #182

tomjw64:
@philschatz @helenemccarron Is it reasonable to merge this now? Does the final issue need to be closed? I posted an image of the issue resolved in the issue comments.

rulesets/mixins/_modify.scss
```diff
@@ -262,16 +334,16 @@
...
-          class: "os-subtitle";
-          container: h4;
-        }
-        [data-type="note"] {
-          > .os-title {
-            class: "os-title";
-            container: h4;
-          }
-        }
+        // .os-note-body > [data-type="title"] {
```
helenemccarron: If it's not needed anymore, let's delete it.

rulesets/books/hs-physics/styleguide/page.note.boundlessphysics.xhtml
```diff
@@ -1,12 +1,19 @@
+<body>
 <div data-type="chapter">
   <div data-type="page">
 
     <!-- Markup starts here -->
-    <div data-type="note" class="boundless-physics">
-      <div data-type="title">Title</div>
-
+    <div data-type="note" class="boudless-physics">
```
helenemccarron: isn't this a typo?

tomjw64:
@philschatz Requesting review before we merge

rulesets/books/ap-biology/styleguide/page.note.ap-science-practices.xhtml
```diff
@@ -1,16 +1,31 @@
-<div data-type="chapter">
-  <div data-type="page">
+<body>
```
philschatz: Why was this element added? I'm a little worried that the styleguide will have multiple `<body>` elements and that browsers will just remove this element. But if it allows browsers to open this ".xhtml" file (not _technically_ XHTML ) then maybe it's ok. But if so, are all the snippets updated?

rulesets/books/statistics/_config.scss
```diff
@@ -178,11 +178,11 @@ $Config_TargetLabels: append($Config_TargetLabels,
...
-  (className: 'lab',        moveSolutionTo: $AREA_NONE, counterName: lab,  titleContent: $_labTitleContent, isBodyWrapped: true ),
-  (className: 'collab',     moveSolutionTo: $AREA_NONE,  labelText: "Collaborative Excerise"),
-  (className: 'chapter-objectives', moveSolutionTo: $AREA_NONE,  labelText: "Chapter Objectives"),
-  (className: 'try',        moveSolutionTo: $AREA_TRASH, counterName: Try, titleContent: $_tryTitleContent), // try it notes are numbered based on the example preceding them
+  (className: 'finger',     moveSolutionTo: $AREA_NONE,  labelText: "Note", replaceHeader: true),
+  (className: 'lab',        moveSolutionTo: $AREA_NONE,  counterName: lab,  titleContent: $_labTitleContent),
+  (className: 'collab',     moveSolutionTo: $AREA_NONE,  labelText: "Collaborative Excerise", replaceHeader: true),
+  (className: 'chapter-objectives', moveSolutionTo: $AREA_NONE,  labelText: "Chapter Objectives", replaceHeader: true),
+  (className: 'try',        moveSolutionTo: $AREA_TRASH, counterName: Try, titleContent: $_tryTitleContent), // try it notes are numbered based on the example preceeding them
```
philschatz: 😊

rulesets/books/statistics/styleguide/page.note.collab.xhtml
```diff
@@ -1,10 +1,18 @@
+<body>
 <div data-type="chapter">
   <div data-type="page">
 
     <!-- Markup starts here -->
     <div data-type="note" class="collab">
       <div data-type="title">Collaborative Excerise</div>
+      <p><span data-type="title">Subtitle</span></p>
```
philschatz: Do all notes in all the books have this subtitle, or just some? It seems like this was added to all notes but I wanted to confirm

rulesets/mixins/_count.scss
```diff
@@ -162,3 +162,25 @@
...
+    '[data-type="footnote-ref"]',
+    '[data-type="chapter"], .appendix'
+  )
+}
+
+/// A basic mixin for counting footnote links in text using css counters
+/// @arg {Text} $name - Name of the css counter that will control footnote link numbering
+/// @group count
+@mixin count_footnoteLinks($name) {
+  @include _count(
```
philschatz: Do the footnote numbers and links to footnote number just happen to number the same (that's what it seems to be) or would it make sense to look up the footnote counter when creating the link (similar to exercise solutions)?

rulesets/mixins/_modify.scss
```diff
@@ -209,39 +209,111 @@
...
-  [data-type="note"].#{$selector} {
-    @if ($labelText) {
-      &::before {
-        container: span;
-        class: "os-title-label";
-        content: $labelText;
-        move-to: bNoteLabel;
+  //I really wish I didn't have to do this. Please change this if there is a better way
+  //Padding is for three parents, just in case
+  :not([data-type="note"]) > :not([data-type="note"]) > :not([data-type="note"]) > [data-type="note"].#{$selector} {
```
philschatz: Why is this check needed? Depending on the answer:

1. Maybe there could be an config option on the note for this? 
1. Is this to get the correct `h#` element name? If so, could this be represented as a loop?

rulesets/books/ap-biology/styleguide/page.note.ap-science-practices.xhtml
tomjw64: All the note snippets require 3 levels of nesting in order to pass code coverage tests. Perhaps I can remove the <body> tags and replace them with the more general <div> tag?

rulesets/books/statistics/styleguide/page.note.collab.xhtml
tomjw64: Ah, I should remove it for non-header-replacement notes

rulesets/mixins/_count.scss
tomjw64: They will happen to number the correct elements. I assume if they end up not doing so, there is a content issue that needs to be fixed anyways... Can you think of a situation where that wouldn't be the case?

rulesets/mixins/_modify.scss
tomjw64: 1. That wouldn't make sense, because level of nesting isn't really specific to any particular kind of note.
2. It's the only way in CSS to make sure that none of the notes parents (within reason) are `data-type="note"`. This is partially used to make sure the note gets the correct `h#`, but mostly to make sure that the buckets for notes and notes nested within other notes do not share any names (because buckets from the parent note were emptying in the child). Additionally, I'm not quite sure how this could be represented as a loop.

rulesets/books/statistics/styleguide/page.note.collab.xhtml
tomjw64: Changes committed

tomjw64:
@philschatz Made the requested changes to snippets

rulesets/mixins/_modify.scss
philschatz: RE 1: ok, makes sense 😄 

RE 2: what about something like:

```scss
@mixin loopStuff($depth, $className) {

  @include validate_type($depth, 'number');
  @include validate_type($className, 'string');

  @if ($depth >= 4) {
    // do nothing
  } @else {
    $headingLevel: $depth + 2;
    $noteDepthSelector: ();
    @for $index from 1 through $depth {
      $noteDepthSelector: append($noteDepthSelector, ":not([data-type='note']) >", 'space');
    }

    #{$noteDepthSelector} > :not([data-type="note"]).#{$className} {
      // ...
      container: h#{$headingLevel};
      // ...
      move-to: noteTitle-#{$depth};
    }
    @include loopStuff($depth + 1, $className);
  }
}

@include loopStuff(1, 'foo');
```

Also, I just noticed that the `$selector` variable name is actually only ever a class name (because of `.#{selector}`. Can it be renamed to `$className` to clarify its value?

tomjw64: I agree that className is more specific. I will make that change. 

The output of that loop is code similar to 
```scss
:not([data-type='note']) > [data-type="note"].foo {
  header: h3; }

:not([data-type='note']) > :not([data-type='note']) > [data-type="note"].foo {
  header: h4; }

:not([data-type='note']) > :not([data-type='note']) > :not([data-type='note']) > [data-type="note"].foo {
  header: h5; }
```
Unfortunately, these selector would select elements and run our note processes more than once. 

An html tree structured as follows will have the note element selected more than once:
```html
  <body>
    <div>
      <div>
        <div>
          <div>
            <div data-type="note" class="foo">
              Note
            </div>
          </div>
        </div>
      </div>
    </div>
  </body>
``` 
When showing the style of the note element in dev tools, all three styles are applied:
note-loop-1

Changing the code to the following:
```scss
@mixin loopStuff($depth, $className) {

  @if ($depth >= 3) {
    // do nothing
  } @else {
    $headingLevel: $depth + 3;
    $noteDepthSelector: ();
    @for $index from 1 through $depth {
      $noteDepthSelector: append($noteDepthSelector, "[data-type='note']", 'space');
    }
    #{$noteDepthSelector} [data-type="note"].#{$className} {
      header: h#{$headingLevel};
    }
    @include loopStuff($depth + 1, $className);
  }
}

@include loopStuff(1, 'foo');
```
yields:
```scss
[data-type='note'] [data-type="note"].foo {
  header: h4; }

[data-type='note'] [data-type='note'] [data-type="note"].foo {
  header: h5; }
```

which _may_ have a use-case somewhere, but I haven't found one for it (the use case would be a note nested within a note nested within a note)

tomjw64: @philschatz pushed some changes.
I've utilized the loop structure to be able handle any max-level of note nesting, without causing any changes to the currently working css files. If we eventually encounter notes nested in notes nested in notes, then we can just change from `$max-depth: 1` to `$max-depth: 2`

See: https://github.com/Connexions/cnx-recipes/pull/229/commits/4f9894b5749ea4051850990312d62c5cb6b58859#diff-5381e44af2366edee49e3f63bd310938R213

PR #231, opened 2017-06-01
Change OpenStax books server to dev.cnx.org

helenemccarron:
Use dev.cnx.org for our books, rather than point at tea.cnx.org.


philschatz:
why switch to using dev.cnx.org instead of the production data? I have a feeling devs will be deploying to dev.cnx.org and so it'll be moving around

helenemccarron:
Because 'tea.cnx.org` is considered the production server for TEA books, I think it might not be a good idea to be testing our content on that server. 

I believe `dev.cnx.org` is designed to be our internal development server (cc @reedstrm). This also makes the distinction between our books and the TEA books clearer.

philschatz:
@helenemccarron is this still needed?

kerwinso:
We are in the midst of trying to get `qa.cnx.org` set up with updated content and accounts from production. So this file would likely need to be updated once that's ready.

helenemccarron:
In the meantime, I think it clarifies what server to use to get the data from when testing for QA. I'm in favor of merging this and modifying it when we have a QA server

Issue #220, opened 2017-05-25
Multiple books => Appendix tables missing table numbers


kerwinso:
The table in Appendix C is missing the number for the table (`Table C1`).

Webview URL: 
https://dev00.cnx.org/contents/GFy_h8cu@10.61:-p-m85Qc@4/Appendix-C-Measurements-and-th

Webview on the left, PDF on the right:
![image](https://cloud.githubusercontent.com/assets/11298165/26458013/bc796e4a-4137-11e7-800b-36693b024afa.png)

kerwinso:
Also an issue in AP Physics:
https://dev.cnx.org/contents/jQSmhtXo@14.5:qvMKVKNW@5/Appendix-A-Atomic-Masses

![image](https://cloud.githubusercontent.com/assets/11298165/26736194/a721cb1a-478a-11e7-8e10-0fd7791750e2.png)

kerwinso:
In AP Physics, this is really a problem in Appendix C which contains multiple tables and a makeshift table of contents linking to each table: https://dev.cnx.org/contents/jQSmhtXo@14.5:cjQyhWg-@7/Appendix-C-Useful-Information#import-auto-id1374373

In webview, the link for every table says `Exercise 34.31`. Compare that to the PDF on the right in this screenshot.

![image](https://cloud.githubusercontent.com/assets/11298165/26789898/6a8180f2-49d7-11e7-9d9d-7134f112eb6e.png)

kerwinso:
List of affected books, each will be checked off when it passes QA:

- [ ]  College Biology
- [ ]  AP Physics

tomjw64:
Addressed in PR #237

tomjw64:
@kerwinso The output rulesets containing the fix can be found here: https://github.com/Connexions/cnx-recipes/tree/appendix-fix/rulesets/output

kerwinso:
Verified fix in [College Biology](https://dev.cnx.org/contents/GFy_h8cu@10.61:-p-m85Qc@4/Appendix-C-Measurements-and-th):

![image](https://user-images.githubusercontent.com/11298165/27092960-b735ceb8-502a-11e7-99d6-1051fba9aef3.png)

kerwinso:
Verified fix in [AP Physics](https://dev.cnx.org/contents/jQSmhtXo@14.5:qvMKVKNW@5/Appendix-A-Atomic-Masses):

[...]

Issue #221, opened 2017-05-25
College Biology => Appendix figures missing figure numbers


kerwinso:
The figures in the appendices are missing the numbers (`Figure B1`, `Figure B2` etc).

Webview URL: 
https://dev00.cnx.org/contents/GFy_h8cu@10.61:rr1s_Dnm@4/Appendix-B-Geological-Time

Webview on the left, PDF on the right:
![image](https://cloud.githubusercontent.com/assets/11298165/26458238/49819920-4138-11e7-9a45-cd8f158d2e60.png)

tomjw64:
Please only note the second commit that referenced this issue. The first commit is the same changes, but the commit message is in error.

tomjw64:
Addressed in PR #237

tomjw64:
@kerwinso The output rulesets containing the fix can be found here: https://github.com/Connexions/cnx-recipes/tree/appendix-fix/rulesets/output

kerwinso:
Verified fix on https://dev.cnx.org/contents/GFy_h8cu@10.61:hYdhUM0j@6/Appendix-A-The-Periodic-Table-. Whoomp, there they are:

![image](https://user-images.githubusercontent.com/11298165/27092784-251d7a9e-502a-11e7-8cf2-92f8ceaa04d3.png)

![image](https://user-images.githubusercontent.com/11298165/27092787-295a60e0-502a-11e7-8eea-886fe8c36f9b.png)

PR #237, opened 2017-06-13
Fix Appendix issues #221 and #220

tomjw64:
Modifies the numbering selector for tables/figures in appendices so that tables/figures that are direct children of appendices will be treated normally.

### DIFF
--- a/rulesets/mixins/_number.scss
+++ b/rulesets/mixins/_number.scss
@@ -162,7 +162,8 @@
     @include trashUnnumberedCaption($elementName, $defaultContainer);
   }
   .appendix {
-    :not(#{$elementName}) > #{$elementName}:not(.unnumbered) {
+    :not(#{$elementName}) > #{$elementName}:not(.unnumbered),
+    > #{$elementName}:not(.unnumbered) {
       @include utils_title($labelingAp, bCaption);
       @include setCaptionsHelper($captionType, $defaultContainer, $hasCaption, $hasTitle);
     }
--- a/rulesets/mixins/styleguide/page.figure.xhtml
+++ b/rulesets/mixins/styleguide/page.figure.xhtml
@@ -15,3 +15,19 @@
 
   </div>
 </div>
+
+<div class="appendix">
+
+  <!-- Markup starts here -->
+  <figure>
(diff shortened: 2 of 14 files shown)


tomjw64:
Fixes #220 and #221

rulesets/mixins/_number.scss
```diff
@@ -162,7 +162,8 @@
     @include trashUnnumberedCaption($elementName, $defaultContainer);
   }
   .appendix {
-    :not(#{$elementName}) > #{$elementName}:not(.unnumbered) {
+    :not(#{$elementName}) > #{$elementName}:not(.unnumbered),
```
philschatz: It seems like this line is "dead code" and never applies because the next line will always apply. Is that true? 

If so, what does this line do? (maybe makes sure subfigures are not numbered but I'm not sure)

tomjw64: @philschatz The line shouldn't be dead for that reason. These two selectors that are separated by the comma are mutually exclusive and there is no element that can satisfy both selectors.

This selector makes sure that tables/figures without the `unnumbered` class will be numbered properly. The issue fixed with this PR was that the
`.appendix :not(#{$elementName}) > #{$elementName}:not(.unnumbered)`
selector cannot select a direct child of an appendix, whereas the added selector
`.appendix > #{$elementName}:not(.unnumbered)`
can _only_ select a direct child of an appendix.

Seeing as I only added to snippets and added an additional selector, rather than modifying it. I'm unsure as to why the code coverage bot is saying that the original selector (left unchanged) is not being used. I'll look into it more.

tomjw64: Aha! Got it! There was a case unaccounted for in the snippets. You win this time Mr. Code Coverage Bot.

philschatz: oh yep, I forgot about the `>` at the beginning of the line ; )

Aside: If you install the codecov.io browser extension you can see which lines in the `output/{BOOK}.css` are no longer covered directly on the Pull Request. For example:

image

/cc @Dantemss because he loves codecov.io 😄

tomjw64:
@helenemccarron Requesting review before merge

PR #238, opened 2017-06-13
Add config option to trash half of EOB solutions

tomjw64:
Add a config location AREA_TRASH_HALF to allow trashing of half of the book's solutions (even solutions). This change is applied to College Bio in this PR.


rulesets/mixins/_link.scss
```diff
@@ -1,22 +1,37 @@
 //EOC composite pages must be created before this runs
 /// Creates a link to the solution and a link back to the problem for an exercise
 /// Note: This replaces the solution id with a new id whose value is `${EXERCISEID}-solution`
-@mixin link_linkToProblemsFromSolutionsEOC() {
+@mixin link_linkToProblemsFromSolutionsEOC($trashHalf) {
+  @if $trashHalf {
+    [data-type="exercise"].os-hasSolution {
+      &[trash-status$="1"],
```
philschatz: oh, clever!

helenemccarron: This could even be a function instead of list :D

tomjw64: 1. I've now added the option to the CustomPart section in the README. I admit though that it feels out of place in the documentation because no other option is listed there...
2. I've now changed the example config to use this option
3. The `exercise` style guide is contained in the general `mixins` styleguide. But I will go ahead and increase the number of exercises.
4. Looks like that may not currently be possible. Ross referred me to this GitHub issue for an enhancement that adds that ability: https://github.com/Connexions/cnx-easybake/issues/30
5?. The change would apply to any non-example exercise solutions moved to the end of chapter. If the note exercise itself were not moved to the end of chapter, the change would not apply. If the exercise were moved, then the change would. I went ahead and made the selector more specific, in that it will only apply trash status on non-example (which would mean, non-note) exercises that are found to have a solution in an earlier pass, just to be safe.
Let me know what you think of the latest commits. :)

PR #243, opened 2017-06-13
University physics recipe

helenemccarron:
Contains a new recipe for University Physics and fixes to allow unsupported functionalities (notes title wrapping, glossary in compound composite pages) to work. 

TODO:
 - [x] Confirm with Alana we do not need to number Chapter Review sections --- per Alana, it's ok to remove the chapter number in the Chapter Review section
- [x] exercises aren't numbered at all!


rulesets/mixins/_modify.scss
```diff
@@ -200,19 +210,22 @@
...
   $replaceHeader: map-get($noteContent, replaceHeader);
+  $useHeader: map-get($noteContent, useHeader);
 
   @include validate_type($className, string);
   @include validate_typeOptional($titleContent, map);
   @include validate_typeOptional($labelText, string);
   // Check that exactly one option is provided
-  @include validate_exists($titleContent or $labelText);
-  @include validate_exists(not($titleContent and $labelText));
+   @include validate_exists($titleContent or $labelText or $useHeader);
```
philschatz: Can you validate the type of `userHeader`?  maybe `@include validate_typeOptional($labelText, boolean);`

rulesets/books/_generator.scss
```diff
@@ -61,7 +64,7 @@ $_symbolsSectionTitle: "Symbols";
       $labelText: map-get($note, labelText);
       $titleContent: map-get($note, titleContent);
       $replaceHeader: map-get($note, replaceHeader);
-
+      $useHeader: map-get($note, useHeader);
```
philschatz: Can you validate the type of this (since it is being used here and not just being passed into another mixin)?

rulesets/books/_generator.scss
```diff
@@ -70,7 +73,10 @@ $_symbolsSectionTitle: "Symbols";
         @error "BUG: Specified both a label and a titleContent";
       } @else {
         // Phil does not know how to do `@else @if { ... }`
```
philschatz: Does this comment still apply?

hmm, maybe the following would need to be changed:

```scss
} @else {
  @if $titleContent {
```

to:

```scss
} @else if $titleContent {
```

rulesets/books/_generator.scss
```diff
@@ -152,17 +158,28 @@ $_symbolsSectionTitle: "Symbols";
...
     @include modify_addUnnumbered($contextSelector, $childSelector);
   }
+}
+:pass(3) {
+  //After: createChapterComposites
+  // Construct the set of CompoundComposite pages by looping through the Config_ChapterCompositePages
+  // and finding all that have a `childPages:` field
+  $_chapterCompoundCompositePages: ();
+  @each $page in $Config_ChapterCompositePages {
+    @if map-get($page, childPages) != null {
```
philschatz: Just curious why `!= null` was included here. In other places (like `@else if $labelText {` above) the `!= null` is not included

rulesets/books/_generator.scss
```diff
@@ -218,21 +235,56 @@ $_symbolsSectionTitle: "Symbols";
   //After: prepBookComposites
   @include reference_refChapterHeaderNodeAs(chapterHeaderNode);
   @include compose_prepChapterAreas($Config_BookCompositePages, chapterHeaderNode);
+
+  // Construct the set of CompoundComposite pages by looping through the Config_ChapterCompositePages
```
philschatz: It seems this is copy/pasted from above. If the code is the same, could it be written once, and maybe moved outside of the `:pass(#) {` block so the variable can be used in both `:pass(3)` and `:pass(4)`

rulesets/books/_generator.scss
```diff
@@ -218,21 +235,56 @@ $_symbolsSectionTitle: "Symbols";
...
     @include validate_enumOptional($moveSolutionsTo, $AREA__PREFIX__);
     @if $moveSolutionsTo == $AREA_EOC {
       @include compose_createEOCSolutions($Config_ChapterCompositePages, $page, sectionHeaderNode);
     } @else if $moveSolutionsTo == $AREA_EOB {
       @include compose_createEOBSolutions($Config_ChapterCompositePages, $page, sectionHeaderNode);
     }
   }
+
+  // @include count_countEOCExercises(exercise, '[data-type="chapter"]');
+  // TODO: Find out why this recounting is necessary; it seems to be for the targetLabels
```
philschatz: Why is this recounting necessary?

rulesets/books/_generator.scss
```diff
@@ -246,26 +298,15 @@ $_symbolsSectionTitle: "Symbols";
...
-  $_chapterCompoundCompositePages: ();
-  @each $page in $Config_ChapterCompositePages {
-    @if map-get($page, childPages) != null {
-      $_chapterCompoundCompositePages: append($_chapterCompoundCompositePages, $page);
-    }
-  }
-  @include compose_EOCcompoundCompositesPages($_chapterCompoundCompositePages);
 }
 :pass(6) {
+  //@include reference_refChapterNumberStringAs(chapterNumberString);
```
philschatz: Why is this comment included?

rulesets/books/_generator.scss
```diff
@@ -246,26 +298,15 @@ $_symbolsSectionTitle: "Symbols";
...
   //After: prepIndexTerms, createBookComposites
   @include modify_addIndexSymbolGroup("index", $_symbolsSectionTitle);
   //After: composite page creation
   @include count_countChapters(chapter);
   @include count_countAppendices(appendix);
-  // @include count_countEOCExercises(exercise, '[data-type="chapter"]');
-  // TODO: Find out why this recounting is necessary; it seems to be for the targetLabels
-  // This is mostly copy/pasta from above
+
   @if map-get($Config_PartType_Exercise, moveTo) == $AREA_EOC {
```
philschatz: It seems like this code was copied above in an earlier part of this pass. Why is it still included here?

rulesets/books/_generator.scss
```diff
@@ -218,21 +235,56 @@ $_symbolsSectionTitle: "Symbols";
...
+    @include validate_enum($resetAt, $RESET__PREFIX__);
+    @include validate_enum($numberAt, $NUMBER__PREFIX__);
+    @include validate_type($titleContent, map);
+
+    $resetSel: null;
+    @if       $resetAt == $RESET_CHAPTER        { $resetSel: '[data-type="chapter"]'; }
+    @else if  $resetAt == $RESET_COMPOSITE_PAGE { $resetSel: '[data-type="composite-page"]'; }
+    @else { @error 'BUG: Invalid enum'; }
+
+    @if $numberAt == $NUMBER_AFTER_MOVE {
```
philschatz: Why does this only apply when `moveTo == $AREA_EOC`?

rulesets/books/u-physics/_config.scss
```diff
@@ -0,0 +1,177 @@
...
+  Summary
+
+  These should be collated at the end of a chapter.
+
+  Markup: ./styleguide/book.composite.summary.xhtml-baked.html
+
+  Style guide: book.2-composite.summary
+*/
+/*
+  Conceptual Questions
```
philschatz: Are many of these areas already defined in another Physics recipe? It looks like they are copy/pasted in here. If so, could this just `@import` the other book and add the features custom to this book?

Also, then the snippets could also not need to be duplicated in this recipe directory

helenemccarron: Every single book has a summary. So far, we included one summary per book. I vaguely remember an ancient conversation about common elements in the styleguide but I do not remember how those work. This would affect all books, for the key terms and summary, mostly. But other collations could fall under a common format: key equations for example.

rulesets/books/_generator.scss
tomjw64: Looks like this change has now been made on the line after this one.

rulesets/books/_generator.scss
tomjw64: I was just being explicit about what what `map-get($page, childPages)` could not be in order to proceed. Leaving off the `!=null` would likely work just as well.

rulesets/books/_generator.scss
helenemccarron: For some reason, this comment didn't go away but validation has been added.

rulesets/books/_generator.scss
tomjw64: Given that, should I remove the `!=null`?

rulesets/books/_generator.scss
tomjw64: wooooooahh, there should not be two of these. This is probably something that happened during a merge conflict.

tomjw64: Duplicate deleted.

rulesets/books/_generator.scss
tomjw64: Comment deleted.

rulesets/books/_generator.scss
```diff
@@ -243,6 +253,7 @@ $_symbolsSectionTitle: "Symbols";
   @include compose_EOCcompoundCompositesPages($_chapterCompoundCompositePages);
 }
 :pass(6) {
+  //@include reference_refChapterNumberStringAs(chapterNumberString);
```
tomjw64: After some discussion, Helene and I think it would be best not to delete the mixin for now, as we can still think of a use case for it (in forming titles like "Chapter 5 Review"). We just happen to not use that in any books right now.

rulesets/books/_generator.scss
tomjw64: It only to `moveTo == $AREA_EOC` applies because the mixins being called only count EOC exercises.

rulesets/books/_generator.scss
tomjw64: Hmmm... I'll delete the comment

rulesets/books/_generator.scss
tomjw64: Gone 😃

rulesets/output/TEAeconomics.css
```diff
@@ -785,689 +785,690 @@
   class: "os-eoc os-ap-test-prep-container";
   data-type: "composite-page";
   data-uuid-key: ".ap-test-prep"; }
-:pass(2) [data-type="page"]:not(.introduction):not(.preface) > [data-type="document-title"] {
+
+:pass(3) [data-type="chapter"] [data-type="page"]:not(.introduction) > [data-type="document-title"] {
```
philschatz: It seems like this is only one of 3 places where other books were affected. Was it checked that these other books were not affected?

rulesets/output/TEAeconomics.css
```diff
@@ -785,689 +785,690 @@
...
   class: os-number; }
 
-:pass(4) body {
+:pass(5) body {
   counter-reset: exerciseAll; }
-:pass(4) .os-eoc [data-type="exercise"] {
+:pass(5) .os-eoc [data-type="exercise"] {
   counter-increment: exerciseAll; }
-:pass(4) div[data-type="chapter"] {
+:pass(5) div[data-type="chapter"] > [data-type="document-title"] {
```
philschatz: It seems like this is only one of 3 places where other books were affected. Was it checked that these other books were not affected?

rulesets/output/TEAeconomics.css
```diff
@@ -785,689 +785,690 @@
...
+:pass(6) [data-type="chapter"]::after {
   container: h2;
   class: "os-title";
   content: "Chapter " string(ChapNum);
   move-to: chaptTitleEOB; }
-:pass(5) [data-type="chapter"]::after {
+:pass(6) [data-type="chapter"]::after {
   class: os-chapter-area;
-  content: pending(chaptTitleEOB) pending(solutions-GETCHAPTER);
+  content: pending(chaptTitleEOB) pending(note-solutions-GETCHAPTER) pending(solutions-GETCHAPTER);
```
philschatz: It seems like this is only one of 3 places where other books were affected. Was it checked that these other books were not affected?

philschatz:
oh, it seems this now has merge conflicts ;(

rulesets/output/TEAeconomics.css
tomjw64: This change was meant to affect those books as well as u-physics. This selector would target pages within appendices (which now can have pages in them). If we do want outlines for composite-appendices, we could modify it, or make a new mixin, but using this one currently would not be appropriate considering the hard-coded class names, etc.

rulesets/output/TEAeconomics.css
tomjw64: I added support for note solutions to be added into the EOB for any book if needed. If we wanted to add a config option for having note solutions that collate to the end of book, then I suppose we could do so and add extra logic to the EOB solutions mixin. I've baked and checked Econ, a book with no notes that collate to the solutions, and it was correctly unaffected by the change.

rulesets/output/TEAeconomics.css
tomjw64: Soooo.... This is a strange one.
As it turns out, this change was made because without the addition, we'd get a node-set of an entire chapter, rather than just the title, as was intended. The only reason that we weren't ending up with entire duplicate chapters in our other books is because either the selector that received the chapter was a miss or that it never found a bucket. We'll revert this change for this PR, but we really need to make this change in another branch and track down the dead end code that makes the original selector not make duplicate chapters.

philschatz:
Thanks for addressing these issues. There are a couple minor ones about checking types of objects but otherwise, 👍

rulesets/books/u-physics/_config.scss
philschatz: Created #267 to revisit this

rulesets/mixins/_compose.scss
```diff
@@ -225,7 +232,7 @@
   $specialPageType: map-get($page, specialPageType);
 
   // Validate
-  @include validate_type($name, string);
+  //@include validate_type($name, string);
```
tomjw64: $name appears to be unused. I've removed the $name variable from this mixin entirely

rulesets/mixins/_compose.scss
```diff
@@ -9,7 +9,7 @@
     $className: map-get($page, className);
 
     // Validate
-    @include validate_type($name, string);
+   // @include validate_type($name, string);
```
tomjw64: It is not optional. I've un-commented the validate_type.

tomjw64:
Latest commit doesn't follow our normal conventions and may look somewhat "hack-y". However, this is to make up for inconsistent content in one very specific situation and should hopefully be a temporary fix that can be removed later, so there is no point in bloating the rest of our framework via trying to accommodate this kind of markup. It should also have absolutely no affect on any other operations.

Issue #242, opened 2017-06-13
AP Physics => Anchor links to tables all say 'Exercise 34.31'


kerwinso:
(this issue is a spinoff of #220)

In AP Physics:  Appendix C contains multiple tables, and a makeshift table of contents linking down to each table: 
https://dev.cnx.org/contents/jQSmhtXo@14.5:cjQyhWg-@7/Appendix-C-Useful-Information

In webview, the link for every table says `Exercise 34.31`. Compare that to the PDF on the right in this screenshot.  Is this a content issue, ruleset issue, or something else?

![image](https://cloud.githubusercontent.com/assets/11298165/26789898/6a8180f2-49d7-11e7-9d9d-7134f112eb6e.png)

tomjw64:
@kerwinso Does the fix for #220 also fix this issue?

kerwinso:
@tomjw64 No, it does not. I thought they might be caused by separate things, hence I opened this new bug. The tables in Appendix C seem to be numbered correctly, see here: https://dev.cnx.org/contents/jQSmhtXo@14.5:cjQyhWg-@7/Appendix-C-Useful-Information

But, the links at top still all say `Exercise 34.31`.

kerwinso:
Verified fix on https://dev.cnx.org/contents/jQSmhtXo@14.5:cjQyhWg-@7/Appendix-C-Useful-Information. 

(Note that the table numbers have periods in them, which they shouldn't -- this will be treated as a separate issue #247.)

![image](https://user-images.githubusercontent.com/11298165/27240459-634735c2-529a-11e7-97f0-465963f1b8c9.png)

PR #244, opened 2017-06-15
Fix issue #242

tomjw64:
Selectors for tables/figures that are direct children of appendices were added to numbering to fix issue #220. This PR adds the same selectors to target labeling.


tomjw64:
Fix was implemented in a different PR

PR #253, opened 2017-06-19
add CONTRIBUTING.md

kerwinso:
An attempt to document some of the workflow decisions we've recently discussed in scrum. 

This flow primarily applies to PRs where a QA person is involved. We can add that note if desired.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,19 +21,16 @@
     You probably won't need a second paragraph. But it's here when you need it.
 ```
 
-## Reference issues when committing
-Please use the issue number within commit messages, for example `Fix #220`. This will create a link within the PR to the issue itself and help track changes. Typically, only referencing the issue is enough but if there are multiple fixes for one issue, you may add a descriptive comment when needed, for example `Fix #220 - add table numbers back to appendix`.
-
 ## Review your own code before pushing
 Before you push, check your code and squash any superfluous commits, such as fast forward merge commits. 
 
-## Comment on the issue when addressed in the PR
-When the fix for an issue has been pushed to github and is ready for review, go to the issue and reference the PR in the comments. For example you may have something like `Addressed in PR #237`. This is also a good time to mention anything that the tester should be aware of or anything we may need to keep track of later on.
+## Update the PR description with the Issue number when addressing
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+    You probably won't need a second paragraph. But it's here when you need it.
+```
+
+## Reference issues when committing
+Please use the issue number within commit messages, for example `Fix #220`. This will create a link within the PR to the issue itself and help track changes. Typically, only referencing the issue is enough but if there are multiple fixes for one issue, you may add a descriptive comment when needed, for example `Fix #220 - add table numbers back to appendix`.
+
+## Review your own code before pushing
+Before you push, check your code and squash any superfluous commits, such as fast forward merge commits. 
+
+## Comment on the issue when addressed in the PR
```
philschatz: Is this step necessary since including the Issue number in the commit (or the Pull Request description) will create a link on the Issue page?

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
+# `cnx-recipes` Contributing Guidelines
+
+
+## Use a topic branch
+*(lifted from [this](https://github.com/puppetlabs/puppet/blob/master/CONTRIBUTING.md))*
```
philschatz: Could this reference/use the https://github.com/openstax/napkin-notes/blob/master/CONTRIBUTING.md one? (with modifications to it listed here)

The items listed below probably apply to most of the repositories worked on by Openstax. If some of these items do apply, but are not in the napkin-notes `CONTRIBUTING.md` file could you make a Pull Request to the napkin-notes one? That way tutor devs/QA would benefit from that additional information

CONTRIBUTING.md
kerwinso: I think that it is, because linking to the issue number in the commit does not reference the _PR_ that includes that commit. It's helpful for QA to know which issues are included in the PR.

Example: look at https://github.com/Connexions/cnx-recipes/issues/250. This is an issue. The commit included the issue number, so it is automatically referenced on the issue page. However, just by looking at the issue, I can't tell which PR addresses that issue. (BTW, it's https://github.com/Connexions/cnx-recipes/pull/252.)

At the very least, a dev should reference the issue number in the PR, so github will create a link between the PR and the issue.

So basically: a dev should comment in either the PR, or the issue itself. I would prefer it to be in the issue itself (and that is our practice with oer.exports), but in any case the goal is to have a hyperlink between the PR and the issue.

philschatz:
yay! Only 1 more review to go! /cc @Stackblocks

PR #256, opened 2017-06-20
Use editable python package installs

pumazi:
This removes the need to uninstall before reinstalling.


philschatz:
This will likely require devs to run `rm -rf ./venv/` after this is merged. /cc @helenemccarron @tomjw64 @Stackblocks

PR #263, opened 2017-06-22
Don't copy entire chapters when trying to get a chapter title

tomjw64:
Fix bug where trying to get a chapter title node giving you the entire chapter. Oops.


tomjw64:
Please do not merge until we figure out why this wasn't affecting all of our titles by creating duplicate full chapters.

tomjw64:
I removed some supposed dead-end code, and diffed the xhtml for stats, econs. U-physics also seems to be unaffected. Since each was unaffected and since that code was written when we only had stats, then it's safe to declare the removed code as "dead-end"

tomjw64:
This is not referencing an issue, so do not assign to QA

PR #265, opened 2017-06-23
ensure the snippets are well-formed XML

philschatz:
I was looking at converting the HTML snippets to be CNXML (see corresponding PR) and realized our snippet files are not valid XML. 

This cleans that up.


helenemccarron:
Yay for cleaner things!

PR #266, opened 2017-06-23
CNXML Snippets

philschatz:
I was looking at converting the HTML snippets to be (mostly) CNXML so that:

1. they more-closely resemble the original content (in the Styleguide)
2. when the CNXML->HTML conversion changes the snippets will also change

And as a bonus, we can:

- show the original CNXML in the styleguide for Content Managers
- add baked HTML validation
- use these snippets for validating the Raw HTML

Not everything can be represented in CNXML, namely the structural chapter/page/appendix elements so those remain HTML

## Example Snippet (with CNXML)

```xml
<div data-type="chapter">
  <h1 data-type="document-title">Chapter Title</h1>
  <div data-type="page">
    <div data-type="document-title">Page Title</div>
    Normal book content goes here

    <!-- Markup starts here -->
    <section xmlns="http://cnx.rice.edu/cnxml" class="ap-test-prep">
      <title>Test Prep for AP<sup>®</sup> Courses</title>
      <exercise id="ex1">
        <problem>
          <para>What is 2+2?</para>
        </problem>
      </exercise>
    </section>
    <!-- Markup ends here -->

  </div>
</div>
```

# Notes

This PR is made to `lint-snippets` instead of `master` so the changes will be easier to see. Once #265 is merged, change this PR to point to `master`.

# TODO

- [x] change the styleguide script to convert the CNXML snippet to raw HTML before continuing
- [x] actually replace the snippet files
- [ ] remove the `script/_*-to-*.xsl` files


rulesets/books/biology/styleguide/book.composite.art-connection-questions.snippet.xml
```diff
@@ -0,0 +1,18 @@
+<div xmlns="http://www.w3.org/1999/xhtml" data-type="chapter">
+  <h1 data-type="document-title">Chapter Title</h1>
+  <div data-type="page">
+    <div data-type="document-title">Page Title</div>
+    Normal book content goes here
+
+    <!-- Markup starts here -->
+    <section xmlns="http://cnx.rice.edu/cnxml" class="art-exercise">
```
reedstrm: Hmm, not sure showing the mixed form is useful here at all. No one will write files like this, and no one will have to process files like this. Perhaps we leave the higher-level organization examples in all xhtml?

philschatz: @reedstrm Yeah this file is a bit awkward. I added CNXML so that when the CNXML->HTML conversion changes (like https://github.com/Connexions/rhaptos.cnxmlutils/pull/157) then the snippets will also change, which helps find breaking changes in the CNXML->HTML conversion.

Easybake needs to have the `<div data-type="chapter"><div data-type="page">` wrapper. And some snippets (like the exercises one) need to have multiple pages/chapters so I was not sure how to mark that up.

Any thoughts on how to solve this without combining XHTML and CNXML into 1 file?

I could just not have the CNXML option in the dropdown but that would make the styleguide useless for CM's or WiseWire...

For simple examples that can be contained inside 1 Page, I could create a snippet format that _only_ allows CNXML

philschatz:
Hmm, fixing indenting might be a manual process unless folks are OK with me running `xmllint --pretty 1` or `xmllint --pretty 2` on the XML. Here are some examples:

Original: `<foo><bar x="1"/> <bar/><bar/> </foo>` (note the lack of space in front of `<bar ...`)

`xmllint --pretty 1` output:

```xml
<foo>
  <bar x="1"/>
  <bar/>
  <bar/>
</foo>
```

`xmllint --pretty 2` output:

```xml
<foo
  ><bar
      x="1"
  /> <bar
  /><bar
  /> </foo
>
```


`xmllint --pretty 2` harder in the eyes but easier for diffing because:

1. whitespace is preserved exactly
1. attributes are on separate lines

(I'm guessing the consensus will be for `--pretty 1`)

/cc @caderitter

caderitter:
@philschatz just for readability, ` --pretty 1`

philschatz:
@caderitter XML Snippets are now always formatted properly. `./script/test` will do that 1st, but you can run `./script/lint-snippets` directly to auto-format them 😄

philschatz:
this PR was automatically closed because #265 was merged and the branch was deleted. so reopening

helenemccarron:
Didn't we talk about the possibility of moving away from CNXML in a not too distant future? If so, is this a path that would be useful down the road?

Also, I agree with Thomas' point that not every dev is able to write CNXML, especially when it comes down to finer details. This would require quite a lot of training before being totally effective. 
One way around this problem would be to copy the generated CNXML. Alternatively, the CMs are currently using tagging legends that contain that information, which they develop and submit to vendors for reference then keep up to date. That is the most reliable source for the CNXML, contains a example of the markup without all the bells and whistles in the generated content. In this case, we would be merely duplicating the CNXML for our own use (coverage) while the CMs use their own tracking system (the tagging legend). This makes things more error prone because if the markup changes, the change must be reflected in two places manually. Unless there is a way to pull the CNXML directly and automate the process of insert it in the snippet, I'm not sure how we would  achieve `when the CNXML->HTML conversion changes the snippets will also change` efficiently and accurately. 

In my opinion, for this to be an accurate representation of what is really used in the markup, it would make more sense for the CMs to create those snippets themselves since they are the authority in that field, but then, they would have to learn about the HTML that doesn't fall into the CNXML, not to mention the extra work involved in creating and maintaining them. This _could_ replace the tagging legend but would require a significant shift in their workflow, not to mention the implications for vendors. They must be presented with documents that are highly readable and devoid of any extra information that might be confusing to them. 

I am also not completely clear on what benefit there is in having those closer to the original content. When developing a recipe, the raw makes more sense, since I'm targeting the HTML, not the CNXML. What use would the CNXML be to me as a developer?

philschatz:
Hmm, it looks like the recipes still use the class name to refer to some elements instead of the `data-type` attribute (see codecov link referring to `.example`)

If there are no objections I will try to switch it to use the data-type attribute

/cc @helenemccarron

rulesets/books/biology/styleguide/book.composite.art-connection-questions.snippet.xml
reedstrm: Yeah, a conundrum. Evil thought, can we _style_ the snippet so that when viewing the CNXML form, we _grey out_ the html wrapper parts? I want to make it real clear that these are not copy & pasteable examples for any real system.

philschatz: I agree on the “gray out the non-CNXML parts of the snippet file” or have some other way of describing the snippets (like specially-formatted comments that say `<!-- START:Page -->` or blatantly non-XHTML/cnxml elements) but I don’t know of a good way to get that done quickly.

I can try out the "specially-formatted comments" one since, for _most_ of the snippets the only XHTML is the chapter/Page wrapper.

reedstrm: Can we suppress the *CNXML* stanza for all the book elements, since they contain NO CNXML AT ALL?  Finding a solution for displaying the wrapped/mixed ones I guess could hold off - the existing red `<-- Markup starts here -->` perhaps `<-- CNXML markup starts here -->` (and we change it when we move off CNXML) is probably enough for now.

reedstrm:
This got buried as comment on old code:

philschatz:
@reedstrm thanks for pulling out that discussion chain 😄  . 

@helenemccarron & @reedstrm : What do you think of f3c0cac ? 

That makes the snippets be *just* CNXML but it's a bit hackish because:

1. you need to format the comments exactly
2. XSLT will be injecting XML that might cause the snippet to generate invalid XML which means you'll get a vague error when easybake runs

helenemccarron:
@philschatz I like the CNXML only version. The advantage to that is that the Content Managers can got and create the CNXML that they would be using in the platform. They wouldn't have to convert it to HTML or have a dev work on them. Regarding the "hackish" nature, how much do you think it's going to be an issue? Is there a way to reduce human error? Maybe we can create an empty snippet that would contain the proper structure. Also, what is the nesting rule for comments? Can we have clear, accessible documentation for the CMs/new hire to use?
Other than that, it looks great!

philschatz:
ok, thanks for the feedback! Can this PR be merged now and have a separate PR for "use Comments instead of mixing HTML and CNXML"?

/cc @helenemccarron @reedstrm

helenemccarron:
@phil, sounds good!

PR #274, opened 2017-06-27
define CSS to use the XHTML namespace

philschatz:
This is related to https://github.com/Connexions/cnx-epub/pull/119 and part of getting cnx-easybake to parse XHTML instead of HTML. 

Easybake does not choke with this code in place.

/cc @reedstrm

### DIFF
--- a/recipes/books/_example/_config.scss
+++ b/recipes/books/_example/_config.scss
@@ -71,7 +71,7 @@ $_compoundTestPrep: (
   Style guide: book.composite.practice
 */
 $Config_ChapterCompositePages: (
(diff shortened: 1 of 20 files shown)


reedstrm:
Correct, it won't choke (good first step) but it fails to do anything with it, because of:
https://github.com/Connexions/cnx-easybake/blob/master/cnxeasybake/oven.py#L151-L152

philschatz:
**Note:** This now depends on https://github.com/Connexions/cnx-easybake/pull/72 being merged. You can just re-run travis once the `easybake` PR is merged

Issue #278, opened 2017-06-28
University Physics => Index missing letter 'A' header


kerwinso:
URL: https://dev.cnx.org/contents/1Q9uMg_a@5.13:cqPvIeML@5.13/Index

The header for index terms starting with the letter 'A' is missing. Note that this index does not seem to have index terms that start with a symbol. Does the ruleset require terms with symbols to be present in order for the `A` to display?

![image](https://user-images.githubusercontent.com/11298165/27648122-22beb536-5bf3-11e7-82db-e8329eaa9f68.png)

tomjw64:
Addressed in PR #281

kerwinso:
Verified fix on https://qa.cnx.org/contents/1Q9uMg_a@5.22:cqPvIeML@5.22/Index after rebaking on `index-symbol` branch.

![image](https://user-images.githubusercontent.com/11298165/27919020-35b48c30-6236-11e7-99cc-18856ca22d69.png)

PR #281, opened 2017-06-29
Only display 'Symbol' in index when there are terms in its section

tomjw64:
Stops a hardcoded "Symbol" group container title from overriding the "A" group container title to resolve issue #278.

### DIFF
--- a/rulesets/mixins/_modify.scss
+++ b/rulesets/mixins/_modify.scss
@@ -106,7 +106,7 @@
   }
 }
 
-/// Modifies the ' Symbols' group name in index
+/// Modifies the ' Symbols' group name in index, should it exist, as the first group-label.
 /// @arg {string} $location - The string of the location where title needs to be added
 /// @arg {string} $name - The string of the title for this group
 /// @group modify
--- a/rulesets/mixins/_modify.scss
+++ b/rulesets/mixins/_modify.scss
@@ -112,7 +112,9 @@
 /// @group modify
 @mixin modify_addIndexSymbolGroup($location, $name) {
   .os-#{$location}-container > div.group-by:first-of-type > span.group-label {
-    content: $name;
+    &:empty {
+      content: $name;
+    }
   }
 }
 /// Adds metadata to the composite pages
--- a/rulesets/mixins/_modify.scss
+++ b/rulesets/mixins/_modify.scss
@@ -106,7 +106,7 @@
   }
 }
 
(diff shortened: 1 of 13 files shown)


rulesets/mixins/_modify.scss
```diff
@@ -112,7 +112,9 @@
 /// @group modify
 @mixin modify_addIndexSymbolGroup($location, $name) {
   .os-#{$location}-container > div.group-by:first-of-type > span.group-label {
-    content: $name;
+    &:empty {
```
philschatz: Woah, that was a weird bug. Is there another way to add this title? Just by looking at the code it does not seem clear to me why `:empty` is added here.

philschatz: Woah, that was a weird bug. Is there another way to add this title? Just by looking at the code it does not seem clear to me why `:empty` is added here.

tomjw64: When we group index terms, we place `group-by` elements as containers to each group. Each `group-label` contains the first letter by which the terms are sorted. For index terms beginning with not `[a-zA-Z]`, rather than their first letter, we instead group them by an empty string (which in-turn puts it at the top of the index. In the past, it looks like we selected that element by just assuming it was the first label in the index. However, since u-physics contains no terms beginning with a symbol, the `group-by` with the label `A` happened to be the first, and thus was replaced with `Symbol`. Adding this `:empty` will now correctly _only_  select the `group-label` that was meant for the `Symbol` group.

An alternative to using this selector to add the correct title could be instead of grouping symbols by `''`, we could group them by `' '` (or some other `[^a-zA-Z]`, i suppose)  and then select the correct element with a `:match(^[\s])` rather than with an `:empty`. Additionally we could group them by `' Symbols'` (with a space at the beginning to make it appear first, but I think that would make it hard to figure out why it was working.)

Do either of those methods seem preferable, or more clear to a reader?

helenemccarron: Would it make more sense if it was commented in the code to offer an insight on the reasons for selecting this approach?

philschatz: @tomjw64 ooh, I like grouping them by `' Symbols'` (starts with a space) instead. I think that would help make the code clearer

tomjw64:
I made the symbols group sort by `' Symbols'` instead of by `''`, then I changed the label modification selector from `:empty` to a `:match`, to remove the leading space from content. Also updated the documentation comment of the label modification mixin to note the change.

PR #287, opened 2017-07-06
replace all references to "ruleset" with "recipe"

philschatz:
There were several variables, documentation, and one directory that referred to "rulesets". This renames those references to be "recipe"

**Note:** This renames the top-level directory to be `recipes`.

/cc @kerwinso & @Stackblocks because the CSS output directory will change from `rulesets/output/` to `recipes/output/`


tomjw64:
I was not added as a reviewer but ✔️ anyways!

Issue #214, opened 2017-05-23
Multiple books => Duplicate learning objective content in the Summary section


kerwinso:
1. Go to https://dev00.cnx.org/contents/GFy_h8cu@10.61:2zzm1QG9@7/34-Proteins.
1. Click the Summary link with the triangle near the top of the page.

Note that the content is a duplicate of the learning objectives in the main body of the page.

![image](https://cloud.githubusercontent.com/assets/11298165/26366192/bebe954c-3faf-11e7-811d-f8ec98b3df5b.png)

helenemccarron:
The ruleset eliminates the abstract/summary but still appears in webview.

kerwinso:
Also an issue in University Physics.

openstaxalina:
We have toyed with the idea of moving the summary **out** of the metadata (< abstract >) and **into** the body of the content, but it would require a content change.  I think it would be best to leave the summary in the metadata/abstract and **not** duplicate it in the content body.  But I may not know the full history - we can discuss further if needed.

reedstrm:
Currently, the recipe does the copying of  the summary into the body. Note that we have a conceptual issue with the metadata - nothing inside the <metadata> tag is interpreted (except for composite-pages/chapters) when the baked HTML is re-interpreted as pages and stored in the database. This means that the only metadata that can be overridden (_i.e._ differ in baked vs. raw) is the title of a page. This is controlled by the title that is _outside_ the metadata tags. It would be wrong to let the baking process change the metadata of the published page.

openstaxalina:
_**The Plan:**_

1) **Tech** For OpenStax books cooked with recipes, hide the Summary drop-down because the 
< abstract > field is either empty or being duplicated within the content tag.

2) **Content** Later on, move the learning objectives from the < abstract > and into the content tag.

3) **Tech** Later-later on, re-enable the Summary drop-down in case we can ever add anything actually useful there.

reedstrm:
Further discussion with @helenemccarron and @tomjw64 lead to leaving this in place, but finding a way to suppress display of the summary _per book_.  We'll need a way to pass a whole-document level style option, store it, and return it. with the json. Will discuss design choices with @philschatz

kerwinso:
The summary toggle for baked books has been hidden with this PR: https://github.com/Connexions/webview/issues/1574. 
Now only 1 set of learning objectives is displaying, which is the desired behavior:

[...]

PR #291, opened 2017-07-10
The ol' metadata switcheroo

tomjw64:
Remove metadata desc from composite elements, re-add it to raw elements per discussion with @helenemccarron and @reedstrm

### DIFF
--- a/rulesets/books/_generator.scss
+++ b/rulesets/books/_generator.scss
@@ -35,8 +35,6 @@ $_symbolsSectionTitle: "Symbols";
   }
   @include modify_GenericNote();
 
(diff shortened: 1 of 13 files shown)


tomjw64:
This is likely failing code coverage because we don't have snippets for metadata descriptions? Is this something we should add even though metadata will never be displayed/styled? @philschatz @helenemccarron

philschatz:
I think that since they are being selected (even if only to be removed) then it should be in the snippets. It might make sense to add a snippet to the `/rulesets/mixins/styleguide/` rather than for each book.

philschatz:
refs #214

Issue #284, opened 2017-07-06
Econ => Figure titles not styling correctly 


Stackblocks:
# Scenario
Figure titles in Principles of Economics are not bold, also some figure captions are not spaced properly from the figure title.

# Reproducibility Details
Persistent throughout the book.

- **Impact area**: figure titles
- **User**: n/a
- **URL / Location**: https://qa.cnx.org/contents/aWGdK2jw@11.345:6RH0nLs4@8/12-What-Is-Economics-and-Why-I
- **How many times were you able to reproduce the issue?** Persistent throughout the book.

## Steps to reproduce

- Open the baked webview on qa.cnx.org
- View a figure.

## Screenshots

![image](https://user-images.githubusercontent.com/12678235/27915857-0f289bfa-622d-11e7-87dc-6cd6a1f32555.png)

![image](https://user-images.githubusercontent.com/12678235/27916006-7d5af8e8-622d-11e7-9ca1-4fe5e79d0a82.png)

# Environment

- **Instance**:  qa.cnx.org
- **OS**: macOS
- **Browser**: Chrome

# Impact

- **Severity**: moderate
- **Frequency of occurrence**: persistent throughout book

caderitter:
Addressed in [webview:1570](https://github.com/Connexions/webview/pull/1571).

kerwinso:
Another example of the missing space between figure title and caption bug: http://qa.cnx.org/contents/aWGdK2jw@11.345:vF-Cvj3d@5/14-How-Economists-Use-Theories
(Source: https://legacy-qa.cnx.org/content/m48593/latest/source)
![image](https://user-images.githubusercontent.com/11298165/28027126-ad694df4-655d-11e7-9e1b-9772b42d7ffe.png)

But, here's one that looks OK, whose markup looks the same (spacing etc.):
https://qa.cnx.org/contents/aWGdK2jw@11.345:JgDXaOLP@11/11-Introduction
(Source: https://legacy-qa.cnx.org/content/m48590/latest/source)
![image](https://user-images.githubusercontent.com/11298165/28027154-bc7a522a-655d-11e7-98e4-c27f0693a8a6.png)

Hmmmm.

tomjw64:
I'll post here a copy pasta of a slack message regarding the spacing issue:

[...]

PR #294, opened 2017-07-13
Style figure titles correctly

caderitter:
This is the recipes part of webview PR #1571 https://github.com/Connexions/webview/pull/1571

Both address #284 in recipes


caderitter:
@helenemccarron rebased!

PR #297, opened 2017-07-27
TEA Final delivery recipe changes

tomjw64:
- Adds prefaces
- Ensures metadata correct on composite-chapters
- Ensures deletion of index, solution eob elements
- Updates books.txt with new book uuids


philschatz:
Closing because it seems this is no longer applicable. Please reopen if I am mistaken @tomjw64

philschatz:
Should this be reviewed now?

philschatz:
@helenemccarron @tomjw64 : Did we ever use `./script/fix-epub-inter-module-links`? I think we ended up doing something else... If we did not, I can remove it so it does not bit rot.

recipes/mixins/_compose.scss
```diff
@@ -351,6 +351,7 @@
 
     $keyName: if($specialPageType == $PAGE_INDEX, '<index>', '.#{$className}');
 
+    // FIXME: body body should never occur
```
philschatz: What does this comment mean?

recipes/mixins/_modify.scss
```diff
@@ -125,12 +125,15 @@
...
 
-  @if $hasCompositeChapter {
-    [data-type="composite-chapter"]  {
-      // Copy-pasta from above
-      content: nodes($bookMetadata) content();
-    }
-  }
+  // Commented in favor of adding metadata during creation of composite-chapters (ordering is incorrect if it's done this way)
+  // Metadata must be after doc-title in DOM, otherwise modify_retitleCompositeMetadata will not work
+  // Will keep this here in case we want to change it back somehow
```
philschatz: Does this still need to be commented code?

recipes/books/_generator.scss
```diff
@@ -9,9 +9,62 @@ $_symbolsSectionTitle: "Symbols";
 //TODO: add a switch when our books have composite chapters, define logic/behavior for that switch. This simply could be in the _config files for all books or set based on the presence of at least one type of composite chapter (appendix, multiple page answer key etc.).
 
 
+@mixin misc_webview_styling() {
+  // Adapted from https://github.com/Connexions/webview/blob/master/src/scripts/modules/media/body/content-common.less#L341
+  //   // Block-ish lists
+  //   ul:not([data-display='inline'])          { @include list_style(block, bulleted); }
```
philschatz: Are these comments useful for others? If not, I can remove them

recipes/mixins/_modify.scss
```diff
@@ -184,29 +187,51 @@
...
-       content: string(noteTitle);
-       move-to: genericNoteLabel;
-     }
-     move-to: trash;
+  @include _genericNoteNesting(1);
+}
+
+@mixin _genericNoteNesting($maxDepth, $depth: 0) {
+  @if ($depth <= $maxDepth) {
+    $headingLevel: $depth + 3;
```
philschatz: Why are these `+3`? Can you include a comment in the SCSS file?

recipes/mixins/_modify.scss
```diff
@@ -184,29 +187,51 @@
...
+    $noteDepthSelector: (':not([data-type="note"]) > :not([data-type="note"]) > :not([data-type="note"]) > ');
```
philschatz: Why does this variable begin with 3 `:not(...) >` selector fragments? Can you include a comment in the SCSS file?

recipes/mixins/_modify.scss
```diff
@@ -184,29 +187,51 @@
...
-      container: div;
-      content: pending(genericNoteLabel);
-      class: "os-title";
-      move-to: genericNote;
-     }
-     &:deferred {
-      content: pending(genericNote) content();
-     }
-   }
+    #{$noteDepthSelector} [data-type="note"]:not([class]) {
```
philschatz: It is not clear why `:not([class])` is included at the end of this selector. Can you include a comment explaining when this would occur?

recipes/mixins/_modify.scss
```diff
@@ -184,29 +187,51 @@
...
+    }
+    @include _genericNoteNesting($maxDepth, $depth + 1);
+  }
+}
+
+@mixin _genericNoteAtLevel($depth, $headingLevel) {
+  $minorLevel: '';
+  @if $depth > 0 {
+    @for $i from 1 through $depth {
+      $minorLevel: $minorLevel + 'Minor';
```
philschatz: Why is `Minor` appended several times rather than just using the `$depth` number? Also, in other places (like `headingLevel`) the value of the variable is a number which does not match with this

recipes/mixins/_modify.scss
```diff
@@ -184,29 +187,51 @@
...
-       class: "os-title-label";
-       content: string(noteTitle);
-       move-to: genericNoteLabel;
-     }
-     move-to: trash;
+  @include _genericNoteNesting(1);
+}
+
+@mixin _genericNoteNesting($maxDepth, $depth: 0) {
+  @if ($depth <= $maxDepth) {
```
philschatz: Since this is a new mixin, should the argument types be checked?

recipes/mixins/_modify.scss
```diff
@@ -471,6 +487,24 @@
   }
 }
 
+/// Will add a title/header with string param to solutions in a selected location
+/// @group modify
+@mixin modify_addSolutionHeader($selector, $titleString) {
+  #{$selector} [data-type="solution"] {
```
philschatz: Since this is a new mixin, should the argument types be checked?

script/qa_bake.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
philschatz: Are these scripts still used by devs or QA folks?

teg-styles.css
```diff
@@ -0,0 +1,179 @@
+
```
philschatz: Is this no longer needed because of the other CSS file?

teg-styles.xhtml
```diff
@@ -0,0 +1,42 @@
+<html xmlns="http://www.w3.org/1999/xhtml">
```
philschatz: @helenemccarron Would it be useful to have an HTML file like this one that exercises the stuff in the TEG-deliveries-something.css file?

recipes/mixins/_compose.scss
tomjw64: Just below this, there is `body::after` selector. Which is fine by itself, but notice that above it's nested within another `body` selector, which means it's looking for `body body::after`, which means this is almost definitely dead code.

recipes/mixins/_modify.scss
tomjw64: Eh, yeah we probably don't need it anymore. I'll get rid of it.

recipes/mixins/_compose.scss
tomjw64: Taking this out changes TEAeconomics and TEAap-physics recipes. However, a search for `body body` within those books returns nothing. I will remove it.

recipes/mixins/_modify.scss
tomjw64: No need to check argument types, as it's only a helper mixin to another which has no arguments

recipes/mixins/_modify.scss
tomjw64: :+1:

recipes/mixins/_modify.scss
tomjw64: We've discussed this one before, heh. I'll leave a comment 😄

recipes/mixins/_modify.scss
tomjw64: :+1:

recipes/mixins/_modify.scss
tomjw64: Yeah, I agree that wasn't the best decision. On the other hand, I'm not positive easybake supports numerals in bucket names and changing this will change all of our rulesets post-QA, which seems like a large price to pay for changing a string that is only used for an easybake bucket name.

tomjw64: I'm going to confirm with Ross that easybake "officially" supports numerals in bucket names

recipes/mixins/_modify.scss
tomjw64: Yup, I'll do that.

recipes/books/_generator.scss
tomjw64: I haven't had to do any styling-related fixes, but I feel they will be helpful for what selectors may be acceptable or not acceptable to change

recipes/mixins/_compose.scss
tomjw64: Oh. Turns out taking this out changes all of the recipes. Well, I feel like if `body body` is a valid selector anywhere, that's a problem in itself.

recipes/mixins/_modify.scss
tomjw64: easybake does support numerals within bucket names. I'll make the change, I suppose, and very carefully.

philschatz: you can run `./script/diff-book-prepare --all` in `master` (& grab a cocktail) and then `./script/diff-book --all` in this branch to verify nothing changed 😄

tomjw64: I'll do that. See you in 30 years!

philschatz:
@tomjw64 Thanks for addressing those issues! It seems the code coverage has decreased a little bit. Could it be fixed by just providing a couple more snippets?

philschatz:
@helenemccarron Any thoughts on having a TEA test HTML file so if we need to send them updated CSS we have an HTML file to use that exercises their selectors? This PR has a file like that called `teg-styles.xhtml` but it seems to have been obsoleted by `TEG-deliverable-CSS.css ` . 

See https://github.com/Connexions/cnx-recipes/pull/297/files#diff-277aa513a8112786c6823b6fc620f5ab

helenemccarron:
I don't know if we even need the CSS for TEG in the repo at all. We have[ issue 325](https://github.com/Connexions/cnx-recipes/issues/325)  which documents exactly what was provided to them. IMO, the HTML isn't that useful since we would have to actually update it with markup from their platform and snippets of their stylesheet to allow for testing locally as opposed to in the browser.

I think we should think of a better method for CSS delivery in the future but I don't see value in maintaining what's currently in the repo.

Issue #179, opened 2017-04-20
Multiple Books => Key Terms not sorted alphabetically. 


Stackblocks:
# Description
The Key Terms in the baked version  are not sorted in alphabetical order. While the example and screenshot linked are from Chapter 1 of Introductory Statistics, this is true throughout the books.

# Related Materials
- [Chapter 1 Key Terms](https://dev00.cnx.org/contents/MBiUQmmY@18.13:u8z6aR1b@11/Key-Terms) 

![image](https://cloud.githubusercontent.com/assets/12678235/25251253/51a221bc-25de-11e7-9ae9-6f3a8d3a69bb.png)

kerwinso:
Also found in College Biology:
![image](https://cloud.githubusercontent.com/assets/11298165/26322878/6ae8b23a-3ef2-11e7-927a-04cace7e325c.png)

Stackblocks:
Also still happening in Econ.
![image](https://cloud.githubusercontent.com/assets/12678235/26371263/d80c58a8-3fbf-11e7-9aee-dd422ba7490e.png)

tomjw64:
The relevant baked glossaries appear with no issue with content grabbed from tea.cnx. Are these examples from dev.cnx? If so, this is almost definitely a content issue with dev.cnx.

# **Stats:**
![stats](https://cloud.githubusercontent.com/assets/1285308/26697908/572dc71e-46d9-11e7-9930-3ca6fafd5bbc.png)


# **Econ:**
![econ](https://cloud.githubusercontent.com/assets/1285308/26697914/5e11c3dc-46d9-11e7-82e7-9d5e0085f67d.png)

Stackblocks:
Tests were performed on dev.cnx, so that full well may be the case.

tomjw64:
I just pulled content from dev and still cannot reproduce the issue. I guess not instead of a content issue this is a ??? issue.

tomjw64:
Never-mind, I found exactly what's happening with statistics.
![untitled](https://cloud.githubusercontent.com/assets/1285308/26701797/0bf664fe-46e8-11e7-99c7-32a42858e75c.png)

tomjw64:
@Stackblocks does this still occur on a fresh fetched book? on latest easybake master version?

kerwinso:
This is occurring in the Glossary sections of AP Physics, which was just baked yesterday:
[...]

PR #299, opened 2017-07-28
Fix u-physics glossary sort

tomjw64:


tomjw64:
Issue #179 is partially addressed by this change (just for u-physics)

PR #330, opened 2017-09-18
Use Javascript implementation of SASS and remove ruby install

philschatz:
Refs https://github.com/openstax/pattern-library/pull/3

It might be worth uncommenting the ruby code but having it run on only when `Gemfile` exists in the repository (so the script can be reused in other projects)


philschatz:
# To test:

```
# Optionally: ./script/fetch-html --all 

git checkout master
./script/diff-book-prepare --all

git checkout js-sass
./script/diff-book --all

# Should output nothing (which means all the books are equivalent)
```

reedstrm:
I'm addition to reviewing the code changes, @philschatz mentioned that the generated css will not be identical, but should be functionally the same. Baking books (with the use-repeatable-ids) should yield identical baked html

philschatz:
If you use `./script/diff-book-prepare` (in the instructions above) instead of `./script/bake-book` then the `--use-repeatable-ids` flag will be set.

tomjw64:
Sass is supposed to support different output styles. Theoretically, we should be able to adjust options in order to have no change in our CSS output. We currently use `nested` output style while the current changes reflect an `expanded` output style. However, I'm not sure that this implementation of sass allows us to change it?
`$(npm bin)/sass --style compact` or `$(npm bin)/sass --style nested` yield error messages, whereas `$(npm bin)/sass --style expanded` works fine. It seems to me that the only option available is `expanded`, which seems very strange...
@philschatz This seems to be a converted version of dart-sass. Given that this is then technically dart-sass, can you see any of these [behavioral differences](https://github.com/sass/dart-sass#behavioral-differences) becoming an issue (such as supporting only UTF-8)? Luckily, it seems a lot of them happen to be with `@extend` and placeholders, which I don't believe we use.
Additionally, I'm curious as to the reason behind choosing [npm sass](https://www.npmjs.com/package/sass) vs. [npm node-sass](https://www.npmjs.com/package/node-sass), which is built using the seemingly well-supported LibSass library.

philschatz:
Thanks for looking through the differences between `node-sass` and `sass`. I don't have a strong opinion between the two libraries but many of the items listed in [behavioral differences](https://github.com/sass/dart-sass#behavioral-differences) seem to be ones that help reduce "gotchas" in the future.

One thing that I have not liked in sass is the source mapping. Since the dart implementation does not provide source maps, I switched to using `node-sass`.

Unfortunately, `node-sass` required some hacky renaming of files in order to get the source maps correct (see most-recent commit).

script/compile-books
```diff
@@ -32,11 +27,19 @@ for recipeName in "${RECIPE_NAMES[@]}"
...
+
+  # Node-sass accepts an output DIRECTORY, rather than an output FILE
+  # so, in order to support sourcemaps, we have to temporarily
+  # rename book.scss to be ap-physics.scss .
+  if [[ -f "${inputSassFile}" ]]; then
+    rm "${inputSassFile}" || exit 1
+  fi
+  cp "${sassFile}" "${inputSassFile}" || exit 1
+  ${sassBin} --source-map true ${inputSassFile} --output ${outputDir} || exit 1
+  rm "${inputSassFile}"
```
tomjw64: Is the renaming and using an output dir necessary?
```
for recipeName in "${RECIPE_NAMES[@]}"
do

  sassFile="./recipes/books/${recipeName}/book.scss"
  cssFile="./recipes/output/${recipeName}.css"

  ${sassBin} --source-map true --output-style nested ${sassFile} ${cssFile} || exit 1
done
```
seems to work for me

philschatz: Fixed. Thank you for finding that! I totally forgot that the output file is one of the CLI options.

philschatz:
@tomjw64 Good point! I was planning on updating `./script/bootstrap` and `./script/setup` soon to be generic scripts that load python, ruby, and JS. The work on that is going on in https://github.com/Connexions/oer.exports/pull/3047#issuecomment-338322769

I can remove the commented code for now. Any thoughts on having a "generic" script to configure JS, Python, and Ruby?

philschatz:
Hmm. it looks like the shell scripts are not covered (hence the patch failure). 🤔 I wonder if it's worth trying to add https://github.com/SimonKagstrom/kcov to run (by prefixing the scripts in `.travis.yml` with `kcov `). We would probably need to download the built file and run it from some place like https://github.com/SimonKagstrom/kcov/archive/v34.tar.gz since it is not an apt package

philschatz:
Yay, thanks to @reedstrm the scripts are now covered too ❤️  so the `codecov/patch` CI check should be be accurate going forward.

It is an :x: now because the CSS files changed enough that it's difficult for codecov to notice what changed

reedstrm:
Yup the CSS diff is primarily white space (blank lines). There's still something slightly funky about the kcov based script coverage - codecov.io can't show me the actual script files, for some reason.

philschatz:
hmm, clicking on the "Files" tab worked for me: https://codecov.io/gh/Connexions/cnx-recipes/pull/330/tree

reedstrm:
Drill down: https://codecov.io/gh/Connexions/cnx-recipes/pull/330/src/script/bootstrap?before=script/bootstrap

reedstrm:
merging anyway. We'll get his coverage thing licked, yet!

PR #347, opened 2017-10-10
adding python packing to allow for installation of recipes into the db

reedstrm:
This adds python package files, as well as a new `manifest.json` file with the attributes  for each recipe, which is generated when books are compiled, or when `script/build-manifest` is run.
~This includes a `dbmigrator` migration that uses the manifest to read in all the recipes and metadata and stores them. TBH, this pushes the boundaries for a "migration". I did it this way because I want to be able to deploy into just a database, without any services running. The alternative would be to add to the publishing API to provide a method for pushing recipes in that way.~ Migration moved to cnx-db


reedstrm:
depends on Connexions/cnx-db/pull/80

recipes/output/manifest.json
```diff
@@ -0,0 +1,67 @@
+[
```
philschatz: 1. Why is this file in the `output` directory? I would expect that the output directory contains files that can be safely removed. Also, by having it far away from the root it is difficult to find and update.

2. Also, it seems that this is a "config" file... can the data in `books.txt` be combined with this data? That way there is one place to update all the config data for this repository rather than relying on developers to update different files in different places and in different formats. I can help with combining these files if this file is switched to being a dictionary rather than a list.

3. Also, the filename does not seem to indicate that the contents is only an array of elements. changing the root element to be a dict is a minor/easy switch that can be done in the future so I'm not complaining, just musing.

philschatz:
Is there a way to have travis test these python files? Also, I'm worried about the migrations being in this repository since this repository does not have enough information to verify that these migrations work.

recipes/output/manifest.json
reedstrm: I put the manifest in the directory that the files it describes are in. Ideally, this should be a generated file, too - hmm, compile-books is all shell script, as is books.txt.  I suppose we can build it from there. Traditionally a manifest _is_ just a list, not a dictionary/hash/map. What other top-level key would you imagine including?

pumazi:
@philschatz It's hard to test the migrations regardless of where they are located. But I definitely understand your concern.

What we need to settle on is the relationship of these files to the database. It seems that this data is now fundamental to the database, no?

recipes/output/manifest.json
reedstrm: Hmm, I've found that we have recipes that are not referenced in books.txt - odd that.

reedstrm: ok - one: hs-physics 0 we have TEAhs-physics as well

reedstrm:
@pumazi - no this data not fundamental, it is the product of the recipe developement process. Where it is deployed depends on how you're going to use it. Checking out the repo is sufficent to do local cli baking with easybake. Insertion into the DB is needed for publish-based baking. I agree that details of our DB implementation have leaked out here. Now that I've written it, it makes sense to move it to cnx-db, perhaps: have it still import the data from cnx-recipes, making cnx-recipes a soft requirement for cnx-db (to get this migration to run). I purposefully made it a non-deferred, but repeatable migration. If it's not in cnx-recipes, then I probably would make it fully-deferred.

reedstrm:
Note that cnx-db doesn't know how to test data-migrations, either. But it's more likely to grow that capability than this repo, I guess.

recipes/output/manifest.json
reedstrm: Boy, bash is like coding in a straight-jacket, isn't it? I've just about got it.

reedstrm:
I've decide dto move the migrations over to cnx-db. STand by for new updates to both PRs

cnxrecipes/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+
+from ._version import get_versions
+__version__ = get_versions()['version']
+del get_versions
+
+recipes = json.load(resource_stream(__name__, 'recipes/manifest.json'))
+# Convert filenames to strings
+for recipe in recipes:
+    recipe['file'] = resource_string(__name__,
+                                     'recipes/{}'.format(recipe['file'],))
```
philschatz: Should these be an absolute path? Or, from an encapsulation standpoint, should we ever directly need to expose the filename?

cnxrecipes/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+
+here = os.path.abspath(os.path.dirname(__file__))
+migrations = os.path.join(here, 'migrations')
```
philschatz: Is this code still needed?

cnxrecipes/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+del here
```
philschatz: Is there a way to test that this file runs properly? Like 

```python
from cnxrecipes import __version__, recipes;
assert(__version__)
assert('statistics' in recipes) # I botched this one up ; )
```

cnxrecipes/__init__.py
reedstrm: no, python resources have to be relative. You have to have some sort of handle, might as well be the filename. It's not actually exposed as part of the exported interface - that's just a dictionary, with the file property replaced w/ the recipe bits.

cnxrecipes/__init__.py
reedstrm: nope. will delete.

reedstrm:
back at ya @philschatz

philschatz:
Thanks! I think it's mergeable now. I am wondering in the future though:

It seems like this adds several files to the root directory of the repository which makes it a bit cluttered. How unpythonic would it be to move those files somewhere else or would that make non-egg development annoying?

I'm concerned that if this repo were to be used as a python and javascript package then there might be many extra root files. Python would require `MANIFEST.in`, `setup.cfg`, `setup.py`, `versioneer.py`, and `tests/` 

Also, what happens if the `tests/` directory contains non-python files? :crossed_fingers: that they are not a problem.

reedstrm:
We've moved to putting the tests "above" the package dir, on the theory that the tests are not part of the code _per se_. The pytest framework is good at finding the python tests, only, and ignoring non-python files. Personally, I don't find the top-level clutter that annoying - better than having to dig down 4 levels to find the README. (I'm looking at you , zope/plone!)

philschatz:
D'oh, I forgot to do 2 things:

1. check that running this code locally works (it doesn't because `pytest` is not found and so the command silently errors
1. check that the python files are included in code coverage

I will try to address both of those in a followup PR

PR #353, opened 2017-10-23
Implement Katalyst extensions

aiwenar:
This PR add support for `cmlnle:reference` (on `<term>`s) and `cmlnle:case` (on `<link>`s).

#### Terms (8378a04)

Requires Connexions/cnx-easybake#74. Also note that ruby Sass rejects `attr(ns|name)` as a syntax error, and node-sass rewrites it to `attr(ns |name)`, which is incorrect (so #330 wont help). Hence `#{"cmlnle|reference"}`.

#### Links (e41a47f)

Changes `link_setTargetLabels` to accept a map as `$labelText`. This map is expected to provide alternative labels (map values) for a set of cases (map keys), which are then assigned to string named `target-label-{case}`. To remain compatible with `<link>`s without `cmlnle:case` attribute the nominative case is also set as `target-label`.

`link_setLinkLabels` creates additional selectors for `[cmlnle|case="case"]`, which pull the new strings.

For now only cases used in Polish language were added.

Example Sass

```Sass
@include link_setTargetLabels((
    (selector: 'term-old', labelText: "just a string"),
    (selector: 'term-new', labelText: (nominative: "just a string", other-case: "other string"))
));

$GRAMMATICAL_CASES: (nominative, other-case);
@include link_setLinkLabels();
```

Example CSS

```css
term-old { 
    string-set: target-label "just a string";
}
term-new {
    string-set: target-label "just a string";
    string-set: target-label-nominative: "just a string";
    string-set: target-label-other-case "other string";
}

a.classes { content: target-string(attr(href), target-label); }
a.classes[cmlnle|case="other-case"] { content: target-string(attr(href), target-label-other-case); }
```

---

This PR intentionally does not include any translations, as I'm not quite sure where to add them. What I initially tried involved

- changing every string (and template) to a mapping `(locale: value)`,
- providing a clever mixin (called `_`) that would extract appropriate string from that mapping based on current locale,
- repeating many rules for every enabled locale and prefixing them with `:lang(locale)`,

so that enabling inflected links for Polish would look something like

Sass
```Sass
$ENABLED_LANGUAGES: (en, pl);

$PL_FIGURE:   (nominative: "Rysunek", ...);

$Config_TargetLabels_Figure: (selector: 'figure', labelText: (
    en: "Figure " counter(chapter) "." counter(figure)),
    // definition of case_prefixed omitted for brevity. 
    // It just copies first argument appending second to its every value
    pl: case_prefixed($PL_FIGURE, counter(chapter) "." counter(figure), " "),
));

@include localize {
    @include link_setTargetLabels($Config_TargetLabels_Figure);
}

// in link_setTargetLabels replace
    $labelText: map-get($target, labelText);
// with 
    $labelText: map-get(_($target), labelText);
```

CSS
```css
:lang(en) table { 
    string-set: target-label "Figure " counter(chapter) "." counter(figure);
}
:lang(pl) table {
    string-set: target-label "Rysunek " counter(chapter) "." counter(figure);
    string-set: target-label-nominative "Rysunek " counter(chapter) "." counter(figure);
    /* more cases */
}
```

But it required quite a number of changes, and generated rather heavier CSSes, so I'm no longer sure it's a good idea. It would be much easier to just create a new config for each translated style, but I don't know how viable that is.


aiwenar:
Codecov complaints about new inflected links

```css
:pass(7) a.autogenerated-content[cmlnle|case="some-case"] {
  content: target-string(attr(href), target-label-some-case); }
```

If someone can tell me what to change to make tests cover those cases I'll add it.

philschatz:
#330 changes the `sass` implementation to be https://www.npmjs.com/package/node-sass

From a quick test it seems like `node-sass` does **The Right Thing** :tm: :

```sass
// test.scss
body {
  h1::before {
    content: attr(ns|name);
  }
}
```

```sh
$(npm bin)/node-sass test.scss 
body h1::before {
  content: attr(ns |name); }
```

That PR should be merged shortly. Once it is merged you should be able to write CSS normally 😄

aiwenar:
Well, if I'm reading the spec correctly that is exactly not the right thing 😛, as there must be no space between namespace prefix and ‘|’, and that's how I implemented Connexions/cnx-easybake#74.

philschatz:
Oh drat, you are correct. I did not read your PR description closely enough; I was just too excited that there might have been an easier fix ; (

Also, it looks like the code does not have testable documentation. Can you add an entry to https://github.com/Connexions/cnx-recipes/tree/master/recipes/mixins/styleguide that shows how each of the items in the [codecov.io Report](https://codecov.io/gh/Connexions/cnx-recipes/compare/1dc9cdcd7e94e11b3cd8489056359bd74d77ccb8...9e4852153794d451f0f8e1125f4eff0d5e42f976/src/recipes/output/TEAhs-physics.css#L1681) are used? That will add it to the autogenerated styleguide and increase the code coverage

philschatz:
@helenemccarron and @tomjw64 : This adds additional CSS into every recipe.

I am interested to hear your thoughts on this option vs something similar to what we did with the TEA books: Global flags that toggle whether features like these are included in the recipe... that way the CSS files have less unused "code" sprinkled inside

philschatz:
@aiwenar I slightly edited the example file and was unable to see the effect in the styleguide. In your commit, the text "link" is still present (because it is not an `.autogenerated-text` link): [See Example](https://packages.cnx.org/cnx-rulesets/17b23f35d899eaae2bd27513446de1baacf48d7c/statistics/section-page.html#kssref-page-link)

And when I convert the links to not have the link text, then the link is left empty: [See Example](https://packages.cnx.org/cnx-rulesets/ecebc6e7f164f7cb28499c17ab253106b6f02f8a/statistics/section-page.html#kssref-page-link)

Ah, I am guessing that to properly exercise this code, we need to add a `katalyst` recipe that sets the various strings so they show up in the example.

🤔 💭  hmm, I wonder if this code should depend on the presence of an `xml:lang="pl"` attribute in the HTML content

philschatz:
Upon reading your Pull Request, I think the correct approach is the one you mentioned originally in _"What I initially tried involved..."_ ... but I agree that it is not a trivial change.

RE the _"generated rather heavier CSSes"_ comment, I think that each book can set which language it is written for and only that CSS can be generated. That way each CSS file does not contain all of the translations in it.

If getting this PR merged is not critical yet, I would like to try and spend some time making your initial suggestion the way it is done in here.

aiwenar:
I've uploaded those changes as https://github.com/katalysteducation/cnx-recipes/tree/i18n, if you want to look at them. That last commit (65cafc5) brings coverage to 84.4% (it's 6% without it). I'm not sure how to get it to what it was before.

> I slightly edited the example file and was unable to see the effect in the styleguide

I probably should stop writing tests at 10 PM. It can't be healthy.

> I think that each book can set which language it is written for and only that CSS can be generated.

I've made it so that you choose a list of languages per book (that's the `$ENABLED_LANGUAGES` variable in the example above).

> If getting this PR merged is not critical yet, I would like to try and spend some time making your initial suggestion the way it is done in here.

We won't need it before our translation is published, and even after that it'll still be usable without it (it'll just look strange), so I don't think it's critical.

aiwenar:
I've limited the default list of cases to just nominative, since it's the only that has some sense in English, changed `link_setTargetLabels` to always generate link for nominative, and removed remaining cases from styleguide. Now links in guide should have labels generated (both the same unfortunately) and covs should be happy.

philschatz:
I just regression-tested this with the existing books (`./script/diff-book-prepare --all` and `./script/diff-book --all`) and there are no changes.

I would still like to do the following later this afternoon:

1. not have default values for variables (instead, explicitly select a language)
1. add a more concrete example in [recipes/books/_example/_config.scss](https://github.com/Connexions/cnx-recipes/blob/master/recipes/books/_example/_config.scss)

/cc @helenemccarron & @tomjw64 to see what your thoughts on this i18n work is

helenemccarron:
@philschatz, I agree with you, I would much prefer having a language switch that would be applied at the `_config.scss` level. Currently, there are only two languages but if we were to add more, we would have potentially extremely bloated CSS. 

The config file and variable systems were designed specifically with the intention of using one per book and exposing only the hard coded values to be translated or modified, so if a book was in a different language, it could use its own config file with the values required filled out there. 
If additional components are required for a specific book, we typically add a statement for it  such as `$config_hasNominative` to include only what is needed while providing flexibility, in my opinion, we should take advantage of that option in this case.

To address the issue of "viability" of a unique config file, two things are possible: 
 - a brand new config file: the existing config file can be duplicated, modified then maintained separately. The risk is to see the two files diverge based on the needs/updates that are made in each config file.
 - import the existing reference book config file  into a new config file:  make language specific changes there (I think that's what Phil was mentioning above about the TEA books). This minimizes the risk of being out of sync but requires to override all the variables set in the original config file and include extra options.

Based on experience with the TEA books, there is great value in "deriving" a config file and creating specific rules for books that differ from their original format, so that's probably the path I'd choose.

Either way, I think that the `_generator.scss` file can contain localization specific options that are selected in the config file, such as a set of "tools" needed for each language and amend the mixins to support the different cases. In addition, all localization code doesn't need to be in the `_generator.scss` file directly, they could be kept in separate files and imported when the language is selected for clarity and maintainability, similarly to the `_TEAspecific` files.

philschatz:
Thanks for making the [i18n branch](https://github.com/Connexions/cnx-recipes/compare/master...katalysteducation:i18n) it is definitely more intrusive and probably does not make too much sense since each recipe is for a specific book (or 2) so only one language ever applies. I think the approach in this PR is cleaner & simpler. Will review/discuss with @helenemccarron

philschatz:
After chatting through this with @helenemccarron we came up with some... 

# Next Steps

- [ ] add comments to the diff so we remember 6 months from now what the code in `links.scss` does
- [ ] merge the PR
- [ ] fix `compose.scss` in a separate PR so that it only uses the attribute value if it is present
- [ ] come up with a better way to only include language-specific cases that use additional attributes (like `nominative`, etc) when the book requires them

philschatz:
@aiwenar Is there a sample recipe you have that we could include in here for testing that shows the expected behavior?

PR #355, opened 2017-10-30
Update scripts

philschatz:
This updates the scripts to work like https://github.com/Connexions/oer.exports/pull/3047 .

It also:

- updates `./script/bootstrap` to be more efficient when sub-scripts run so the running time for `./script/test` is the same as before.
- adds a linter check for the shell scripts

# TODO

- [x] fix lint errors
- [x] snake_case the variable names in the shell scripts
- [x] follow https://google.github.io/styleguide/shell.xml style
- [ ] follow https://github.com/progrium/bashstyle style


philschatz:
This now includes a fix for https://github.com/Connexions/cnx-easybake/issues/82 (failure installing cnx-easybake because of PyICU)

/cc @helenemccarron

PR #358, opened 2017-11-13
Empty compound composite child pages are no longer titled

tomjw64:
[Errata 4476](https://openstax.org/errata/4476) resolved via recipe change. Additionally, removes delete-me's missed by easybake as part of the end of bake clean up (as it was required to do so to resolve the issue, and also seems like a reasonable thing to do)

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -449,5 +449,8 @@ $_symbolsSectionTitle: "Symbols";
   @include toc_navTOCUnwrap();
 }
 :pass(100) {
+  // Get rid of delete-me objects that easybake misses
+  @include utils_cleanUp();
+  // Make sure trashed objects are deleted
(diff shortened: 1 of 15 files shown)


philschatz:
Does this attempt to fix a bug in `cnx-easybake` (just asking from the message in the PR description)?

kerwinso:
Related Trello card: https://trello.com/c/zu9xLIbn/326-empty-exercise-headers-in-webview

tomjw64:
@philschatz It does attempt to solve an issue with easybake. I would definitely prefer not to, but I could see no good alternative way of removing the header.

reedstrm:
This does not fix the easybake bug - it works around it.

tomjw64:
@reedstrm yes, you're correct. I should have watched my wording a little more closely.

PR #364, opened 2017-12-12
Move hardcoded strings to _common-stuff.scss

aiwenar:
Otherwise we can't translate then into Polish.


philschatz:
This PR has no regressions (tested with `./script/diff-book --all` in about 75min)

PR #367, opened 2017-12-13
Properly sort terms with cmlnle:reference in index

aiwenar:
With #353 it is now possible to have terms with different text in index than in document, but they are still grouped according to their document text, since `group-by` attribute was still being set from text.

The additional pass and temporary `span[data-type="term-content"]` are required because we can't regex-match on values of attributes.


philschatz:
This PR has differences in the Index (tested with `./script/diff-book statistics`)

`diff ./data/statistics-prepared.xhtml ./data/statistics-baked.xhtml`

```diff
283060c283060
<                   group-by=" Symbols"
---
>                   group-by="&#x3B1;"
489745,489764d489744
<           >Symbols</span
<         ><div
<             class="os-index-item"
<           ><span
<               group-by=" Symbols"
<               class="os-term"
<             >&#x3B1;</span
<           ><a
<               href="#auto_d8133eb3-69f8-41af-a26a-37b59a4ba156_term186"
<               class="os-term-section-link"
<             ><span
<                 class="os-term-section"
<               >9.4 Rare Events, the Sample, Decision and Conclusion</span
<             ></a
<           ></div
<         ></div
<       ><div
<           class="group-by"
<         ><span
<             class="group-label"
492926a492907,492926
>       ><div
>           class="group-by"
>         ><span
>             class="group-label"
>           >&#x391;</span
>         ><div
>             class="os-index-item"
>           ><span
>               group-by="&#x3B1;"
>               class="os-term"
>             >&#x3B1;</span
>           ><a
>               href="#auto_d8133eb3-69f8-41af-a26a-37b59a4ba156_term186"
>               class="os-term-section-link"
>             ><span
>                 class="os-term-section"
>               >9.4 Rare Events, the Sample, Decision and Conclusion</span
>             ></a
>           ></div
>         ></div
```

aiwenar:
This change is actually expected, since ‘α’ is in fact a word character (`/\w/`). Unfortunately I see no way of avoiding changes to generated XHTML for your books without changing their source. If that's acceptable you should be able to force certain terms to be treated as symbolic by prepending their index form with a zero width symbol, say U+2060:

```xml
<term cmlnle:reference="&#8288;α">α</term>
```

philschatz:
Oh, interesting hack!

But I would rather not add zero-width characters... I will try changing the english recipes to use `/[a-z]/` instead of `/\w/`.

/cc @reedstrm

philschatz:
Now statistics works but it seems there is a new problem in `econ`. I am not sure why this is happening; the code change does not seem to indicate that this would happen. The `book > metadata > description` is now copied into every Page metadata. This did not happen on `master`


```diff
...
10413,10446d10412
<               class="description"
<               data-type="description"
<               itemprop="description"
<             >
<         
<         <div
<               ><div
<                 >
< <p
<                   ><em
<                       data-effect="italics"
<                     >Principles of Economics</em
<                   > covers the scope and sequence for a two-semester principles of economics course. The text also includes many current examples, including; discussions on the great recession, the controversy among economists over the Affordable Care Act (Obamacare), the recent government shutdown, and the appointment of the United States&#x2019; first female Federal Reserve chair, Janet Yellen.</p

...
```

philschatz:
@aiwenar there still seems to be a regression where the Book metadata description is being copied into each Page.

Maybe @helenemccarron could help in finding out why the Book metadata is being copied over.

recipes/books/_generator.scss
```diff
@@ -230,6 +230,10 @@
   }
 }
 :pass(3) {
+  // After: modify_prepIndexTerms
+  @include modify_finishPrepingIndexTerms($Config_Index_NotSymbolRegexp, $Config_Index_SymbolRegexp);
+  @include utils_clearTrash();
```
helenemccarron: I believe this line is causing the descriptions to be trashed. On l.106 above it looks like we tried to remove them unsuccessfully
```
  //the metadata has been moved to the content, remove it from the metadata
  @include modify_trash('[data-type="metadata"] > [data-type="description"]');
``` 
Maybe this is actually a desirable change? I don't remember if we need to keep the descriptions in the metadata or not. @reedstrm or @tomjw64 may recall/know.

aiwenar: Yes, the `utils_clearTrash` call is the culprit. I haven't submitted change to this because either removing, or changing to an equivalent that affected only the nodes I needed, caused it to fail converting our book correctly.

aiwenar:
Rebased on master and fixed regressions in your books. Output is different then it was for out book and it'll take me some time to check if it's OK (~30k lines diff), but terms with `cmlnle:reference` are being sorted correctly, so I don't think it's blocking merge.

helenemccarron:
Great, let's merge it, we can always open a separate PR to remove the descriptions to keep things clean.

kerwinso:
I think @philschatz said this one was regression test only.

PR #368, opened 2017-12-15
Update baking language used in recipes

philschatz:
This changes the recipes to use the language in `css-plus` rather than `cnx-easybake`.

# Approach

This PR walks through `_generator.scss` and slowly converts the mixins in the order they are presented in the file. Commits with 📝  attempt to change 1 or 2 mixins at a time to make it easier to follow the changes.

### Pros

- easy to see how the language changes
- ensures that the changes work for all recipes

### Cons

- work that is split among many mixins (because they need to run in different passes) no longer needs to be so they should probably be combined


philschatz:
# Current Problem (& options) for Discussion

It is not clear to me how to describe the following: "Remove even-numbered solutions and link odd-numbered exercises to their solutions"

```less
$IS_ODD_EXERCISE: is-odd(count-of-type('chapter', 'exercise'));

// This approach does not require any additional changes.
// But the syntax is a little awkward because it uses `if(...)` and display:none; to discard work that was done
exercise:has(> solution) {
  // This element is the exercise number (ie "4.3") and is either a link (odd-numbered) or just bold (even-numbered)
  &::before {
    // make the link to the solution if odd-numbered, and just make a number if even-numbered
    tag-name-set: if($IS_ODD_EXERCISE, "a", "strong");
    // We need to add the href if it's an odd-numbered exercise but not add it if it is an even-numbered exercise.
    // The only way I can think to do it is to add the attribute and then only remove it if
    // the exercise is even-numbered.
    attrs-add: "href" "#" descendant-context('> solution', attr(id));
    attrs-remove: if($IS_ODD_EXERCISE, , "href");

    // The number "4.3"
    content: count-of-type('body', 'chapter') "." count-of-type('chapter', 'exercise');
  }
  > solution {
    display: if($IS_ODD_EXERCISE, default, none); // Discard the solution if it is even-numbered
  }
}


// This approach creates a new pseudo-class `:x-is(...)` which is supposed to be
// similar to `:nth-of-type(...)` but allows functions in the selector which is a bit odd
// for CSS.
exercise:has(> solution) {
  &:x-is($IS_ODD_EXERCISE) {
    &::before {
      tag-name-set: "a"; // make the link to the solution
      /* ... */
    }
  }
  &:not(:x-is($IS_ODD_EXERCISE)) {
    > solution {
      display: none; // discard the solution
    }
  }
}


// This approach makes `:x-is(flagName)` which is just a boolean variable that is
// set via `flags-add: flagName $trueFalseValue;`
//
// This approach is a little weird because we need to evaluate the Declarations
// and then evaluate selectors again which is difficult to do currently.
// On the plus side, it does allow the functions to be in a more "natural" place
// since they are not inside the selector
exercise:has(> solution) {
  flags-add: flagName $IS_ODD_EXERCISE;
  &:x-is(flagName) {
    &::before {
      tag-name-set: "a"; // make the link to the solution
      /* ... */
    }
  }
  &:not(:x-is(flagName)) {
    > solution {
      display: none; // discard the solution
    }
  }
}
```

PR #370, opened 2017-12-20
Introduction to Business Recipe and StyleGuide

brittany-johnson:
Testy McTestFace - CTE Assignable Webview Ruleset/Recipe bookmaps: 
https://docs.google.com/spreadsheets/d/1QM28dLjuEH3xPtVty9l7Zfeqp4mg4fJtIAzv-oXH5Ik/edit#gid=44631156

Business Series Testing and Requirements:  https://docs.google.com/spreadsheets/d/1CY56IPhwap2aR_t2VQYsorOMp_8AxLaUiGywtaPscNs/edit#gid=606798872


recipes/books/intro-business/styleguide/page.note.link-learning.snippet.xml
```diff
@@ -0,0 +1,11 @@
+<div xmlns="http://www.w3.org/1999/xhtml">
+  <div data-type="chapter">
+    <div data-type="page">
+      <!-- Markup starts here -->
+      <note xmlns="http://cnx.rice.edu/cnxml" class="lab">
```
philschatz: I think there is a simpler way to write this snippet so that it only contains CNXML. Something like:

```xml
<hack-root xmlns="http://cnx.rice.edu/cnxml">
  <!-- START:Chapter -->
  <!-- START:Page -->
  <!-- Markup starts here -->
  <!-- Markup starts here -->
  <exercise type="check-understanding">
    <problem>
      <para><emphasis>SHOULD_REMAIN</emphasis> and this is the check-understnding exercise (it should be unnumbered)</para>
    </problem>
    <solution>
      <para>No, the coach&#x2019;s new stopwatch will not be helpful. </para>
    </solution>
  </exercise>
  <!-- Markup ends here -->
  <!-- END:Chapter -->
  <!-- END:Page -->
</hack-root>
```

Also, I added this just so you can try out the https://atom.io/packages/pull-requests plugin 😄

recipes/mixins/_compose.scss
```diff
@@ -287,6 +287,7 @@
         }
       }
     } @else {
+      $sourceSelector: if($specialPageType == $PAGE_CITATIONS, '[data-type="note"].#{$className}', 'section.#{$className}');
```
TxsAdamWest: What are l.290 & l.250 doing?   Searching for elements with those data-type values? 🤔

.gitignore
```diff
@@ -5,6 +5,8 @@
 /bin/
 /lib/
 /include/
+.DS_Store
```
philschatz: @brittany-johnson Can you add this to your `~/.gitignore` file (in your home directory)? These files are OS-specific and that way these files are excluded for every repository.

recipes/books/_common-stuff.scss
```diff
@@ -46,6 +46,37 @@ $Config_TargetLabels: (
   (selector: '.appendix > figure:not(.unnumbered)',                             labelText: "Figure " counter(appendix, upper-alpha) counter(figure))
 );
 
+
+// All of these should be allowed to be empty.
```
philschatz: I am concerned with too many defaults because then it is difficult to remember what the default is for every configuration field. 

What was the problem that caused these to be beneficial?

recipes/books/_common-stuff.scss
```diff
@@ -46,6 +46,37 @@ $Config_TargetLabels: (
...
+$Config_Coverage_MayHaveIframes: true;
+$Config_Coverage_MayHaveMissingExercises: true;
+
+// Everything enabled on captions by default. Better to have too much than nothing at all.
+$Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: true, hasTitle: true);
+$Config_SetFigureCaption: (captionType: $CAPTION_FIGURE, defaultContainer: figcaption, hasCaption: true, hasTitle: true);
+
+
+//if the book has citations (references)
+$Config_hasCitation: false;
```
philschatz: Can you include this in the `books/.../README.md` file? This way all config fields are documented

recipes/books/intro-business/book.css
```diff
@@ -0,0 +1,89 @@
+/*
+Error: $map: ((hasCaption: false)) is not a map for `map-get'
```
philschatz: It looks like there was a SASS error that should be fixed.

recipes/books/intro-business/styleguide/book.composite.crit-think.snippet.xml
```diff
@@ -0,0 +1,11 @@
+<div xmlns="http://www.w3.org/1999/xhtml" data-type="chapter">
```
philschatz: Can you use the CNXML-based snippet format instead? Example: https://github.com/Connexions/cnx-recipes/blob/master/recipes/mixins/styleguide/page.exercise.snippet.xml

```xml
<hack-root xmlns="http://cnx.rice.edu/cnxml">
  <!-- START:Chapter -->
  <!-- START:Page -->
  <!-- Markup starts here -->
  <exercise id="ex1">
    <problem>
      <para>If an apple fell 3 meters in 2 seconds, what was the speed of the apple?</para>
    </problem>
    <solution>
      <para>It fell 1.5 meters per second</para>
    </solution>
  </exercise>
  <!-- END:Page -->
  <!-- END:Chapter -->
</hack-root>
```

recipes/mixins/_number.scss
```diff
@@ -336,6 +336,22 @@
...
 
+
+@mixin number_citations($citationCounter, $citationLinkCounter ) {
+  [data-type="cite"] {
+    &::before {
+      container: span;
+      class: 'os-citation-number';
+      content: counter($citationLinkCounter);
+    }
+    .reference::before {
```
philschatz: Is this class in the CNXML content, only in the HTML content, or a class that was added as part of the recipe generation? If it is in the content then I guess this is ok but an Issue should probably be created in https://github.com/Connexions/cnxml to change it to not be a class.

styles/styles.css
```diff
@@ -79,7 +79,7 @@ section, .figure {
   > [data-type="footnote-refs"] > ol > li::before {
     content: "" !important; }
 
-[data-type="chapter"] {
+[data-type="chapter"], [data-type="composite-page"], [data-type="page"] {
```
philschatz: What is this file used for? If it is for TEA then should they be notified of changes?

helenemccarron: This is purely for local development purposes and because we had to remove and add the file each time we switched branches. This is what the devs and Alan are using not to kill our eyes staring at vanilla HTML during recipe creation and facilitate testing (a gray box for feature is easier to spot than a continuous chunk of text, for example?)

recipes/mixins/_number.scss
helenemccarron: This is how the citations were handled in American Government. The raw looks like this 
```
<a xmlns="http://www.w3.org/1999/xhtml" data-type="cite" href="#auto_844f199e-2df1-4159-836d-e0f759968a16_rf-1"><div class="reference" data-type="note" display="inline" id="auto_844f199e-2df1-4159-836d-e0f759968a16_rf-1">Reference data here.</div></a>
```
@reedstrm  implemented it a while back and can speak more to the reasons behind the final results.

recipes/books/_common-stuff.scss
helenemccarron: @brittany-johnson and @TxsAdamWest were having issues generating recipes from scratch without defaults. This allows to create a config file from scratch without having to duplicate an existing one, so I see value in it. I also think that if all the defaults are in one location (common), all it takes is to check those when needed. We could add those to the documentation as well if that would help.

helenemccarron:
The failed css file isn't used, we only use `output/intro-business.css`, I removed the offending file. As far as I could find, all the collations, notes and special opening section have been included.

helenemccarron:
The broken css file shouldn't have been committed, we only use `output/intro-business.css` so this should have no incidence. I have removed it. As far as selectors are concerned, currently, all notes, all EOC collations and the EOB references have been added, including the opening section for learning outcomes.

recipes/mixins/_compose.scss
helenemccarron: This is setting the source selector if the `specialPageType` is a `$PAGE_CITATIONS` to `'[data-type="note"].#{$className}'`, otherwise, it sets it to ` section.#{$className}`

recipes/mixins/_number.scss
philschatz: What does the CNXML look like? Does it contain the `class="reference"`? If so, it seems like the class name is custom to the book and should be a configuration field, no?

recipes/books/_common-stuff.scss
philschatz: It could be a separate PR. I just didn't want to have people need to remember to update the `recipes/README.md` whenever a new config field is introduced.

.gitignore
philschatz: Yes, I was suggesting that folks update the `~/.gitignore` (in your home directory) which is independent of any repository

recipes/mixins/_number.scss
helenemccarron: This is the markup for the citation. `<cite target-id="rf-1"><note class="reference" display="inline" id="rf-1">Reference data here.</note></cite>` Before we change it, we would have to test in thoroughly because the conversion is done via XSLT and there's quite a bit of template work done to support the functionality in the PDF

.gitignore
helenemccarron: ah, gotcha!

recipes/mixins/_number.scss
philschatz: I was just suggesting that the class name (`reference`) be a `$Config_*` parameter since it applies to a book and since all class names in content are specified in config parameters (like the notes or sections that need to be collated)

helenemccarron: Mhh, I don't know if it would work with a different class, @reedstrm could let us know. I understand the structure is pretty rigid for it to work as intended.

PR #374, opened 2018-01-09
Update Books.txt

TxsAdamWest:
This PR is for changes to books.txt such as adding new books, editing book list and changing server names.


philschatz:
From jenkins, it looks like `nvm install v6.9.1` is no longer available. I will try v8.9.4 (from https://nodejs.org ) and see if that helps

philschatz:
I updated the node version. Is there anything else blocking this from merging?

TxsAdamWest:
Unsure.   I believe we are waiting until we have business books pushed to devb before updating the booklist.txt with new titles / uuid's & committing it.  @helenemccarron Can you confirm?

helenemccarron:
No, I missed this, I'll merge it now.

Issue #375, opened 2018-01-19
Wrong note titles in non-English modules


aiwenar:
# Scenario

Notes with non-default type and without explicit label are labelled in English, regardless of the document language.

# Reproducibility Details

- **Impact area**: 
- **User**: 
- **URL / Location**: katalyst01.cnx.org, but likely every CNX instance
- **How many times were you able to reproduce the issue?** always

## Steps to reproduce:

- **Step 1**: Create a non-English module with a typed note.
- **Step 2**: Publish it.

## Screenshots

![image](https://user-images.githubusercontent.com/385442/35149688-5cd37e9e-fd17-11e7-9b65-aa770e1d875a.png)

# Environment:

- **Instance**: katalyst01.cnx.org

# Impact

- **Severity**: 
- **Frequency of occurrence**:

aiwenar:
In the example above a `<note type="warning">` has been given title of _warning_, even though the document is in Polish and the correct world is _ostrzeżenie_. This happens because [CNX to XHTML conversions set `data-label="note-type"`](https://github.com/Connexions/rhaptos.cnxmlutils/blob/master/rhaptos/cnxmlutils/xsl/cnxml-to-html5.xsl#L505-L514), and it's then used by webview to display label:

```css
[data-label]:not([data-label=""]):not(.ui-has-child-title):not(.btn-link)::before {
    content: attr(data-label) ': ';
}
```

(Both Firefox and Chrome fail to point to where this rule is defined)

This seems to be workable around with recipes

```sass
$_note_types: (
  warning: "ostrzeżenie",
  tip: "podpowiedź",
  important: "ważne",
);


:pass(20) [data-type="note"] {
  @each $type in map-keys($_note_types) {
    $title: map-get($_note_types, $type);

    &[data-element-type="#{$type}"] {
      data-label: $title;
    }
  }
}
```

but I'm still waiting to verify whether lack of change mean this didn't work, or just caching.

reedstrm:
example URL: https://katalyst01.cnx.org/contents/T0X24ak3@7.4:-mY69h8v@3/41-Przemieszczenie-i-prdko

reedstrm:
The recipe-fix worked, once we got around which recipe was being used. Probably should be an issue against rhaptod.cnxmlutils cnxml -> html transforms for putting english attributes in regardless of the document language.

openstaxalina:
Although this issue is in cnx-recipes, it's likely related to cnxml --> html transforms.

[...]

PR #377, opened 2018-01-29
WIP: Add recipe for Katalyst Education's translation of University Physics

aiwenar:
This PR also changes node-sass to use [at-import/node-sass-import-once](https://github.com/at-import/node-sass-import-once). We need it to be able to overwrite settings from `_common-stuff.scss` before it gets imported by `u-physics`'s `_config.scss`.

I'll add CSS for the recipe after all issues have been resolved and I've rebased it.

### Depends on

- [x] #353 Implement Katalyst extensions
- [x] #364 Move hardcoded strings to _common-stuff.scss
- [x] #367 Properly sort terms with cmlnle:reference in index
- [x] #375 Wrong note titles in non-English modules
- [x] Connexions/rhaptos.cnxmlutils#164 Fix handling of nested terms in glossary
- [ ] (no issue yet) A simple PDF style in oer.exports


aiwenar:
I've been working on merging this PR with master, and unfortunately that causes a lot of changes to baked book (diff had at least 50000 lines, but it might actually be in hundred thousands). I'm now looking into some XML diffing tools to make sense of it, but this is what I found so far:

- [ ] `class="os-toc-counted-page"` added to TOC entries for appendices
- [x] `group-by` attribute on terms changed form first letter of content to first letter of their reference form (#367) - this should have been included in the original recipe
- [ ] Accessibility alt-text in tables replaced with contents of table caption (cabdde3? from #417)
- [ ] `<div class="os-table">` added around some (not all) tables (8a9888d? from #611)
- [ ] Some `span.os-divider`s changed from `.` to `. ` (space after colon) (somewhere in #611 but can't pinpoint)
- [x] One of `<div data-type="metadata">`s is gone - this is probably a non-issue, since it's never displayed to the user

(Checked items are either resolved or non-issues)

Note that I can't determine exactly which commit caused each change, as it would seem that you're squashing all commits from each sprint into a single one, so even if I find code responsible for a change it only gives us which sprint it came from, not which commit (those two commits I pointed to above aren't even present in a checkout, only on GitHub).

aiwenar:
After a talk with @helenemccarron we've decided to implement all functionality required for Polish recipe in small PRs separate from this one, and then re-create the Polish recipe from scratch, deprecating this PR.

PR #379, opened 2018-02-02
Don't fail when people don't have nvm, but have proper node version

aiwenar:
I personally use [nodenv](https://github.com/nodenv/nodenv) and thus need to remove nvm check every time I switch a branch, which is really annoying.

PR #380, opened 2018-02-14
Ensure pytest runs

philschatz:
This fixes 2 things:
- `pytest` was not installed as part of `./script/setup` 
- When missing, it silently failed instead of erroring


kerwinso:
found while testing #379

PR #382, opened 2018-03-12
Require python version 2.7 in virtual env

brittany-johnson:
In my virtual env, I ran into an issue where the latest version of python was running. This gave me issues when trying to generate the book. 

This line requires python 2.7 in the virtual env, and allowed me to generate my book once again without issues.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -3,7 +3,7 @@
 
 # Resolve all dependencies that the application requires to run.
 # From https://github.com/github/scripts-to-rule-them-all/blob/master/script/bootstrap
-virtualenv -p /usr/bin/python2.7
+
 if [[ $0 != "-bash" ]]; then
   cd "$(dirname "$0")/.." || exit 111
 fi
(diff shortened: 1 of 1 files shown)


philschatz:
It might be worth having others run this to see if it is just me having issues.

script/bootstrap
```diff
@@ -102,7 +102,7 @@ if [[ ! ${BOOTSTRAP_ALREADY_RAN} ]]; then
       fi
       if [[ ! -d "./venv/" ]]; then
         do_progress_quiet "Setting up a python virtualenv" \
-          virtualenv ./venv/
+          virtualenv --python=python2.7 ./venv/
```
pumazi: I'd recommend doing ``... --python=`which python2.7` ...``

philschatz: @brittany-johnson : 👍 I think that @pumazi's suggestion would fix the problem I ran into. To keep it syntactically similar to the rest of the shell scripts, it might be better to write `... --python=$(which python2.7) ...`

brittany-johnson: @philschatz @pumazi Thanks! Just pushed the change. Did this fix the problem that you ran into?

philschatz: It did not work for me. It looks like the `--python=...` argument is correct but something still tries to use `/usr/local/opt/python/bin/python2.7`

```
$ ./script/setup && ./script/test 
==> Setting up a python virtualenv ... 
./script/bootstrap: /usr/local/bin/virtualenv: /usr/local/opt/python/bin/python2.7: bad interpreter: No such file or directory
./script/setup: ERROR: could not run [virtualenv --python=/usr/bin/python2.7 ./venv/]
```

Any thoughts @pumazi

philschatz: reinstalling virtualenv fixed it for me!

The last thing to do is to fix the [Travis failure](https://travis-ci.org/Connexions/cnx-recipes/builds/353430095#L5896) and then it's ready to merge!

brittany-johnson:
@philschatz Travis looks happy now. Am I in the clear to merge?

philschatz:
Oh, it looks like you will run into a similar problem the next time you try to run a script in `oer.exports`... According to https://brew.sh/2018/01/19/homebrew-1.5.0/ it seems like there might need to be a different way of fixing this problem

philschatz:
I opened #383 to hopefully fix this by using pyenv which is consistent with how we install ruby and node using the scripts (using `rbenv` and `nodenv` respectively). We can meet about the approaches tomorrow and I will try to open corresponding Pull Requests for `cnx-easybake` and `oer.exports`

philschatz:
Is there a decision on whether to proceed with this PR or #383?

brittany-johnson:
@philschatz I will probably just close this. I did just comment on #383 because I'm having issues again :(

PR #383, opened 2018-03-14
Use pyenv to specify python version

philschatz:
Python 3 is now the default for Homebrew installs but we still need 2.7. See https://brew.sh/2018/01/19/homebrew-1.5.0/

This changes the scripts to use `pyenv`. We already use `rbenv` and `nodenv` to configure ruby and node so it seemed natural to configure python similarly.

Refs #382


script/bootstrap
```diff
@@ -88,24 +98,46 @@ if [[ ! ${BOOTSTRAP_ALREADY_RAN} ]]; then
...
-        if [[ -f "Brewfile" ]] && [[ "$(uname -s)" = "Darwin" ]]; then
-          do_progress_quiet "Unlinking homebrew python (just-in-case)" \
-            brew unlink python
-          do_progress_quiet "Relinking homebrew python" \
-            brew link python
+      virtualenv_dir="./venv/"
+      python_version="$(< .python-version)"
+
+      # Check that a .python-version file is set
+      [[ -f "./.python-version" ]] || die "Repository is missing a .python-version file. Please add one"
```
helenemccarron: Should we document how to create that file and what it should contain?

philschatz: Would "Repository is missing a .python-version file. Please add one. The contents should contain the python version (ie `2.7`)" work?

helenemccarron: Yes, that sounds good. Would the read me specify which version of Python we are currently supporting as this may change?

philschatz: What if the README said something like "See `./.python-version` for the supported python version"? That way we would not have to remember to update the README when the python version changed?

brittany-johnson:
@philschatz I'm now getting this error again when trying to bake a book. Very strange. I've tried running the clean and setup scripts to fix this issue, but I did not have any success when trying to bake a book. I am on the use-pyenv branch and I'm up to date with the latest commit. 
screen shot 2018-03-23 at 12 44 04 pm

philschatz:
Is that the same error that you ran into which caused the need to explicitly specify python2? It seems like this is a different error... maybe an Issue with `cnx-easybake`?

/cc @reedstrm

brittany-johnson:
This confirms that pyenv is using the correct version of python, correct? @philschatz 
screen shot 2018-03-23 at 12 53 03 pm

brittany-johnson:
Python 2.7 is also installed globally 
```
Brittanys-MacBook-Pro-2:lib bj20$ which python2.7
/usr/bin/python2.7
```
But it looks like python3.6 in /cnx-recipes/venv/lib/python3.6 is still being used in someway? 
I was getting this traceback before 
```
TypeError: startswith first arg must be str or a tuple of str, not bytes
```
And I _think_ thats what tipped @reedstrm that there was a conflict with the python version being used.

philschatz:
I have one more fix for this.... I was getting an error when running `pip install virtualenv`

brittany-johnson:
```Brittanys-MacBook-Pro-2:cnx-recipes bj20$ ./script/bake-book intro-business
==> Activating Python virtualenv ... OK
==> Installing correct version of node ... OK
==> Compiling CSS file ... OK
==> Baking ./data/intro-business-raw.xhtml using intro-business (~1min) ... 
  File "/Users/bj20/openstax/cnx-recipes/venv/bin/cnx-easybake", line 11, in <module>
    load_entry_point('cnx-easybake', 'console_scripts', 'cnx-easybake')()
  File "/Users/bj20/openstax/cnx-recipes/venv/src/cnxeasybake/cnxeasybake/scripts/main.py", line 83, in main
    args.coverage_file, args.use_repeatable_ids)
  File "/Users/bj20/openstax/cnx-recipes/venv/src/cnxeasybake/cnxeasybake/scripts/main.py", line 19, in easybake
    oven = Oven(css_in, use_repeatable_ids)
  File "/Users/bj20/openstax/cnx-recipes/venv/src/cnxeasybake/cnxeasybake/oven.py", line 111, in __init__
    self.update_css(css_in, clear_css=True)  # clears state as well
  File "/Users/bj20/openstax/cnx-recipes/venv/src/cnxeasybake/cnxeasybake/oven.py", line 164, in update_css
    rules, _ = tinycss2.parse_stylesheet_bytes(css, skip_whitespace=True)
  File "/Users/bj20/openstax/cnx-recipes/venv/lib/python3.6/site-packages/tinycss2/bytes.py", line 110, in parse_stylesheet_bytes
    css_bytes, protocol_encoding, environment_encoding)
  File "/Users/bj20/openstax/cnx-recipes/venv/lib/python3.6/site-packages/tinycss2/bytes.py", line 33, in decode_stylesheet_bytes
    if css_bytes.startswith(b'@charset "'):
TypeError: startswith first arg must be str or a tuple of str, not bytes
./script/bake-book: ERROR: could not run [cnx-easybake --coverage-file ./data/intro-business-baked.xhtml.lcov ./recipes/output/intro-business.css ./data/intro-business-raw.xhtml ./data/intro-business-baked.xhtml]
```
I'm still getting the same error. I pulled the latest and ran the clean and setup scripts again. I then tried to bake a book and got the error above.

philschatz:
@brittany-johnson : Do you still get this error when you run the code in https://github.com/Connexions/cnx-recipes/pull/382 ? 

I think that the error you are running into might be a bug in `cnx-easybake` (and is unrelated to this PR) but a quick look by @reedstrm 's 👀  might help confirm that.

brittany-johnson:
@philschatz No, I don't get the error when I run the code in #382 .

philschatz:
ooh, I failed to read the stack trace accurately: `File "/Users/bj20/openstax/cnx-recipes/venv/lib/python3.6...` . Yep, that's a bug. I wonder why it does not happen to me locally.

PR #384, opened 2018-03-21
WIP - Style stuff

helenemccarron:
## Style Baked PDF
- convert the Business theme into a modular framework with a scheme and a color map to style Baked PDF
- create a default style to use as a base for other themes

## Add Styles documentation to KSS
- add a source to KSS to read comments form `./styles/`. The purpose of using comments from `./styles` is to document global styling principles such as typography or tables designs along with the structural documentation provided in `./recipes`. 

## Style content in KSS
- use the actual template CSS to style the styleguide snippets. Styles apply to both recipes and design elements

## Style KSS itself
- make KSS a bit easier on the eye and match our brand more closely so it can be a tool for vendors to use.

## Install KSS in the setup script
use `npm install kss` in the meantime

### To view: 
- **Styleguide**: generate guide and view styles in browser. 
`./script/generate-guide intro-business`

- **Baked HTML**: Add `<link rel="stylesheet" type="text/css" href="../styles/output/intro-business.css" /> ` to the baked file or bake with that link and view in the browser.
 `./script/bake-book intro-business`

### TODO
- [x] Get Travis and Jenkins to Pass 😄  so _anyone_ can take a look at the Styleguide for themselves without having to run `./script/generate-guide` on their machine
- [x] update KSS to use same (or newer) version as pattern library
- [ ] experiment with using markdown inside of comments to document non-recipe sections
- [x] separate color map
- [ ] add resources maps
- [x] rename variables (`CamelCase_partialname: ""`)
    note: because we are using functions, this is redundant: `color-map(someName-color_map)`
- [ ] fix intro-business recipe to not use references when none
- [ ] transfer data from openguide, test linking + screenshots
- [ ] _Maybe?_ scope the template CSS file so that it is only in the rendered blocks in the Guide
- [x] _Maybe?_ fix the empty-title (see `[PAGE_TITLE]`) to show up and be more human-friendly (e.g. `Example Page Title`)
- [ ] fix Page Element to display title (`[PAGE_TITLE]` was replaced)
- [ ] document the include order (config files vs CSS cascading) system
- [ ] fix the recipe to link properly


philschatz:
Nice! I wanted to take a look at the styleguides but it seems Jenkins failed because there are some files missing. See [Travis Error](https://travis-ci.org/Connexions/cnx-recipes/builds/356589313#L1251)

helenemccarron:
hum, that's actually a good question: there is no `./styleguide/1-global.typography.xhtml-baked.html` because this is puling the snippet from the `styles` and not the `recipes`. We need global style specifications in the styleguide, such as typography, folio, table colors etc. The styleguide will work as is without the baked file for typography.

js/kss/builder.js
```diff
@@ -0,0 +1,137 @@
+'use strict';
+
+/**
```
philschatz: Is this file copied from somewhere? If so, can you include the source in a JS comment around here?

If this file was customized, can there be include comments to point out where it was customized? This will help if we want to upgrade kss so we know where our changes were made.

js/kss/index.hbs
```diff
@@ -0,0 +1,275 @@
+<!DOCTYPE html>
```
philschatz: Is this file copied from somewhere? If so, can you include a link to where it came from?

If this file was customized, can there be include comments to point out where it was customized? This will help if we want to upgrade kss so we know where our changes were made.

js/kss/kss-assets/kss-fullscreen.js
```diff
@@ -0,0 +1,59 @@
+(function (window, document) {
```
philschatz: Is this file copied from somewhere? If so, can you include a link to where it came from?

If this file was customized, can there be include comments to point out where it was customized? This will help if we want to upgrade kss so we know where our changes were made.

js/kss/kss-assets/kss-guides.js
```diff
@@ -0,0 +1,26 @@
+(function (window, document) {
```
philschatz: Is this file copied from somewhere? If so, can you include a link to where it came from?

If this file was customized, can there be include comments to point out where it was customized? This will help if we want to upgrade kss so we know where our changes were made.

js/kss/kss-assets/kss-markup.js
```diff
@@ -0,0 +1,40 @@
+(function (window, document) {
```
philschatz: Is this file copied from somewhere? If so, can you include a link to where it came from?

If this file was customized, can there be include comments to point out where it was customized? This will help if we want to upgrade kss so we know where our changes were made.

script/compile-books
```diff
@@ -20,7 +20,7 @@ for recipe_name in "${RECIPE_NAMES[@]}"; do
   sass_file="./recipes/books/${recipe_name}/book.scss"
   css_file="${output_dir}/${recipe_name}.css"
 
-  do_progress_quiet "Generating ${css_file}" \
+  do_progress_quiet "Gåenerating ${css_file}" \
```
philschatz: Testing unicode?

js/kss/builder.js
philschatz: oh ok. In that case do those files actually need to be stored in `git`, or could they be re-built when the user types `./script/setup` (and added to `.gitignore`)?

Whatever command you used to create the custom builder could be included in `./script/_setup.sh` so that it runs whenever someone runs `./script/setup`

helenemccarron: I need to dig into what I did but I think we will need to customize some of those files, which would make it impossible to reinstall on `setup`. If that's the case, I agree with you we should rename the folder and document the customizations made to the files.

PR #416, opened 2018-06-05
Inline the 'display: none' style for metadata

tomjw64:
After a brief discussion between Helene and I, we figured that since we are trying to make the books read acceptably even without stylesheets, we would inline some style to `display: none;` the metadata in baked books, since we can't think of a case in which we would not want them to be hidden.

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -455,6 +455,7 @@
 :pass(10) {
   //After: All metadata tasks are completed
   @include modify_suppressURI($Config_hasCompositeChapter);
+  @include modify_hideMetadata();
   @include toc_prepTOCElements($Config_hasCompositeChapter);
 }
 :pass(11) {
--- a/recipes/mixins/_modify.scss
+++ b/recipes/mixins/_modify.scss
@@ -60,6 +60,11 @@
   }
 }
 
+@mixin modify_hideMetadata() {
+  [data-type="metadata"] {
+    attr-style: "display: none;";
+  }
+}
 
 
 // Replace module titles with the ones set by collection.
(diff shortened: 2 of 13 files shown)


helenemccarron:
I think it should be _present_, not necessary visible. It has to be hidden in all cases to display the content in a meaningful manner, and it cannot be removed.

tomjw64:
Perhaps it's info that's used elsewhere? I don't know where all it is used, but Ross certainly does.

philschatz:
But what is the reason for having it be present but not visible to anyone? How would it be different than making the metadata an HTML comment? (or removing it altogether)

tomjw64:
I believe it has something to do with how webview renders a baked book and/or how the epub is built. e.g. epub will not validate if metadata is not present for every entry in the table of contents.

reedstrm:
ok move up a level - this per-page metadata exists in the raw epub for each page as proper HTML `<metadata>` tags. When we construct the single-file-html, this metadata is dumped into each page, so that it is available when we strip the baked single-file-html back to individual pages. It is present in the single-file-html so that it is available for modification by recipes.

This single-file-html is _not_ a format that we export or ship or deliver to anyone: it is format unique to our internal pipeline. 

So, those working with these files (textbook devs) are asking for it to be hidden, because it gets in the way of their examining the results of baking. For that case, I think having the code that builds single-file-html (from cnx-epub) add the `display:none` style attribute makes the most sense. Although I can see @philschatz argument that perhaps it should be a comment, I have a knee-jerk reaction against parsing comments as valid values for anyone except a human. In this case, it's not just knee-jerk: the recipe has to be able to easily update  (and construct, for composite-pages and sections) such data - comments are just-a-string.

tomjw64:
Closing in favor of https://github.com/Connexions/cnx-epub/pull/129

PR #418, opened 2018-06-05
Support top-titled tables

helenemccarron:
A quick and dirty modification to the table caption helper to support top titled tables. This is currently addressing the work-around in the content to support titles above the table. We will need to move the title from the table first header to the title tag then we will need to modify the mixin to something a little less hacky.


recipes/mixins/_number.scss
```diff
@@ -306,7 +317,18 @@
...
+      &.top-titled {
+        &::after {
+          class: "os-table-title";
+          content: string(TopTitleContent);
+          move-to: tableTopTitle;
+        }
+      }
+    }
+  }
+    &::outside {
```
tomjw64: indentation change?

recipes/mixins/_number.scss
```diff
@@ -173,23 +174,34 @@
   [data-type="chapter"] {
     :not(#{$elementName}) > #{$elementName}:not(.unnumbered) {
       @include utils_title($labeling, bCaption);
-      @include setCaptionsHelper($captionType, $defaultContainer, $hasCaption, $hasTitle);
+      @if $captionType == $CAPTION_TABLE {
+        @if ($hasTopTitle) {
+          &.top-titled {
```
philschatz: If the table has a top-title is the class name necessary or could the recipe just inject the title at the top? Maybe I am not understanding the problem.

Also, should this class name be prefixed with `os-` since it is being added by the recipe?

helenemccarron: The class `top-titled`is currently added in the cnxml. This issue lists all the possible tables that need to be supported: https://github.com/Connexions/oer.exports/issues/3399.

It is currently impossible to separate the title from the caption in the Docbook pipeline so the title was hard coded in the first table header row and styled in the PDF to resemble a "fake title". This PR supports the current markup and moves the title outside of the table to a more semantic and less hacky title + table based on the table's class determined in the content. 

When Press is working, we will move the title from the table header to the title tag as it should be. We will then need modify this mixin to behave properly: display the title above the table rather than within the caption based on the class of the table (top-titled or not).

philschatz: OK, in that case can you create a Card to denote the future change we will need to do? Otherwise, I think it would be easily forgotten

helenemccarron: Yes, I can create an issue here, or would you rather we create a card on the book's board?

philschatz: I think that the books board might be more visible to @openstaxalina @brittweinstein et al.

helenemccarron: Done! https://trello.com/c/0WMutV1q/639-fully-support-top-titled-tables

PR #419, opened 2018-06-11
Baked PDF styles

helenemccarron:
This PR contains only the styles files for Baked PDF. This is a spinoff of #384 without modifications to the styleguide or other scripts.


brittany-johnson:
After going through the remaining PRs in `tasty-tanuki` I realized that I forgot about the `styles` PR.  Since you @helenemccarron created it and cannot merge it into `tasty-tanuki`, I added both of our labels so that once you do the final review I can be notified to merge it.

PR #421, opened 2018-06-15
Figure Styling

damondemar:
Styled figures to match styling of Intro Business pdf


styles/framework/mixins/_captions.scss
```diff
@@ -0,0 +1,20 @@
...
+@mixin caption-style() {
+  font-family: font-map(serif);
+  font-size: size-scale(-2);
+  caption-side: bottom;
+  color: color-map(caption);
+  text-align: left;
+  page-break-before: avoid;
+  line-height: 1.6em;
+  display: inline;
+}
```
tomjw64: Seems like a good amount of duplication. Any chance that you could separate these into more concise mixins or make use of cascading to achieve the common features?

styles/themes/theme3/style/figures.scss
```diff
@@ -3,51 +3,82 @@
...
-  float: column-top; // size the surrounding container to the size of the image container to allow the
-  // left justified alignment of the Figure numbering and associated caption
-  // default width is (hero) full size @ 625px. Specifying no class or width defwidth: 100%;aults
-  // to full size.
-  // 3/4 size art
-  //.scale-padding(3, margin-bottom);
-
-  &.medium {
-    //width: @figure-width-medium;
+  float: column-top;
```
tomjw64: Are there any alternatives to avoid using float?

styles/themes/theme3/style/figures.scss
```diff
@@ -3,51 +3,82 @@
...
+    padding-bottom: size-scale(-4);
   }
 
-  > .body {
-    //#content>#figure>#body>.style(any; any);
-    //.scale-padding(-4, padding-bottom);
+  > .os-caption-container {
+    padding-left: 0px;
+    padding-bottom: size-scale(-1);
+
```
tomjw64: Here might be a nice place to put the cascading I was referring to above.

styles/themes/theme3/style/figures.scss
```diff
@@ -3,51 +3,82 @@
...
+      margin-left: size-scale(-5);
+    }
   }
+}
+
+// Figure Structure 2
+section > figure {
+  display: block;
+  margin: auto;
+  max-width: 6.5in;
```
tomjw64: Use constant PAGE_CONTENT_WIDTH()

styles/themes/theme3/style/figures.scss
```diff
@@ -3,51 +3,82 @@
...
+
+    .os-caption {
+      @include caption-style();
+      margin-left: size-scale(-5);
+    }
   }
+}
+
+// Figure Structure 2
+section > figure {
```
tomjw64: Do both of these figure structures actually occur? Both `section > figure` and `os-figure` match different things? There's a good chance only one is used in baked markup, and I have a hunch that even if both of those selectors match, they will be on the same elements.

styles/themes/theme3/style/figures.scss
```diff
@@ -3,51 +3,82 @@
...
+    margin-bottom: auto;
+  }
+
+  > span[data-type="media"] img {
+    padding-bottom: size-scale(-4);
+  }
+  
+  > div[data-type="title"] {
+    @include caption-number-style(); 
+    padding-top: size-scale(-6);
```
tomjw64: For very negative values, it might also be worth using a more fine alignment scale.

styles/framework/mixins/_captions.scss
```diff
@@ -0,0 +1,26 @@
...
+
+@mixin x-center-image() {
```
tomjw64: We've moved away from prefacing mixin names with 'x-'.

styles/framework/mixins/_captions.scss
helenemccarron: We should also consider that the mixins can be used for both figures and tables.

PR #422, opened 2018-06-15
Add temp additive alignment scale, toc alignment

tomjw64:
alignment scale located in typography (not sure if that's the right place), and it is based on the base font size, although it allows you to change the base.


styles/framework/mixins/_typography.scss
```diff
@@ -24,8 +24,12 @@
   @return settings(base-font-size);
 }
 
-@function size-scale($level) {
-  @return round(base-font-size() * pow(1.25, $level));
+@function size-scale($level, $alignment: 0px) {
+  @return $alignment + round(base-font-size() * pow(1.25, $level));
+}
+
+@function align-scale($level, $base: base-font-size()) {
```
helenemccarron: I wonder if we need some way to make explicit what value to expect. I could be hard to use without getting an idea of the scale (actual size).

brittweinstein:
@tomjw64 please put points on this or attach it to an issue.

PR #423, opened 2018-06-15
Recipe change for toc alignment

tomjw64:
Will give classes to top level TOC pages which are numbered, allowing us to style them differently.


brittweinstein:
This can slide out to next sprint.

openstaxalina:
@tomjw64 We currently have this in the current release (code freeze September 13).  Do you want to target this for the current release?  cc @helenemccarron @brittany-johnson

tomjw64:
This is ready to go and should be fine in r.r. if QA is not overloaded

PR #424, opened 2018-06-15
Try to quiet travis

tomjw64:


script/generate-guide
```diff
@@ -88,7 +88,7 @@ while IFS= read -r -d '' snippet_file_name; do # generic for loop
   # ./script/bake-file ${recipe_name} ${snippet_file_name} /dev/stdout | xmllint --html --format - > ${snippet_file_name}-baked.xhtml
   # use `source` so we do not have to re-initialize python/node/ruby
   do_progress_quiet "Baking ${raw_file_name}" \
-    source ./script/bake-file "${recipe_name}" "${raw_file_name}" "${raw_file_name}-temp1-${baked_or_prepared_suffix}.xhtml" --use-repeatable-ids
+    source ./script/bake-file "${recipe_name}" "${raw_file_name}" "${raw_file_name}-temp1-${baked_or_prepared_suffix}.xhtml" --use-repeatable- --quiet
```
philschatz: Is this a typo? It looks like `--use-repeatable-ids` was chopped off

tomjw64: Oops, not sure how that got chopped off. Will fix that. Generate-guide is what makes all the noise in travis. I didn't really know if adding an extra argument to every single script was necessary. However, it's true that this would quiet easybake for anyone using generate-guide. Would it make sense to set a variable at the beginning of the ci script and then tack it on to other calls if it exists?

tomjw64:
Generate-guide is what makes all the noise in travis. I didn't really know if adding an extra argument to every single script was necessary. However, it's true that this would quiet easybake for anyone using generate-guide. Would it make sense to set a variable at the beginning of the ci script and then tack it on to other calls if it exists?

tomjw64:
It's a little weird. Sometimes the warnings are useful; Sometimes they aren't.
Strings and buckets generate warnings when used while empty. The warnings can be really useful for figuring out why something is not working properly. (e.g. "The bucket is empty so my selector must not have matched any element"). However, there are use-cases for things that generate warnings.
The following is a pretty important one where we use the empty string as a pseudo-if-statement to tell us if a parent has a relevant child.
```css
parent {
  string-set: test "";
}
parent:deferred {
  class: string(test);
}

parent > child {
  string-set: test "child-exists"; /* string not set if child doesn't exist */
}
```

PR #433, opened 2018-06-27
SVG Icons, EOC, and Notes

brittany-johnson:
SVG Icons and EOC styling


styles/framework/mixins/_titles.scss
```diff
@@ -97,8 +97,8 @@
 @mixin titles-h2-style($font: font-map(serif), $title-color: color-map(h2-title), $noMargin: fals) {
   font-family: $font;
   color: $title-color;
-  font-size: size-scale(4);
-  line-height: line-height(4);
+  font-size: size-scale(3);
```
helenemccarron: Can you please delete the deprecated mixins for titles and verify they aren't used anywhere any more? If they are used, they should be replaced with `@include title(h3);` where `h3` is the proper header level

brittany-johnson: I removed the use of `size-scale` for `font-size` in the files that I committed. Since this PR is large, I'll save the complete removal of `size-scale` being used for `font-size` for another PR. @helenemccarron

Issue #445, opened 2018-07-06
Intro to Business: Preparing for Tomorrow's Workplace Skills


clteam18:
Globall for all Chapters

![image](https://user-images.githubusercontent.com/39794953/42384417-8d7f55ee-8157-11e8-91b0-a717edfb747e.png)

Heading "**Preparing for Tomorrow's Workplace Skills**" displays as "**Preparing for Tomorrow's Workplace**", the word "**Skills**" is cut in online view.

openstaxalina:
For Monday: I'm fairly certain this is a quick fix in the recipe.  We changed the title late, and may have only updated the PDF styling.

tomjw64:
![screenshot-1](https://user-images.githubusercontent.com/1285308/42641843-17a078b8-85bb-11e8-9652-84cca92199c0.png)

Stackblocks:
Displaying as specified with updated version of content / code.
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/9d46c888-0b76-427c-b58e-ed7ee1492b10)

PR #465, opened 2018-07-12
Add word 'Skills' to composite page title

tomjw64:
Word was missing. Simply added it onto the title in the config for the composite page. Link: Issue #445


tomjw64:
@helenemccarron Since the selector did not change, I believe a meaningful snippet change is not necessary. However, I went ahead an updated the title on the CNXML snippet to reflect the change in the feature title.

PR #468, opened 2018-07-13
removed empty caption container

Adrianltd:
I think this code is resposible for setting again a caption with no content, it has to be tested well.
I can no longer see empty caption tag in baked html, so hopefully that's enough.


Adrianltd:
@philschatz any idea why Travis Builds can't complete?

recipes/books/physics/_config.scss
```diff
@@ -63,7 +63,7 @@ $_equationTitleContent       : (os-number: counter(chapter) "." counter(equation
 /*==========================================================
   PHYSICS CONFIG
 ==========================================================*/
-$Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: false, hasTitle: true);
+//$Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: false, hasTitle: true);
```
tomjw64: Deletion is generally preferred over commenting.

Adrianltd: Yes, you're right

recipes/books/physics/_config.scss
```diff
@@ -63,7 +63,6 @@ $_equationTitleContent       : (os-number: counter(chapter) "." counter(equation
 /*==========================================================
   PHYSICS CONFIG
 ==========================================================*/
-$Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: false, hasTitle: true);
```
helenemccarron: While this works, what happens here is the config file will default to what is specified in the common file (`$Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: true, hasTitle: true);`) which enables both captions and titles. However, this title doesn't have a caption, only a title. The result is an empty span in the caption.


 I think we should investigate in `_number.scss` to ensure that empty captions are discarded properly when using a title but no caption.

recipes/mixins/_number.scss
```diff
@@ -184,6 +185,7 @@
       @include setCaptionsHelper($captionType, $defaultContainer, $hasCaption, $hasTitle);
     }
     @include trashUnnumberedCaption($elementName, $defaultContainer);
+    @include trashEmptyElement($defaultContainer);
```
helenemccarron: We can remove this, since we are adding it in the `_generator`

recipes/mixins/_number.scss
```diff
@@ -176,6 +176,7 @@
...
```
helenemccarron: We can remove this, since we are adding it in the `_generator`

recipes/books/_generator.scss
```diff
@@ -454,7 +454,10 @@
   @if $Config_TableSummariesAreCaptions {
     @include modify_tableSummaryToParent();
   }
+  @include trashEmptyElement (caption);
```
helenemccarron: We should use a more specific selector (captions in tables only).

books.txt
```diff
@@ -46,7 +46,7 @@ BOOK_CONFIGS=(
   "biology            biology           185cbf87-c72e-48f5-b51e-f14f21b5eabd    devb.cnx.org          mixins/"
   "apbiology          ap-biology        27cf60e4-bd82-418a-a7bb-a939c40f8f50    tea.cnx.org          mixins/,books/biology"
   "physics            physics           031da8d3-b525-429c-80cf-6c8ed997733a    devb.cnx.org          mixins/"
-  "apphysics          ap-physics        8d04a686-d5e8-4798-a27d-c608e4d0e187    devb.cnx.org          mixins/,books/physics/"
+  "apphysics          ap-physics        8d04a686-d5e8-4798-a27d-c608e4d0e187    katalyst01.cnx.org          mixins/,books/physics/"
```
tomjw64: Should this be a new entry instead of a modification?

helenemccarron: You mean a different PR? Yes, we can separate this into it's own PR. It might be most efficient to update all the Assignable Webview books to point at katalyst01.cnx.org along with this book.

PR #470, opened 2018-07-13
Fix Travis

philschatz:
Travis was complaining about needing to have sudo access in order to install [kcov](https://github.com/SimonKagstrom/kcov) . This fixes the install so that it does not require sudo access.

Refs #468


brittany-johnson:
@philschatz  How do you recommend testing this PR?  I can just approve it but I don't really understand it or know if it works😅

Adrianltd:
@brittany-johnson You can see Code Coverage Reports on PRs now, kcov is installed properly now ;)

Issue #454, opened 2018-07-06
Modify answer key's structure


helenemccarron:
In reference to https://github.com/Connexions/webview/pull/1625 and https://github.com/Connexions/webview/pull/1665, the recipes should modify the answer key to allow for better styling and flexibility

**See requirements below**

openstaxalina:
Requirements:
Find a more elegant way for the assessment number and content to appear on the same line.

We need to see what happens to images with this fix.  If the image falls onto the next line, we will need a separate fix for that.

PatBoj:
I've created a separate container for problem (`class="os-problem-container"`) and solution (`class="os-solution-container"`) content.

Before:
![image](https://user-images.githubusercontent.com/20907906/43519144-8d507dd2-958e-11e8-9b6b-cc1ab7a74dba.png)

After:
![image](https://user-images.githubusercontent.com/20907906/43519151-9043d9f8-958e-11e8-8131-a6ccacf84b7c.png)

wilkus27:
Verified on katalyst01, changes connected with recipes display properly:
![number_text](https://user-images.githubusercontent.com/40228854/43531257-b5a42e8e-95af-11e8-9447-c4e5784d6f20.JPG)

PR #478, opened 2018-07-19
Modified answer key's stucture

PatBoj:
#454 

Before:
image

After:
image


PatBoj:
Before:
image

After:
image

Now, we need to add 
```
div.os-problem-statment, div.os-solution-statment { 
  display: inline; 
}
```
and all should display properly.

recipes/mixins/_number.scss
```diff
@@ -46,23 +46,27 @@
   [data-type="chapter"] [data-type="page"] [data-type="exercise"]:not(.unnumbered) {
     @if $exerciseTitleContent {
       [data-type="problem"] {
+        div { class: os-problem-statment };
```
tomjw64: It might be safer to use a selector along the lines of ` > div:first-child` or `> div:first-of-type` in this case since we don't want to change the class of every div that happens to be in an exercise.

helenemccarron: My suggestion would be to wrap all the content _inside_ `[data-type="problem"]` and `[data-type="solution"]`. And yes, `::inside` is a hint ;) A good place to start looking would be in the mixin `modify_wrapExercises`

PatBoj:
Before:
image


After:
image

recipes/mixins/_modify.scss
```diff
@@ -564,8 +564,9 @@
         container: div;
         content: pending(bTitleLabel) content();
       }
-
     }
+    [data-type="problem"]::inside { class: "os-problem-container"; }
```
helenemccarron: can you add the same for solutions?

PatBoj: Done :) It's line below.

helenemccarron: Awesome, thank you! All that's left now is the feature flag so we don't have to test all books :D

Issue #505, opened 2018-07-26
Diagnose why AP Biology (w/ recipe) is not available on devB


brittweinstein:
@brittweinstein commented on [Tue Jul 24 2018](https://github.com/Connexions/content-managers/issues/128)

AP Biology has its own recipe; however, it does not show in the dropdown on DevB.  We should also test other servers, including katalyst01, staging, production, content##.![Screen Shot 2018-07-24 at 3.59.08 PM.png](https://images.zenhubusercontent.com/5a01faba8a75884b90866af7/31732e26-a08e-43af-895f-b3447856314f)

openstaxalina:
@Stackblocks as you are looking into this, could you verify whether AP Bio was tested with the correct recipe on katalyst01?

helenemccarron:
I think the symlink hasn't been created properly in OER.exports.

Stackblocks:
Still not seeing AP Bio w/ Recipe in the dropdown list, followed up with Ross and he pointed out that there is no ap-bio.css in the CSS folder of oer-exports, so it would not populate in the list.

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/03d7ca24-b1e5-454a-98ce-df56c8cb6eff)

---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/f24e806d-af72-4809-a27c-aa5079e63205)

---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/07f42880-38f6-4089-ad37-76bec44d3fd4)

Please add the appropriate .css file here and we'll check and see if the list updates correctly.

helenemccarron:
It would appear I didn't create the proper link when I created the symlink 😓 

https://github.com/Connexions/oer.exports/pull/3736/files#diff-ebe52cb872178530e0cabdba23357c3fR1

helenemccarron:
Issue moved to [Connexions/oer.exports #3748](https://github.com/Connexions/oer.exports/issues/3748) via [**ZenHub**](https://www.zenhub.com/)

PR #542, opened 2018-08-13
Create a symlink for AP Biology

helenemccarron:
Issue #505


helenemccarron:
Should be in OER.exports, closing

PR #549, opened 2018-08-14
Re-merge Issue481

brittany-johnson:
https://github.com/Connexions/cnx-recipes/pull/529/files


helenemccarron:
This is no longer needed,

PR #582, opened 2018-08-29
changed endpoint server for stats

Adrianltd:
Updated books.txt for the statistics books to point at katalyst instead of devb.


Adrianltd:
This might fix https://github.com/Connexions/webview/issues/1872

Adrianltd:
Updated books.txt for statistics to point at katalyst01 server

Issue #604, opened 2018-09-07
AP Physics - math not showing up in index


PatBoj:
@lwtchu commented on [Wed Sep 05 2018](https://github.com/Connexions/webview/issues/1916)

![Screen Shot 2018-09-05 at 3.37.44 PM.png](https://images.zenhubusercontent.com/5b57310fb26dd832dfe45b43/21b42c1c-c6ee-4f08-922d-abf539417f4a)

Should look like this: 
![Screen Shot 2018-09-05 at 3.41.23 PM.png](https://images.zenhubusercontent.com/5b57310fb26dd832dfe45b43/dc49f844-30d1-4ec3-a0fb-d78e815cbe87)

https://staging.cnx.org/contents/jQSmhtXo@17.1:Xm0hU3Qa@17.1/Index

---

@PiotrKozlowski commented on [Thu Sep 06 2018](https://github.com/Connexions/webview/issues/1916#issuecomment-419028389)

This looks like content issue but maybe a recipes issue? @PatBoj, @Adrianltd can you confirm?

---

@oscryan commented on [Fri Sep 07 2018](https://github.com/Connexions/webview/issues/1916#issuecomment-419262652)

A problem in College physics as well:
<img width="810" alt="screen shot 2018-09-06 at 5 32 23 pm" src="https://user-images.githubusercontent.com/1895432/45188631-d9463f00-b1fa-11e8-92e9-4c4fa944e44a.png">

openstaxalina:
The resolution we currently have will break University Physics (Polish and maybe also English). 

We will **not** resolve this issue for any titles for Assignable Webview for Half.  Instead, we will develop a more comprehensive solution in future sprints.

cc @PatBoj @helenemccarron

PatBoj:
Krzysiek already fixed this issue in easybake https://github.com/Connexions/cnx-easybake/pull/96.

PR #605, opened 2018-09-07
Fixed math in the index

PatBoj:
Issue #604 

### **Before:**
image

### **After:**
image

### **Before:**
image

### **After:**
image


recipes/mixins/_compose.scss
```diff
@@ -252,7 +252,7 @@
         // Create an index entry for the term by creating a new element and then moving it
         &::after {
           // i18n: use the @reference="..." version of a term when adding it to the Index (if the attribute is available). See #353
-          content: attr(#{"cmlnle|reference"}, content());
+          content: attr(#{"cmlnle|reference"}) content();
```
helenemccarron: @reedstrm I vaguely remember an old conversation about something similar, this makes sense to me but I don't remember why we needed to select the `attr`'s content. Can you take a look and let me know if you think this will not have any unforeseen effect?

reedstrm: OK, had to go read code to determine the diff: the two term one (on the left) looks up the value of an attr, but if that is None, it uses the current content as the default. The one on the right looks up an attribute and _preprends_ it to the existing content. If both the attr and the content exist, they are different. If only one of the two exists, they are the same.

recipes/output/ap-biology.css
```diff
@@ -1713,7 +1713,7 @@
   string-set: sectionHeaderString content(); }
 
 :pass(5) div[data-type="page"] span[data-type="term"]::after, :pass(5) div[data-type="composite-page"] span[data-type="term"]::after {
-  content: attr(cmlnle|reference, content());
+  content: attr(cmlnle|reference) content();
```
reedstrm: same, of course

recipes/mixins/_compose.scss
reedstrm: The reason this fixes the math problem, is that the "set default value" one (current code) passes the content through the string-value-extractor.  The presumption was that it was going to be used to set the value  of an attribute, which can't contain tags. Should probably be context sensitive, or something.

helenemccarron:
That makes sense. Just one thing: what do you mean by context sensitive, @reedstrm? 

@PatBoj should we test this on a Polish book to make sure we didn't break the `cmlnle|reference`?

PatBoj:
@helenemccarron that also was my concern, because we use `cmlnle|reference` very often in out books. I will check that and if everything will be ok I will merge this PR.

Adrianltd:
@PatBoj you have some conflicts unresolved

PatBoj:
@helenemccarron I've added a feature flag as a temporary solution. Could you please look at it and if it's ok then I will merge this PR.

aiwenar:
I had a look (since it's caused by code I wrote), and it looks like a bug in `cnx-easybake`. According to [CSS Values and Units Module Level 4](https://www.w3.org/TR/2018/WD-css-values-4-20180814/#attr-notation) (and level 3 as well)

> Note that the default value need not be of the type given, if the attr() expression is the entire property value.

and that's how I intended it to work (as in using current node set), but as @reedstrm noted `cnx-easybake` currently turns it into a string. 

I've started working on a solution, and expect to send it in one or two days more.

Also you don't have to block on UP translation, it's currently using a custom ruleset, so it won't be affected by any changes to recipes.

PR #607, opened 2018-09-10
removed unnumbered equations config

Adrianltd:
https://github.com/Connexions/cnx-recipes/issues/597 

After digging into it, I've figured out that the difference in numbering between PDF and webview is there not because of the wrong numbering mixin or wrong order of moving stuff and numbering it, but because we set some of the equations as `unnumbered`. 

This PR removes the part of code that is responsible for that. 
But I'm not sure if we want that. Someone had to set it unnumbered for some reason.
@helenemccarron @openstaxalina what do you think?

**After:**
image


Adrianltd:
@helenemccarron Now both of the books (ap physics and college physics) should have the same numbering as PDFs. I will fix the display in the webview now. Ping me if I can push this to testing.

PR #621, opened 2018-09-17
Tables styles

helenemccarron:


styles/framework/mixins/_tables.scss
```diff
@@ -0,0 +1,227 @@
...
+  @if $style == unstyled {
+    @include baseTableBody();
+  }
+  @if $style == text-heavy {
+    @include baseTableBody();
+  }
+}
+
+@mixin titleInCaption() {
+  background: green;
```
brittany-johnson: Is this a value that should be in the settings?

helenemccarron: no, it shouldn't be :D

PR #683, opened 2018-10-17
Reverse PR #657

helenemccarron:
This reverses PR #657 - see https://github.com/Connexions/cnx-easybake/issues/97 for proper way to fix this

### DIFF
--- a/recipes/mixins/_compose.scss
+++ b/recipes/mixins/_compose.scss
@@ -122,9 +122,10 @@
         class: "os-eoc os-#{$className}-container";
         data-type: "composite-page";
         data-uuid-key: "#{$keyName}";
-        @if feature-enabled(toc-comp-page-linking) {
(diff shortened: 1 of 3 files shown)


recipes/mixins/_compose.scss
```diff
@@ -122,9 +122,10 @@
         class: "os-eoc os-#{$className}-container";
         data-type: "composite-page";
         data-uuid-key: "#{$keyName}";
-        @if feature-enabled(toc-comp-page-linking) {
-          attr-id: "#{$className}" string(chap-id);
```
tomjw64: The string-set declaration above for chap-id should also be commented, as it was added in the reverted PR

helenemccarron: Yes, I left a comment to that effect, because I didn't like it either but wanted to preserve this code for the next iteration :D

Issue #638, opened 2018-09-20
Suppress Answer Key in Biology 2e, Concepts of Biology


oscryan:
# Scenario

Different answers are collated in staging pdf and on staging webview (staging pdf matches prd pdf).

# Reproducibility Details

- **Impact area**: staging.cnx.org
- **User**: all
- **URL / Location**: https://staging.cnx.org/contents/s8Hh0oOc@16.1:O6nBsAJu@16.1/Answer-Key
- **How many times were you able to reproduce the issue?**  1

## Steps to reproduce:

- Rebake/enqueue collection
- compare pdf to baked webview

## Screenshots

<img width="1379" alt="screen shot 2018-09-20 at 4 02 13 pm" src="https://user-images.githubusercontent.com/1895432/45846859-a7e76c00-bcee-11e8-809f-ddb195ddfd05.png">

# Environment:

- **Instance**: staging.cnx.org
- **OS**: 10.13.6
- **Browser**: firefox

# Impact

- **Severity**: 
- **Frequency of occurrence**:

openstaxalina:
The Answer Key should contain every-other-answer, beginning with the first question in each section.  The sections are:

Art Connection Questions
Review Questions
Critical Thinking Questions

For example, if Art Connections are questions 1-3, Review Questions are questions 4-20, and Critical Thinking Questions are 21-30, the Answer Key would include answers to questions 1, 3, 4, 6, 8, 10, 12, 14, 16, 18, 20, 21, 23, 25, 27, 29.

openstaxalina:
I am moving this issue into the next release.  We will discuss with the Katalyst team when we get the chance!

Bkblodget:
https://github.com/Connexions/oer.exports/issues/1637 I believe Kerwin's last comment in this issue captures the behavior we are seeing here

lwtchu:
Confirmed this affects Biology 2e also

helenemccarron:
I established that the root of this issue is the discrepancy between the PDF removing half of the answers ignoring odd and even while the recipe remove strictly only odd solutions. @openstaxalina was the decision to remove the answers from the content finalized or should we try to find a hack in recipes to make this work?

openstaxalina:
@helenemccarron we haven't been able to finalize a decision yet.  We will park at Books scrum tomorrow.

openstaxalina:
(If this requires coding, I understand that this will move to the next release.)

openstaxalina:
[...]

PR #697, opened 2018-11-05
issue-638

PatBoj:
Issue #638.

I've deleted all solutions from answer key.

### DIFF
--- a/recipes/books/biology/_config.scss
+++ b/recipes/books/biology/_config.scss
@@ -68,7 +68,7 @@ $Config_ChapterCompositePages: (
 );
 
 $Config_BookCompositePages: (
-  (className: "solutions",  clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_EOB, name: "Answer Key"),
+  (className: "solutions",  clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_TRASH, name: "Answer Key"),
   (className: "index",      clusterBy: $CLUSTER_NONE, specialPageType: $PAGE_INDEX, name: "Index"),
 );
 /*==========================================================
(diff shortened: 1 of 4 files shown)


recipes/books/biology/_config.scss
```diff
@@ -68,7 +68,7 @@ $Config_ChapterCompositePages: (
 );
 
 $Config_BookCompositePages: (
-  (className: "solutions",  clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_EOB, name: "Answer Key"),
+  (className: "solutions",  clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_TRASH, name: "Answer Key"),
```
helenemccarron: While this works, I think it would be cleaner to remove the composite page and use `$Config_PartType_Solution: (moveTo: $AREA_TRASH  );` to ensure all solutions

PatBoj: @helenemccarron Ok, done.

PatBoj:
@helenemccarron I changed the code as you suggested, but when I'm changing `moveTo: $AREA_TRASH_HALF, ` to `moveTo: $AREA_TRASH, ` Travis build fails. I don't know what's wrong, in other recipes this works.

recipes/books/biology/_config.scss
```diff
@@ -155,6 +155,6 @@ $Config_PartType_Exercise: (moveTo: $AREA_EOC, resetAt: $RESET_CHAPTER, numberAt
 $Config_PartType_Example:  (moveTo: $AREA_TRASH, ); // Oddly, this does not trash the examples... But maybe it shoudl be something other than NONE?
 $Config_PartType_Chapter:  (outlineTitle: "Chapter Outline", );
 $Config_PartType_Equation: (titleContent: null, );
-$Config_PartType_Solution: (moveTo: $AREA_TRASH_HALF, );
+$Config_PartType_Solution: (moveTo: $AREA_TRASH, );
```
Adrianltd: How about removing all of those `,` inside brackets?

helenemccarron: Historically, we kept those trailing `,` because for some reason, Sass will complain if they are missing, even at the the end of a declaration.

PR #703, opened 2018-11-08
Examples in appendices numbering

Adrianltd:
Transformed numbering of Examples in Appendices into feature-flag.

Should fix https://github.com/Connexions/cnx-recipes/issues/690 when https://github.com/Connexions/cnx-recipes/pull/671 will be merged.


recipes/books/_generator.scss
```diff
@@ -187,10 +187,11 @@
   @include validate_enumOptional(map-get($Config_PartType_Example, moveTo), $AREA__PREFIX__);
   @if map-get($Config_PartType_Example, moveTo) == $AREA_NONE {
     $titleContent:  map-get($Config_PartType_Example, titleContent);
+    $apTitleContent: map-get($Config_PartType_Example, apTitleContent);
     $solutionTitleContent:  map-get($Config_PartType_Example, solutionTitleContent);
 
     @include count_countExamples(example);
-    @include number_numberExamples($titleContent, $solutionTitleContent);
+    @include number_numberExamples($titleContent, $solutionTitleContent, $apTitleContent);
```
helenemccarron: I think we should use a feature flag here, since we have books that are live on production, this would require thorough testing before it could make it to prod.

recipes/mixins/_number.scss
```diff
@@ -107,6 +107,17 @@
   }
 }
 
+@mixin number_numberExamples($exampleTitleContent, $solutionTitleContent, $appendixTitleContent) {
+  [data-type="chapter"] {
+    @include _numberExamples($exampleTitleContent, $solutionTitleContent);
+  }
+  @if $appendixTitleContent {
```
helenemccarron: Like mentioned before, a feature flag would be safer. In addition, the  `$apTitleContent` is being applied to all books so the `@if $appendixTitleContent` doesn't seem to be relevant. All books, even those without examples, we need to make sure the `number_numberExamples` applies only to books with examples.

recipes/books/_common-stuff.scss
```diff
@@ -43,6 +43,7 @@ $Config_addSolutionHeader: ();
 $Config_TargetLabels: (
   (selector: '.os-eoc [data-type="exercise"]',                                  labelText: "Exercise " counter(chapter) "." counter(exercise)),
   (selector: 'div[data-type="chapter"] [data-type="example"]',                  labelText: "Example " counter(chapter) "." counter(example)),
+  (selector: '.appendix [data-type="example"]',                                 labelText: "Example " counter(appendix, upper-alpha) counter(example)),
```
helenemccarron: you may want to add a feature flag here to make sure only selected books are impacted.

recipes/output/ap-physics.css
```diff
@@ -586,57 +586,57 @@
 :pass(3) [data-type="chapter"], :pass(3) .appendix {
   counter-reset: example; }
 
-:pass(3) div[data-type="chapter"] [data-type="example"] {
+:pass(3) div[data-type="chapter"] [data-type="example"], :pass(3) .appendix [data-type="example"] {
   counter-increment: example; }
 
-:pass(3) [data-type="example"]::before {
+:pass(3) else [data-type="example"]::before {
```
helenemccarron: I am not sure where this else is coming from but that definitely will break the recipe. I would make sure this is fixed before merging.

PR #707, opened 2018-11-13
Move to a Jenkinsfile to release the Python package on tag

pumazi:
This fixes the release statements by installing git (needed by versioneer) and cleaning up after the build that mounts the source.

This change has similar benefits and behaviors to https://github.com/Connexions/cnx-db/pull/169 & https://github.com/Connexions/cnx-archive/pull/612

Note, this isn't a beautiful solution, but it does get the job done. I'll likely move this process into the pipeline library to allow for better handling of the individual operations. It'll likely, believe it or not, be easier to read and think about as Groovy code.


pumazi:
Also, I have tested the statement against test.pypi.org. The operation does successfully release.

Issue #704, opened 2018-11-08
Numbering by Step #


openstaxalina:
@wilkus27 commented on [Mon Nov 05 2018](https://github.com/Connexions/webview/issues/2038)

In _Elementary Algebra_ in **How To** note and in section **Key Concepts** there are lists which should be numbered by _Step 1, Step 2 etc._

webview (e.g. https://katalyst01.cnx.org/contents/CImQfPDv@3.14:Tj87ETCR@4/3-1-Use-a-Problem-Solving-Strategy)

![step_how_to](https://user-images.githubusercontent.com/40228854/48010120-4738b500-e11d-11e8-8b36-dcdb39309c6e.JPG)

PDF:
![step_how_to_pdf](https://user-images.githubusercontent.com/40228854/48010170-5d467580-e11d-11e8-8792-a27b918dd1fe.JPG)

webview (e.g. https://katalyst01.cnx.org/contents/CImQfPDv@3.14:7BOr1QuT@3.14/Key-Concepts)

![step_key](https://user-images.githubusercontent.com/40228854/48010210-72230900-e11d-11e8-80f8-aad07db06021.JPG)

PDF:
![step_key_pdf](https://user-images.githubusercontent.com/40228854/48010235-7d763480-e11d-11e8-82d6-e678d41a6a7d.JPG)

---

@openstaxalina commented on [Mon Nov 05 2018](https://github.com/Connexions/webview/issues/2038#issuecomment-436087142)

Hopefully these have the `stepwise` class on them; we should check in sprint planning.

---

@lwtchu commented on [Wed Nov 07 2018](https://github.com/Connexions/webview/issues/2038#issuecomment-436792627)

@openstaxalina they do :)!
![screen shot 2018-11-07 at 3 57 22 pm](https://user-images.githubusercontent.com/11299125/48163637-da045b80-e2a5-11e8-9edf-c50638ec1e16.png)

---

@openstaxalina commented on [Wed Nov 07 2018](https://github.com/Connexions/webview/issues/2038#issuecomment-436806517)

Thanks @lwtchu !
In that case, @PatBoj could you use that class to add the word "Step" before each number?  I think this is what we decided to do with all lists with that class.

cc @helenemccarron

openstaxalina:
Adding labels since this occurs across the DevMath series.  
Also, we noted that Chemistry has stepwise lists in some instances.  @oscryan is verifying whether the class "stepwise" was also used in those cases.

cc @lwtchu

[...]

PR #718, opened 2018-11-16
Add "Step" before number

PatBoj:
Issue #704.
I've added "Step " before the number in the lists with `class="stepwise"`. It's not the complete solution, I still need to add some styling on the webview (already did https://github.com/Connexions/webview/pull/2072).

## **Before:**
image

## **After:**
image


recipes/mixins/_modify.scss
```diff
@@ -60,7 +60,29 @@
   }
 }
 
-
+/// Adds `Step ` before number in lists with `class=stepwise`
+/// @group modify
+@mixin modify_stepwiseLists () {
```
helenemccarron: I would make this a more utilitarian mixin by passing the class via a variable, and maybe rename the mixin to modifyLists or something similar. I would like to be able to use this for other things in the future, should we need to create similar functionality for other classes or lists.

recipes/mixins/_modify.scss
```diff
@@ -60,7 +60,29 @@
...
+    > li {
+      &::inside {
+        container: span;
+        class: "os-stepwise-content";
+      }
+      &::before {
+        container: span;
+        class: "os-stepwise-token";
+        counter-increment: stepwiseCounter;
+        content: "Step " counter(stepwiseCounter) ". ";
```
helenemccarron: I would create a variable for the counter and one for the ` content: "Step " counter(stepwiseCounter) ". ";` as well.

recipes/books/_feature.scss
```diff
@@ -31,5 +31,9 @@ $FEATURES: (
 
   add-os-table-class: enable-for("statistics", "ap-biology", "biology", "economics", "ap-physics", "physics", "intro-business", "business-ethics", "dev-math", "chemistry", "astronomy"),
 
+  numbering-continuously: enable-for("dev-math"),
+
+  stepwise-lists: enable-for("dev-math"),
```
helenemccarron: Instead of a feature-flag, I think we should have a config option (this will be used only in certain books)

recipes/books/_generator.scss
```diff
@@ -474,6 +474,9 @@
   @if feature-enabled(add-os-table-class) {
     @include modify_addTableClass();
   }
+  @if feature-enabled(stepwise-lists) {
+    @include modify_stepwiseLists();
```
helenemccarron: Instead of a feature-flag, I think we should have a config option (this will be used only in certain books) so we can treat it as a functionality feature per book rather than system wide.

recipes/books/dev-math/_config.scss
```diff
@@ -169,6 +169,10 @@ $Config_UnnumberedExercises: (
   (contextSelector: '[data-type="example"]', childSelector: '[data-type="exercise"]'),
 );
 
+$Config_List: (
```
helenemccarron: I think we might want to clarify the name of this mixin to something `customList` or something descriptive, I am afraid we will forget what this does in 6 months from now.

PR #725, opened 2018-11-21
quick fix to the headings in appendices

Adrianltd:
Latest fix to headings in examples in appendices didn't take into the consideration headings which weren't inside `p` tags, so their levels weren't assigned correctly. 
1st depth section had h3, 2nd depth section had h2.
Now it should be correct.


wilkus27:
**PDF:**

pdf

**webview**

before

befor

after

after

PR #732, opened 2018-12-03
Add --with-resources to script/fetch-html

karenc:
script/fetch-html only gets the single html with no resource files.
Sometimes we need to see the content with images.  This change adds an
option `--with-resources` to download all the files in the `resources/`
directory.


karenc:
@philschatz @reedstrm I've rebased this branch, ready for review now.  Thanks!

PR #750, opened 2018-12-14
 Change class 'className' to 'os-className' in custom lists

PatBoj:
As in the title. This small change is reply to the Piotr comment here https://github.com/Connexions/webview/issues/2127#issuecomment-447075571.

### DIFF
--- a/recipes/mixins/_modify.scss
+++ b/recipes/mixins/_modify.scss
@@ -77,6 +77,7 @@
 
       .#{$className} {
         counter-reset: #{$counterName};
+        class: os-#{$className};
         > li {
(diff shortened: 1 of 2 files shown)


helenemccarron:
Mhh, I am not sure this is a desirable change (merged to master as well, can we undo?), the prefix `os-` is used for class that are added in the recipes only, in this case, we are modifying it. I prefer to keep the class as is and enable the `has-stepwise` config on a per book basis, so we can control which books should support stepwise. Regarding the css, with the text in the content, do we need the flex box or can we do without? cc @PiotrKozlowski

PR #753, opened 2018-12-18
Section exercises numbering

PatBoj:
I've created numbered section exercises title using template `Section [#chapter].[#section] Exercises`. My fix affects every single book, so regression tests are needed.

## **In the end of section**
image

## **In the _Answer Key_**
image

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -350,12 +350,12 @@
   }
 }
 :pass(8) {
+  @include count_countChapters(chapter);
+  @include count_countSections(section);
   //Only move solutions after exercises/solutions are numbered
   @include reference_refSectionHeaderNodeAs(sectionHeaderNode);
   @each $page in $Config_SectionCompositePages {
-    $name: map-get($page, name);
     $moveSolutionsTo: map-get($page, moveSolutionsTo);
-    @include validate_type($name, string);
     @include validate_enumOptional($moveSolutionsTo, $AREA__PREFIX__);
      @if $moveSolutionsTo == $AREA_EOC {
       @include compose_createEOCSolutions($Config_ChapterCompositePages, $page, sectionHeaderNode);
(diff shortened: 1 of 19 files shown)


recipes/books/_generator.scss
```diff
@@ -428,6 +426,7 @@
 
   //After: composite page creation
   @include count_countChapters(chapter);
+  @include count_countSections(section);
```
Adrianltd: duplicate mixins for counting chapter and section (line 353 and 354)

recipes/books/_generator.scss
```diff
@@ -350,12 +350,12 @@
   }
 }
 :pass(8) {
+  @include count_countChapters(chapter);
```
helenemccarron: I would test this with the continuous numbering, is this going to break it? Also, we should make sure this applies only to books that need them, currently, this applies to all books with wide implications to numbering and would require full regression testing on all books to ensure numbering isn't affected.

recipes/books/_generator.scss
PatBoj: Yes, but they are in different passes, they need to be duplicated. Otherwise numbering won't work.

PatBoj:
cc @helenemccarron @Adrianltd I've answered for your concerns and change the code a little bit.

recipes/books/_generator.scss
```diff
@@ -350,12 +350,14 @@
   }
 }
 :pass(8) {
+  @if feature-enabled(eos-custom-titles) {
```
helenemccarron: I was thinking something along the lines of `@if $Config_SectionCompositePages != ()` which checks if we are using EOS or not, do you think that would work? The benefits would be we wouldn't have to manage a feature flag.

PatBoj:
@helenemccarron I believe I finally finished this issue :D. @Adrianltd could you take a look as well.

Adrianltd:
I understand that `@include count_countChapters(chapter);` together with `@include count_countSections(section);` are replicated in two different passes for some reason, but they would apply always anyway because of the same conditional. So they apply first in the beginning on pass 8 and then somewhere in pass 9. Is this behaviour required @helenemccarron?

helenemccarron:
based on @PatBoj, they are both required for his fix to work. I am not sure what happens when you remove one or the other, so i can't answer without testing both cases (with and without the duplicate).

PR #756, opened 2018-12-19
Fix wrong exercise link label

PatBoj:
I've tried to fix wrong exercise link labels by adding a new selector to the target label config, but there is a little problem. The counter is shifted by 1 and I don't know why. I tried to put `@include link_setTargetLabels($Config_TargetLabels); @include link_setLinkLabels();` at the beginning and at the end of the `:pass(9)`, but nothing changed.

## **Before:**
image

## **After:**
image

## **Should be:**
image


helenemccarron:
@PatBoj when does the numbering differ? Could there be a hidden exercise that is counted? Upon looking at your code, it seems that they are in the proper pass, I'm not sure why they are not working either. cc @tomjw64 in case he has an idea or vague recollection why this might be happening

PatBoj:
@helenemccarron I checked and it's not hidden or skipped exercise. I will be working on this issue to find the cause.

Adrianltd:
@tomjw64 did you push it to the remote repo somewhere, so QA (@wilkus27 maybe?) can test it?

tomjw64:
@Adrianltd Just pushed to this branch with a fix

Issue #766, opened 2019-01-07
Missing Answer Key section


wilkus27:
On katalyst01 there is no section **Answer Key** on webview. It should be after **Chapter 28** and before **Index**.

![answer](https://user-images.githubusercontent.com/40228854/50768751-de3afe00-1281-11e9-9e7b-9304c1d6c5d2.JPG)

This is how Answer Key looks in PDF:

![answer key_pdf](https://user-images.githubusercontent.com/40228854/50768806-20fcd600-1282-11e9-809b-3d922a6eec9e.JPG)

We should have headers with information from which chapter answers are. 

Inside of Answer Key there are solutions to exercises from three sections from EoC: 

![eoc](https://user-images.githubusercontent.com/40228854/50768959-a08aa500-1282-11e9-9a43-48eb67a4c0cc.JPG)

At this moment solutions are inside the sections in option _Show Solution_.
E.g. https://katalyst01.cnx.org/contents/FPtK1zmh@12.1:h6GJJHO6@12.1/Interactive-Link-Questions

![interactive](https://user-images.githubusercontent.com/40228854/50768999-d3cd3400-1282-11e9-8bb3-16028fb4f80e.JPG)

Exercises are numbered continuously through the chapter. For example, on screenshot above the last exercise has number **5.** , so first exercise in Review Questions has number **6.**

![review](https://user-images.githubusercontent.com/40228854/50769137-49d19b00-1283-11e9-9919-ab3312bead3e.JPG)

openstaxalina:
Actually, in this book, we need to **suppress the answer key** like we did for Biology 2e, Concepts of Biology, and Psychology.  Like those other books, the content contains answers that should not be in the answer key, but we will not be able to get the webview and PDF answer keys to match.  

I believe this is the final book for which we need to suppress the Answer Key.  

cc @wilkus27 @Adrianltd @helenemccarron

wilkus27:
Verified on katalyst01. 

There is no Answer Key section in EOB:

![answer_end](https://user-images.githubusercontent.com/40228854/51383758-9be3ad80-1b1a-11e9-887d-2a1ce1332308.JPG)

And there are no more solutions inside the sections from EOC:

![show_sol](https://user-images.githubusercontent.com/40228854/51383790-b027aa80-1b1a-11e9-8d81-10a8b9b5352b.JPG)

PR #775, opened 2019-01-10
added answer key and trashed half solutions

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/766
Created Answer Key and trashed half of the solutions.
Still there's a difference between the webview and PDF versions of Answer Key, because in PDF the numbers start odd and then are even.

answerkey


Adrianltd:
Removed Answer Key according to https://github.com/openstax/cnx-recipes/issues/766#issuecomment-451959638

PR #813, opened 2019-01-31
Implement normalize.css

brittany-johnson:
Originally this was the PR: 
https://github.com/openstax/cnx-recipes/pull/651

But it was created so long ago and I don't want to deal with merging it :) 

We can discuss what things we want to include and remove from the file in the future and should keep in mind resets for different outputs.


tomjw64:
I notice the case file is directly imported. This might present problems after a few browser releases since I don't believe we plan to maintain the module. Is there a chance we could use the npm library here https://www.npmjs.com/package/normalize-scss and include it in our builds so that we can really several benefits including easy updates?

brittany-johnson:
@helenemccarron @tomjw64 removed the .css file and added the dependency!

brittany-johnson:
@tomjw64 I'm not sure 😅On the NPM page it says that you can install it using any of these methods: 
- Download directly from the project page.
- Install with npm: npm install --save normalize-scss
- Install with Bower: bower install --save normalize.scss
- Install with Ruby Gem: gem install normalize-scss

I guess if it was actually implemented though, we would have a change in the .css files?

tomjw64:
Try what's mentioned [this section](https://www.npmjs.com/package/normalize-scss#how-to-use-it) > Approach 2 > import scss immediately

PR #821, opened 2019-02-04
Allow npm in scripts

tomjw64:
Scripts currently error whenever yarn is not installed even though npm should be able to handle the job just fine

### DIFF
--- a/script/update
+++ b/script/update
@@ -7,4 +7,6 @@ do_progress_quiet 'Reinstalling python packages' \
 
  # Use --force so that the packages are reinstalled
 do_progress_quiet "Reinstalling NodeJS package.json" \
-  yarn install --prefer-offline --force
+  rm ./../node_modules || die "No JS modules installed"
+  [[ $(command -v yarn) ]] && yarn install --prefer-offline || \
+    [[ $(command -v npm) ]] && npm install
--- a/script/update
+++ b/script/update
@@ -7,6 +7,6 @@ do_progress_quiet 'Reinstalling python packages' \
 
  # Use --force so that the packages are reinstalled
 do_progress_quiet "Reinstalling NodeJS package.json" \
-  rm ./../node_modules || die "No JS modules installed"
+  rm -r ./../node_modules || die "No JS modules installed"
   [[ $(command -v yarn) ]] && yarn install --prefer-offline || \
     [[ $(command -v npm) ]] && npm install
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -209,7 +209,9 @@ if [[ ! ${BOOTSTRAP_ALREADY_RAN} ]]; then
 
   # ensure yarn is installed (for installing JS packages)
   if [[ -f "./package.json" ]]; then
-    [[ $(command -v yarn) ]] || die "Install yarn by going to https://yarnpkg.com or adding it to Brewfile (macos)"
+    [[ $(command -v yarn) ]] || \
+      [[ $(command -v npm) ]] || \
+        die "yarn or npm must be installed"
   fi
 
   # Note: Skip rbenv for Travis because it already has ruby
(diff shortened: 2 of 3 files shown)


philschatz:
I think it works but I think yarn was voted on because:

1. it resolves merge conflicts in the lockfile (unlike npm)
1. does not require network access to install packages (`yarn --prefer-offline`)
1. had better version-pinning support for 3rd-party libraries (X and Y) that both imported different versions of Z

I am curious, what were your reasons for not using yarn?

And if we go this route, can you also submit a Pull Request to https://github.com/openstax/setup-scripts . Other repos have this same boilerplate code in them (`./script/bootstrap` and `./script/setup`)

tomjw64:
@philschatz 
My reasons for not using yarn is that npm and yarn pretty much have feature parity. I'm not necessarily suggesting that we should make it the preferred option, as I still left the checks for yarn before the checks for npm, but npm is the default package manager and it seemed strange not to support it. Additionally npm has fewer emojis than yarn (yarn's argument, not mine). To defend npm a bit:
1) While it may not be quite as fancy, npm can handle lockfile conflicts semi-automatically (i.e. not human editing the conflicted lockfile) [explained in the npm docs here](https://github.com/npm/cli/blob/latest/doc/files/npm-package-locks.md#resolving-lockfile-conflicts)
2) [Additional docs](https://docs.npmjs.com/misc/config#offline) indicating that npm does indeed support offline installation. I'm not quite sure why we are using --prefer-offline in the first place. Is speed the only reason?
3) I'm not sure I understand this one. When you say 3rd party libraries, do you mean those not available via the cli installation tools for npm and yarn, like a github repo?

philschatz:
> 1. While it may not be quite as fancy, npm can handle lockfile conflicts semi-automatically (i.e. not human editing the conflicted lockfile) [explained in the npm docs here](https://github.com/npm/cli/blob/latest/doc/files/npm-package-locks.md#resolving-lockfile-conflicts)

I did not realize they added it as well. Nice!

> 2. [Additional docs](https://docs.npmjs.com/misc/config#offline) indicating that npm does indeed support offline installation. I'm not quite sure why we are using --prefer-offline in the first place. Is speed the only reason?

I occasionally code without an internet connection or tethering to my phone so it seemed like a waste to require people be connected to the internet when they were only reinstalling a repo or switching between branches but already had the packages cached (in yarn's case they're stored in `~/.yarn/cache` or something)

> 3. I'm not sure I understand this one. When you say 3rd party libraries, do you mean those not available via the cli installation tools for npm and yarn, like a github repo?

If packages X and Y both depend on Z npm would install Z twice... once under `./node_modules/X/node_modules/Z` and again under `./node_modules/Y/node_modules/Z`. But maybe recent versions of npm fixed that as well?

tomjw64:
@philschatz 
ah, yes. I believe this has been added in more recent npm versions. See [npm dedupe](https://docs.npmjs.com/cli/dedupe) which I believe is run implicitly upon `npm install` (e.g. I have deduped packages in my cnx-recipes using npm)

tomjw64:
Interesting... The logs indicate that it is still using yarn, no? The problem is that node-sass@4.11.0 is being installed but our recipes are only compatible with node-sass@<4.5.3

script/update
```diff
@@ -7,4 +7,6 @@ do_progress_quiet 'Reinstalling python packages' \
 
  # Use --force so that the packages are reinstalled
 do_progress_quiet "Reinstalling NodeJS package.json" \
-  yarn install --prefer-offline --force
+  rm ./../node_modules || die "No JS modules installed"
```
brittany-johnson: We found out that we have to add -r for this to work!

brittweinstein:
@tomjw64 what should I do with this PR?

tomjw64:
This can be closed. Its very old and this CLI is being retired soon

Issue #632, opened 2018-09-17
Create Precalculus Recipe


Stackblocks:
## Books title:
- Precalculus
- Algebra and Trigonometry
- College Algebra

### Collections: 
Precalculus: https://katalyst01.cnx.org/contents/_VPq4foj@7.52:vEOnJry_@3/Introduction-to-Functions
Algebra and Trigonometry: https://katalyst01.cnx.org/contents/E6wQevFf@10.1:nU8Qkzwo@4/Introduction-to-Prerequisites
College Algebra: https://katalyst01.cnx.org/contents/mwjClAV_@7.1:nU8Qkzwo@4/Introduction-to-Prerequisites

### Checklist
- [x] Update/create book map
- [x] Update books.txt with new book’s UUID
- [x] Create all notes config entries
- [x] Create all EOC config entries
- [x] Create all EOB config entries
- [x] Create all book specific config entries
- [x] Create all styleguide snippets
- [x] Update books.txt RECIPE_DESC with the new book’s description
- [x] Update manifest.json with the new book entry
- [x] Request a symlink to be created in OER.exports (+ verify it’s working before release)
- [x] Finalize book map to match recipe (include any change that might have been decided while developing the recipe)

PR #836, opened 2019-02-11
Precalculus recipe 632

Adrianltd:
This is a branch for the Precalculus book recipe.
#632


recipes/books/_common-stuff.scss
```diff
@@ -96,3 +96,6 @@ $Config_EOBSolutionsChapterSectionTitle: "Chapter " string(ChapNum);
 $Config_Index_SymbolsSectionTitle: "Symbols";
 $Config_Index_SymbolRegexp: \^\[\^a-zA-Z\];
 $Config_Index_NotSymbolRegexp: \^\[a-zA-Z\];
+
+// Set this to true in the book config in order to test has-unnumbered-elements release-flag
+$Config_hasUnnumberedElements: false;
```
tomjw64: Does this mean that this config option should be deleted after a certain period of time?

tomjw64: But would it not be the case that the following are equivalent: `length($Config_UnnumberedElements) == 0` and `$Config_hasUnnumberedElements == false`?

Adrianltd: Well, it seems so. Would it be better if we get rid of the `Config_hasUnnumberedElements` and use `length($Config_UnnumberedElements` instead in the conditionals?

tomjw64: I think it would be. Doing so would avoid unexpected behaviour in the long run

Adrianltd: @tomjw64 can you take a look at this now? surprisingly `!== 0` gives a syntax error, but `> 0` should be fine too. I've removed the change from the other books so right now it applies only to precalculus, other books use unnumbered exercises and unnumbered equations again.

tomjw64: Sass uses `!=` rather than `!==`.

tomjw64: This is getting better. I have another question: is it ever the case that `length($Config_UnnumberedElements) > 0` AND `length($Config_UnnumberedExercises) > 0`?

Adrianltd: Some books have `Config_UnnumberedExercises`, some have `Config_UnnumberedEquations`, some have both of them. Few books were set in their own config `$Config_UnnumberedExercises: (); // There are no unnumbered exercises` for some reason. And some books have none of that.

In `precalculus` we needed to set `unnumbered figures` so instead of creating 3rd config, which does the same stuff, I've unified all of that to `unnumbered elements` and right now it's applied only to `precalculus` recipe.

So theoretically `if length($Config_UnnumberedExercises) > 0` or `length($Config_UnnumberedEquations) > 0` or `length($Config_UnnumberedSplashFigures) > 0` then `length($Config_UnnumberedElements) > 0`. Practically it applies only to `precalculus` right now.

When it would be possible to test that change in all of the other books (I'm not sure if testing is that much required if output css doesn't change) we would replace those config options with `$Config_UnnumberedElements`.

tomjw64: If we don't expect the CSS output to change, it  is possible (and preferred) to implement such a change in every book right away instead differing how some books are handled for a period of time. However, given that this is work to be done on other books, I'm happy with us making a card for this instead of pushing back this PR any further.

PR #841, opened 2019-02-11
Compound composite answer key

helenemccarron:
Separate the answer key into a composite chapter with composite pages. 
No release flag used, enable $Config_hasCompositeChapter to apply.

List of books enabled:
- [x] ap-physics
- [x] american-government
- [x] ap-biology
- [x] chemistry
- [x] dev-math
- [x] history
- [x] microbiology
- [x] econ


MartaSztuka:
@helenemccarron I've baked and check Answer key in all of those books.
Books with Answer Key separeted and devided by chapters:
* ap-physics
* american-government
*  chemistry
* dev-math
* history
* microbiology
* economy

No Answer Key at all:
*ap-biology
* dastronomy
* university physics
* microbiology

So almost all of them are correct :)

recipes/mixins/_compose.scss
```diff
@@ -73,16 +73,45 @@
...
+            );
+          @include utils_title($titleContent, null, h1, "document-title");
+        }
+      }
+    } @else {
+      [data-type="composite-page"].os-eob.os-#{$className}-container {
+        $titleContent: (
+          os-text: $name
+          );
+        @include utils_title($titleContent, null, h1, "document-title");
```
tomjw64: Are the two above `else` blocks the same? Can we simplify this by using `@if $Config_hasCompositeChapter and $compoundComposite` with a single `else` block?

recipes/mixins/_compose.scss
```diff
@@ -748,26 +777,50 @@
...
+        class: "os-eob os-#{$solutionClassName}-container";
+        data-type: "composite-chapter";
+        data-uuid-key: ".#{$solutionClassName}"; /* TODO: Is this the correct key? We'll be stuck with it because it is used for generating the collated page uuid's */
+      }
+    } @else {
+      &::after {
+        content: pending(#{$solutionClassName}-TOCOMPOSITE);
+        class: "os-eob os-#{$solutionClassName}-container";
+        data-type: "composite-page";
+        data-uuid-key: ".#{$solutionClassName}"; /* TODO: Is this the correct key? We'll be stuck with it because it is used for generating the collated page uuid's */
```
tomjw64: Can we only wrap the different declarations in the `if/else` block to avoid repetition?

recipes/books/biology/_config.scss
```diff
@@ -156,5 +156,3 @@ $Config_PartType_Example:  (moveTo: $AREA_TRASH, ); // Oddly, this does not tras
 $Config_PartType_Chapter:  (outlineTitle: "Chapter Outline", );
 $Config_PartType_Equation: (titleContent: null, );
 $Config_PartType_Solution: (moveTo: $AREA_TRASH, );
-
-$Config_hasCompositeChapter: false;
```
tomjw64: Should we put `$Config_hasCompositeChapter: false;` in common_stuff as the default?

recipes/mixins/_compose.scss
```diff
@@ -73,16 +73,45 @@
...
-      @include utils_title($titleContent, null, h1, "document-title");
+    @if $Config_hasCompositeChapter {
+      @if $compoundComposite {
+        [data-type="composite-chapter"].os-eob.os-#{$className}-container {
+          $titleContent: (
+            os-text: $name
+          );
+          @include utils_title($titleContent, null, h1, "document-title");
+          [data-type="composite-page"] {
+            > span.os-text {
```
tomjw64: Looks like this title has already been generated. Is it safe to move just the `.os-text`? Additionally, it is a bit unclear why this movement is required in the first place.

recipes/mixins/_compose.scss
```diff
@@ -73,16 +73,45 @@
   @each $page in $compositePages {
     $name: map-get($page, name);
     $className: map-get($page, className);
-
+    $compoundComposite: map-get($page, compoundComposite);
```
tomjw64: This value should be validated on the following lines

recipes/books/biology/_config.scss
helenemccarron: already is :D that's why I removed it here

recipes/mixins/_compose.scss
helenemccarron: This is generating the `<span class="os-text">Chapter 1</span>` , the new composite-page inside the composite-chapter's title. It's not possible to create that title when creating the composite-page because you do not have access to the config value at that time. Originally, the subsection Chapter # was hard coded in `_compose.scss`, now it's being passed as a config entry.

recipes/mixins/_compose.scss
helenemccarron: yup, done!

recipes/mixins/_compose.scss
helenemccarron: done

recipes/mixins/_compose.scss
helenemccarron: done

recipes/books/biology/_config.scss
helenemccarron: There is a strange interaction happening with `@if $compoundComposite and $Config_hasCompositeChapter ` line 85 in `_compose.scss` though, for some reason, it will not work unless `$Config_hasCompositeChapter: false;` is placed at the end of the config file. I don't have time to investigate fully but I suppose it's an order of execution type of issue.

PR #845, opened 2019-02-13
EOC & Answer Key composite chapters

helenemccarron:
Convert compound composite chapters into composite chapters.
Applied to:

- u-physics
- astronomy
- dev-math
- microbiology

Please note that this builds on #841. It does modify the CSS for the books impacted in that PR as well as the books listed above. 

List of books impacted:
- ap-physics
-  american-government
-  ap-biology
-  chemistry
-  history
-  econ

This PR also created a release flag to add an extra level of support in the TOC that needs to be enable d in all books.


recipes/mixins/_toc.scss
```diff
@@ -2,10 +2,19 @@
...
+          > h3 > span,
+          > div > span {
+            move-to: title-spans;
+          }
+        } @else {
+          > h1 > span,
+          > h2 > span,
+          > div > span {
+            move-to: title-spans;
+          }
```
tomjw64: Can we shrink this `if/else` block to only contain only the differing selector: `> h3 > span`?

recipes/mixins/_toc.scss
```diff
@@ -153,6 +216,60 @@
...
+              container: ol;
+              move-to: cc-li;
+            }
+            &::after {
+              content: pending(cc-link) pending(cc-li);
+              container: li;
+              move-to: eoc-toc-pages;
+            }
+          }
+        }
```
tomjw64: This large block seems to be repeated. Why exactly is this the case?

recipes/mixins/_toc.scss
helenemccarron: I tried but I couldn't use the if/else in the middle of a selector

tomjw64: How about:
```scss
> a {
  > h1 > span,
  > h2 > span,
  > div > span {
    move-to: title-spans;
  }
  @if feature-enabled(composite-eoc) {
    > h3 > span {
      move-to: title-spans;
    }
  }
}

Comment (helenemccarron) [reply recipes/mixins/_toc.scss]: ah! Didn't think of that one! 👍

File: recipes/mixins/_toc.scss (hunk 2 above)
Comment (helenemccarron) [reply recipes/mixins/_toc.scss]: This one is inside ` > div[data-type="unit"] {`, the first one targets the TOC without Units.

File: recipes/mixins/_compose.scss [hunk 3]
```diff
@@ -73,16 +77,37 @@
...
-
+    $compoundComposite: map-get($page, compoundComposite);
     // Validate
     @include validate_type($name, string);
     @include validate_type($className, string);
-
-    [data-type="composite-page"].os-eob.os-#{$className}-container {
-      $titleContent: (
-        os-text: $name
+    //@include validate_typeOptional($compoundComposite, bool);
```
tomjw64: This line was added based on the old review, but it is commented. Are there any issues that occur when this line is not commented?

recipes/mixins/_compose.scss
```diff
@@ -73,16 +77,37 @@
...
+    @if $compoundComposite and $Config_hasCompositeChapter {
+      [data-type="composite-chapter"].os-eob.os-#{$className}-container {
+        $titleContent: (
+          os-text: $name
         );
-      @include utils_title($titleContent, null, h1, "document-title");
+        @include utils_title($titleContent, null, h1, "document-title");
+        [data-type="composite-page"] {
+          > span.os-text {
```
tomjw64: Still a bit interested in this and the following lines :stuck_out_tongue: What happens to this span's former parent? I assume it was wrapped initially in some sort of header?

recipes/mixins/_toc.scss
tomjw64: :astonished: I just realized something here. The eventual removal of this release flag will mean that the `> h3 > span` selector will remain separate from the other (and so my earlier suggestion may not have been the best). Any thoughts?

recipes/mixins/_toc.scss
tomjw64: Ah! I see! Could we mixinify this?

recipes/mixins/_compose.scss
helenemccarron: nope, an oversight, gone!

recipes/mixins/_toc.scss
helenemccarron: what I had before? 😉

recipes/mixins/_compose.scss
helenemccarron: No, it's created and moved on top of the container as is:
```
    @if $Config_hasCompositeChapter {
      &::after {
        container: span;
        class: "os-text";
        content: $Config_EOBSolutionsChapterSectionTitle;
        move-to: eobCompositeTitlesContent;
      }
      &::after {
        content: pending(eobCompositeTitlesContent) pending(note-#{$solutionClassName}-GETCHAPTER) pending(#{$solutionClassName}-GETCHAPTER);
        class: "os-eob os-#{$solutionClassName}-container";
        data-type: "composite-page";
        data-uuid-key: ".#{$solutionClassName}" string(ChapNum);
        move-to: #{$solutionClassName}-TOCOMPOSITECHAPTER;
      }```

Then I wrap a title around it. The reason is at the time of creation, I couldn't access `$name` which contains a chapter counter, so I just create the span as a place holder then access it at a later pass.

File: recipes/mixins/_toc.scss (hunk 1 above)
Comment (tomjw64) [reply recipes/mixins/_toc.scss]: Hmmm..... maybe so :)

Comment (tomjw64) [reply recipes/mixins/_toc.scss]: Or, perhaps as inspired by other parts of this pull request and earlier code: 
```scss
$nestedSpansSelector: ("> h1 > span", "> h2 > span", "> div > span");
@if feature-enabled(composite-eoc) {
  $nestedSpansSelector: append($nestedSpansSelector, "> h3 > span", comma);
}
#{$nestedSpansSelector} {
  move-to: title-spans;
}
```
But honestly this repetition would be so small, it's not worth the effort

PR #850, opened 2019-02-14
change a:link to a[href]

philschatz:
refs #843 and  openstax/unified#167

`a:link` _only_ applies to unvisited links but I think the intent is for all links to have this styling, right? See https://developer.mozilla.org/en-US/docs/Web/CSS/:link

If that is not the case then this can be closed.

### DIFF
--- a/styles/framework/base.scss
+++ b/styles/framework/base.scss
@@ -22,7 +22,7 @@ body {
   //vs line-height: line-height(0); vs $base-line-height???
 }
 
-a:link {
+a[href] {
   @include link-style(color-map(bodyLink));
 }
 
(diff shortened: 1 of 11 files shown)


helenemccarron:
In PDF, it would not make a difference :D I wonder if we want to keep the a:link for the web though, since there might be a difference there.

philschatz:
Right, we can add those when they occur (but they don't occur yet, right?):

- `a[href]` for all links
- `a[href]:link` for unvisited links
- `a[href]:visited` for visited links
- `a[href]:hover` for hover styling
- `a[href]:active` for currently-selected

helenemccarron:
Yes, that would be cleaner.

Issue #631, opened 2018-09-17
Create Calculus Recipe


Stackblocks:
### Collection: 
https://katalyst01.cnx.org/contents/i4nRcikn@2.89:H2TLb2-S@2/Introduction
https://katalyst01.cnx.org/contents/HTmjSAcf@2.61:rrzms6rP@2/Introduction
https://katalyst01.cnx.org/contents/oxzXkyFi@2.103:72YaCFgv@2/Introduction

### Checklist
- [x] Update/create book map
- [x] Update books.txt with new book’s UUID
- [x] Create all notes config entries
- [x] Create all EOC config entries
- [x] Create all EOB config entries
- [x] Create all book specific config entries
- [x] Create all styleguide snippets
- [x] Update books.txt RECIPE_DESC with the new book’s description
- [x] Update manifest.json with the new book entry
- [x] Request a symlink to be created in OER.exports (+ verify it’s working before release)
- [ ] Finalize book map to match recipe (include any change that might have been decided while developing the recipe)

PR #854, opened 2019-02-18
Calculus recipe 631

wilkus27:
Branch for Calculus recipe. 
https://github.com/openstax/cnx-recipes/issues/631


recipes/mixins/_modify.scss
```diff
@@ -7,6 +7,12 @@
   }
 }
 
+@mixin modify_trashElement($trashConfig) {
```
tomjw64: Still not a big fan of this mixin. Seems like one could just call the modify_trash mixin directly, since it's probably desired behaviour to not trash anything upon receiving null. What about something that takes a list instead?

recipes/books/_generator.scss
```diff
@@ -475,6 +475,11 @@
   }
 
   @include modify_retitleHeaders($Config_hasCompositeChapter); // Update the h2 h3 ... levels to make sense
+  @if release-flag-enabled(custom-lists) {
+    @include modify_lists($Config_CustomList);
+  }
+
+  @include modify_trashElement($Config_TrashSelectedElement);
```
tomjw64: Still think this is an unnecessarily vague mixin call. Shouldn't it take a list, or shouldn't the config option specify why the thing being trashed is being trashed?

wilkus27:
In 7473619 applied for calculus changes discussed in https://github.com/openstax/cnx-recipes/pull/1014

Issue #864, opened 2019-02-19
Missing space between numbers and text.


MartaSztuka:
In PDF we have 'Try It #10' and task:

![screenshot from 2019-02-19 09-42-16](https://user-images.githubusercontent.com/47241114/53002056-19595100-342c-11e9-8235-ee6550b1be90.png)

In webview Its 'Try' and additionally 'Exercise 10Evaluate the following' without space.

![screenshot from 2019-02-19 09-42-35](https://user-images.githubusercontent.com/47241114/53002221-78b76100-342c-11e9-8ff5-ddd1bc576234.png)

This issue occures in all book, needs to be fixed altogether with other 'try it' issues

helenemccarron:
@Adrianltd there's a way to number Try Its in the config file, see dev math or statistics.

MartaSztuka:
It's fixed

PR #871, opened 2019-02-19
changed numbering of try it notes

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/864

changed the way of numbering Try It notes

### DIFF
--- a/recipes/books/precalculus/_config.scss
+++ b/recipes/books/precalculus/_config.scss
@@ -69,7 +69,12 @@ $Config_ChapterCompositePages: (
   (className: "practice-test", clusterBy: $CLUSTER_NONE, name: "Practice Test"),
 );
 
-$Config_PartType_Exercise: (moveTo: $AREA_EOC, resetAt: $RESET_CHAPTER, numberAt: $NUMBER_BEFORE_MOVE, titleContent: $Config_exerciseTitleContent, );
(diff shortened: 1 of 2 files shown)


recipes/books/precalculus/_config.scss
```diff
@@ -69,7 +69,12 @@ $Config_ChapterCompositePages: (
   (className: "practice-test", clusterBy: $CLUSTER_NONE, name: "Practice Test"),
 );
 
-$Config_PartType_Exercise: (moveTo: $AREA_EOC, resetAt: $RESET_CHAPTER, numberAt: $NUMBER_BEFORE_MOVE, titleContent: $Config_exerciseTitleContent, );
+$_tryItTitleContent: (os-title-label: "Try It ", os-number:"#" counter(try),     os-divider: " ");
+$_tryItSolutionTitleContent: (os-number: counter(chapter) "." counter(try), os-divider: " ");
+
```
helenemccarron: are those extra spaces necessary?

Adrianltd: There's a mess with testing, so I cannot check it properly. I will merge this to the main `precalculus` branch so it's easier to test all of the issues without a need to rebake. If that's unnecessary I will remove those, but I think that there's a problem with math in solutions (I think the solutions starting with math start right after the number of solution)

PR #874, opened 2019-02-20
Add Dockerfile for cnx-recipes

karenc:
Create a docker container for cnx-recipes so installation is easier for
developers.

There's example usage at the top of the Dockerfile:

```
To create the image:
   docker build -t openstax/cnx-recipes:latest .

To start the container:
   docker run --mount type=bind,source=$(pwd),target=/repos/cnx-recipes -it openstax/cnx-recipes:latest /bin/bash
```


.dockerfiles/docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+set -e
+
+if [ ! -d /repos/cnx-recipes/node_modules ]
+then
+    . $HOME/.nvm/nvm.sh
+    cd /repos/cnx-recipes/
+    # We need to do this when the container starts up instead of at build time
+    # because ./script/setup adds files to /repos/cnx-recipes/ which are wiped
+    # when we start the container because we mount cnx-recipes
+    ./script/setup
```
pumazi: Hmm... seems like a small issue to be resolved at some later time.

pumazi:
```
$ docker build -t openstax/cnx-recipes:dev .
$ docker run -it --rm -v $(pwd):/repos/cnx-recipes openstax/cnx-recipes:dev /bin/bash
==> Setting up a Python virtualenv ... OK (16 sec)
==> Activating Python virtualenv ... OK
==> Installing JavaScript Packages ... OK (96 sec)
==> Installing python packages ... OK (89 sec)
==> Installing ./vendor/rhaptos.cnxmlutils/ ... OK
==> Checking out branch in ./vendor/rhaptos.cnxmlutils/ (see https://github.com/Connexions/rhaptos.cnxmlutils/pull/157) ... OK
Branch 'rng-fixes' set up to track remote branch 'rng-fixes' from 'origin'.
Switched to a new branch 'rng-fixes'
Already up to date.
Successfully set up
/repos/cnx-recipes
root@adc8997bdaa3:/repos/cnx-recipes#
```

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+WORKDIR /
+RUN apt-get -y install python-dev
+RUN apt-get -y install python-pip
+RUN pip install virtualenv
+RUN mkdir ~/.virtualenvs
+RUN pip install virtualenvwrapper
+
+FROM install-python as install-node
+COPY . /repos/cnx-recipes
+WORKDIR /repos/cnx-recipes/
```
pumazi: One small ask would be to either use `/src` or `/code` for consistency sake. We've got both of these in existing docker files in other repos. Seeing a third form added would be disappointing.

.dockerfiles/docker-entrypoint.sh
karenc: `node_modules` is created in the current directory when `./script/setup` is run.  So when the image is built, the directory is there but of course if we mount the current directory at `/repos/cnx-recipes`, then it won't be there...  Do you know how to fix it?

Dockerfile
karenc: Yes of course, I'll change it to `/src`. :+1:

karenc:
Ready for review again

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+RUN mkdir ~/.virtualenvs
+RUN pip install virtualenvwrapper
+
+FROM install-python as install-node
+COPY . /src
+WORKDIR /src
+RUN curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
+RUN /bin/bash -c '. $HOME/.nvm/nvm.sh && nvm install $(cat .node-version)'
+
+FROM install-node as install-yarn
```
philschatz: I was curious, what is the reason for all the `FROM ... as ...`? Is it just for documentation?

karenc: I didn't look deeply into this.  @kevinburleigh75 What is the reason? :D ...

philschatz: Oh, in that case it might be useful to replace some of these steps with https://github.com/openstax/ci-image/blob/master/Dockerfile#L9-L22

kevinburleigh75: That's how the docker tutorial told me to do it. :)  

There are more details here:
   https://docs.docker.com/v17.09/engine/userguide/eng-image/multistage-build/
but maybe it's overkill for our specific application.

I kinda like it because it tells me what the next block of code is about to do, and it seemed to make rebuilding things a lot faster.

philschatz: Thanks! It seems that Docker caches each step even if you do not add the stages.

I found out from the [slack message](https://openstax.slack.com/archives/CCMNYV29X/p1551106830069900) by  @TomWoodward about why [tutor-js/Dockerfile](https://github.com/openstax/tutor-js/blob/master/Dockerfile) has multiple stages (so different tools can pick different stages to use).

I added the stages back in to clarify: installing OS-dependencies, installing 3rd-party dependencies, and then copying the code over.

philschatz:
I think that @pumazi is out for the week. Pinging @TomWoodward in case he wants to review. Otherwise, I'm happy to merge it if @karenc does not have any objections

PR #890, opened 2019-02-27
Separate pdf specific styles

helenemccarron:
First pass at separating the platform specific styles. 

Some of the docs I am referring to while working on this: 
https://docs.google.com/document/d/12_-pRy3guvyi1P4GoLU7dFX8dsS5Iqph3_1mmv62qe4/edit#

Round 1 : toying with the idea of mixins that can be called based on the platform settings (doesn't solve `@import` limitations).

Todo: 
- [ ] remove styles from the framework and themes
- [ ] convert styles in reusable "modules" that are independent from context
- [ ] identify more PDF only styles and convert them
- [ ] rename/organize/refine the `plaftform` concept


helenemccarron:
@tomjw64, @philschatz, @brittany-johnson, @TomWoodward this is my very early stab at an idea I had for the framework. I am working on the rest but wanted to push this and get your feedback super early on this particular approach to styling components based on the platform. In essence, a mixin contains only what it needs based on the platform, is empty when it doesn't. I haven't added anything for Rex yet but the idea would be the same.

helenemccarron:
Also, I realize the diff is horrendous, I merged crusty.cronut into it, hopefully, it'll be a bit nicer when that gets merged, please look at the `styles/...` files

styles/platforms/modules/_paged.scss
```diff
@@ -0,0 +1,126 @@
...
+  color: color-map(folioChapterHeader);
+  font-weight: bold;
+}
+
+@mixin folio-chapter-header-reset {
+  content: CONTENT_BLANK();
+}
+
+
+@mixin paged() {
```
philschatz: I am a little worried about adding single-platform files like `_paged.scss` if it only applies to 1 platform (PDF).

styles/platforms/modules/_paged.scss
```diff
@@ -0,0 +1,126 @@
...
+  @if settings(platform) == "PDF" {
```
philschatz: Will these switch statements be sprinkled around the codebase or is there a way to have a platform file that provides these?

PR #901, opened 2019-03-07
Update node sass to latest

tomjw64:
This PR incorporates the changes from https://github.com/openstax/cnx-cssselect2/pull/8, which allows us to update node-sass/libsass to its latest version. The diff affects the css for every book still being maintained, but there should be no actual difference in the baked output. To confirm, I compare the output from the head of master to this branch's output. So far I have ensured that the checksum before/after matches on:
- [x] accounting
- [x] american-government
- [x] anatomy
- [ ] ap-biology
- [x] ap-physics
- [x] astronomy
- [x] biology
- [x] business-ethics
- [x] chemistry
- [x] econ
- [ ] entrepreneurship
- [x] history
- [x] intro-business 
- [x] microbiology
- [x] physics
- [x] prealgebra
- [ ] principles of management
- [x] psychology
- [x] sociology
- [x] statistics
- [x] u-physics


tomjw64:
Confirmed with Alan that we have tested a sufficient number of recipes

Issue #902, opened 2019-03-07
Example Numbers Do Not Match PDF


openstaxalina:
# What happened?
Example numbering restarts with each section in the PDF.  On webview is it currently continuous.

# Reproducibility Details

- **Does it also happen in an incognito window?**: Yes
- **Does this happen in one place, a few places, or everywhere?**: Examples everywhere

# Environment:

- **Book(s)**: _College Algebra_ (likely also _Precalculus_, _Algebra & Trigonometry_)
- **URL/Server**: https://katalyst01.cnx.org/contents/mwjClAV_@8.1:Blwxgq2s@10/1-2-Exponents-and-Scientific-Notation 
- **Location (page number, chapter/section)**: Multiple; example is from section 1.2
- **Suggested nearby searchable text**: "Using the Product Rule"
- **Browser**: Chrome

# Screenshots
![caexamplenumbering](https://user-images.githubusercontent.com/7863733/53993275-88a19700-40f4-11e9-9fa9-50f81025fd15.png)

MartaSztuka:
Baked&Tested in all 3 books
![exm](https://user-images.githubusercontent.com/47241114/54262360-42be5600-456e-11e9-82c2-e55a7f4604a1.jpeg)

helenemccarron:
This has been addressed with PR # #928 which was merged into the precalculus branch and will go out with it. Should we close this issue? cc @openstaxalina

openstaxalina:
@helenemccarron I put this back in the 3/26 release.

Here is what I understand from the above:

- The Precalculus recipe was created and merged.
- The Precalculus recipe will be available in the dropdown in the next release (3/26).
- However, the Precalculus recipe is not ready for regression testing yet.
- Therefore, we will not select the Precalculus recipe for our books on production until it's been through regression testing, resolved any bugs, and also once we've made relevant content updates.

If this is accurate, I recommend that @MartaSztuka verify this issue once we have code on staging early next week.  If the fix can be verified, this issue can be closed.

helenemccarron:
No, the Precalculus recipe has not been merged yet, only the PR specifically related to this issue, which was merged into the Precalculus branch. In order to test this issue, if she hasn't yet, @MartaSztuka should use the Precalculus branch to test.

MartaSztuka:
I've tested it today again, looks good.
[...]

PR #928, opened 2019-03-12
fixed numbering of examples inside sections

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/902

added a new counter for Examples, so it should count those in sections and reset on a page.
will work only for those that are not moved.

### DIFF
--- a/recipes/mixins/_count.scss
+++ b/recipes/mixins/_count.scss
@@ -130,7 +130,7 @@
   @include _count(
   $name,
   'section > [data-type="example"]',
-  'div[data-type="page"]'
+  'div[data-type="page"], .appendix'
   );
 }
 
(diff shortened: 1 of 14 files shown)


recipes/mixins/_count.scss
```diff
@@ -123,6 +123,17 @@
   }
 }
 
+/// A basic mixin for counting examples within a section using css counters
+/// @arg {Text} $name - Name of the css counter that will control example numbering
+/// @group count
+@mixin count_countSectionExamples($name) {
```
helenemccarron: This seems to affect all books, maybe the selector is too broad? Also, it looks like the approach you took on solving numbering of Try It per module might work here as well?

Adrianltd: Done

Adrianltd: And probably they also need to reset at `.appendix` as the `Try It` notes, so I'll add it also.

Issue #903, opened 2019-03-07
Try It Numbering Does Not Match PDF


openstaxalina:
# What happened?
Try It numbering restarts with each section in the PDF.  On webview is it currently continuous.

# Reproducibility Details

- **Does it also happen in an incognito window?**: Yes
- **Does this happen in one place, a few places, or everywhere?**: Try Its everywhere

# Environment:

- **Book(s)**: _College Algebra_ (likely also _Precalculus_, _Algebra & Trigonometry_)
- **URL/Server**: https://katalyst01.cnx.org/contents/mwjClAV_@8.1:Blwxgq2s@10/1-2-Exponents-and-Scientific-Notation 
- **Location (page number, chapter/section)**: Multiple; example is from section 1.2
- **Suggested nearby searchable text**: "Write each of the following products with a single base. Do not simplify further."
- **Browser**: Chrome

# Screenshots
![catryitnumbering](https://user-images.githubusercontent.com/7863733/53993474-1aa99f80-40f5-11e9-9c4e-fed892f5e13e.png)

MartaSztuka:
Same issue occurs in Algebra&Trigonometry, and Precalculus
![try](https://user-images.githubusercontent.com/47241114/54040355-6fb4e680-41c5-11e9-9899-06fb6036fd52.jpeg)

helenemccarron:
This has been addressed with PR # #934 which was merged into the Precalculus branch and will go out with it . Should we close this issue? cc @openstaxalina @MartaSztuka  . (edit: I am not sure about state of testing at this point)

openstaxalina:
@helenemccarron I put this back in the 3/26 release.

Here is what I understand from the above:

- The Precalculus recipe was created and merged.
- The Precalculus recipe will be available in the dropdown in the next release (3/26).
- However, the Precalculus recipe is not ready for regression testing yet.
- Therefore, we will not select the Precalculus recipe for our books on production until it's been through regression testing, resolved any bugs, and also once we've made relevant content updates.

If this is accurate, I recommend that @MartaSztuka verify this issue once we have code on staging early next week.  If the fix can be verified, this issue can be closed.

MartaSztuka:
I've tested all 3, it's good
https://katalyst01.cnx.org/contents/_VPq4foj@8.1:O5iKX5Vm@11/1-3-Rates-of-Change-and-Behavior-of-Graphs
[...]

PR #934, opened 2019-03-13
added reset at module to try it notes in precalc books

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/903

### DIFF
--- a/recipes/mixins/_count.scss
+++ b/recipes/mixins/_count.scss
@@ -176,7 +176,7 @@
   @include _count(
     $name,
     '[data-type="note"].#{$type}',
-    '[data-type="page"]'
+    '[data-type="page"], .appendix'
   );
 }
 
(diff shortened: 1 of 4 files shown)


recipes/mixins/_count.scss
```diff
@@ -168,6 +168,18 @@
...
 
+/// A basic mixin for counting notes within a module using css counters
+/// @arg {Text} $name - Name of the css counter that will control note numbering
+/// @arg {String} $type - A string of the selector of the note type you want to count
+/// @group count
+@mixin count_countSectionNote($name, $type) {
+  @include _count(
+    $name,
+    '[data-type="note"].#{$type}',
+    '[data-type="page"]'
```
helenemccarron: You may want to keep the `.appendix` selector in case those notes are used in the appendix (in which case they would reset per Appendix and not per module.

Adrianltd: @helenemccarron Done

Issue #880, opened 2019-02-26
Figure numbering is wrong


MartaSztuka:
# What happened? In PDF examples that refers to Figure 6 have caption: Figure 6a, and Figure 6b
In webview Figure numbers are always one ahead; there is Figure 7 with examples Figure 8.7a and Figure 8.7b, also they are linked to the picture right below.

- **Browser**: https://katalyst01.cnx.org/contents/mwjClAV_@8.1:mdOHcEnH@11/8-1-The-Ellipse#Figure_10_01_005

# Screenshots
PDF:
![screenshot from 2019-02-26 13-29-19](https://user-images.githubusercontent.com/47241114/53417410-d112d380-39d5-11e9-9847-09d7f90db1d0.png)

webview:

![screenshot from 2019-02-26 13-30-12](https://user-images.githubusercontent.com/47241114/53417529-16370580-39d6-11e9-8258-5dd4cd57588f.png)
![screenshot from 2019-02-26 13-30-23](https://user-images.githubusercontent.com/47241114/53417541-1afbb980-39d6-11e9-881f-cd360ade3324.png)

brittweinstein:
@theresamasciale please put points on this when you can

oscryan:
This is a difference in XYZ pdf numbering and current webview numbering. Katalyst is apparently already working on removing the chapter prefix (the first number in #.# of figure numbers) from webview for the XYZ books; the next step would be to determine what is numbering in webview that's not in the pdf that makes this figure 7 instead of 6. Let's not forget to check collating material.

theresamasciale:
Issue seems to be with the PDF -- The figure numbering restarts with module 1, so there are two "figure 1's" for each chapter (one in the intro and one in the regular chapter). 

PDF:

![Screen Shot 2019-03-04 at 3.56.16 PM.png](https://images.zenhubusercontent.com/5c51e1f3d9d1826f38fa04cc/0c96b377-a51b-44ac-9646-127caa8bd4e4)

![Screen Shot 2019-03-04 at 3.56.25 PM.png](https://images.zenhubusercontent.com/5c51e1f3d9d1826f38fa04cc/7b62a9e2-2b5e-4b24-8845-7d35ecc675cf)

openstaxalina:
@theresamasciale could you post the markup when you get a chance?  it should include the markup that contains the link to the figure as well as the markup for the figure itself.

theresamasciale:
Here is the markup for the initial splash image:

[...]

PR #935, opened 2019-03-13
fixed numering of the figures

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/880

Removed the splash image from the counter in the `Precalculus` trio.
Changed the way of adding `unnumbered` class to elements so now it's being done with once in the generator instead separete for every element.

Should not cause any regressions but affects books (changed the pass from 3 to 1 when the `unnumbered` class is being added):
`accounting`, `ap-physics`, `dev-math`, `physics`, `statistics`, `u-physics`


recipes/output/accounting.css
```diff
@@ -596,6 +596,9 @@
   class: "os-title";
   attr-data-type: "title"; }
 
+:pass(1) .your-turn [data-type="exercise"]:not(.unnumbered) {
```
helenemccarron: It appears this change moved this to a different pass. This also affects all books. Should this be release flagged and tested thoroughly to ensure this doesn't have any unintended consequences?

Adrianltd: I believe that adding an unnumbered class can go into any pass as long as it's earlier than the adding the numbers, so this does not require a deep testing. 
Adding `unnumbered` class to equations was previously in 1st pass, adding `unnumbered` class to exercises was in 3rd pass. In order to unify this together with unnumbering other elements, I've decided that 1st pass will be safer because of above.
But of course, you never expect the unexpected.
Precalc recipe branch has not yet merged the `release-flag` change hence I will add it as a `feature-flag` for now.

Adrianltd: @helenemccarron done. Now in order to test that in other books we only have to enable the feature-flag and set `$Config_hasUnnumberedElements` to `true` in the book's config. 
As mentioned before, the feature-flag in the config files was breaking the mixin so it didn't see variables so I did it with your suggestion.

Issue #888, opened 2019-02-27
Wrong captions


MartaSztuka:
In Learning Objectivies in whole book captions are different i PDF and webview.
https://katalyst01.cnx.org/contents/i4nRcikn@3.7:MM9PEdsz@4/5-1-Approximating-Areas

PDF:
![screenshot from 2019-02-27 12-45-54](https://user-images.githubusercontent.com/47241114/53488148-bd2aa880-3a8d-11e9-9a32-1ae4da582ac3.png)

webview:
![screenshot from 2019-02-27 12-47-02](https://user-images.githubusercontent.com/47241114/53488185-ddf2fe00-3a8d-11e9-881f-d732230c65fb.png)

Same issue occures in every chapter, every subsection at the begining.

wilkus27:
@helenemccarron I'm not sure in which way it should be resolved. The list in abstract is just a regular `<ul>` without any data-type or class: 

![calculus_lo](https://user-images.githubusercontent.com/40228854/53955503-7403d580-40d9-11e9-95e5-1ddfbb1a11d9.JPG)

I've checked and I didn't find any other book with numbered list in abstract. Probably Calculus is the only one textbook with such case.

I guess there are two ways to resolve it.

1. First one:
- maybe  `<ul>` lists should be changed to `<ol>` list with `type=1` and some class (e.g `class="abstractList"`) in content
- then we could use Patryks mixin:

![list_mixin](https://user-images.githubusercontent.com/40228854/53956137-18d2e280-40db-11e9-9b83-9c101f092146.JPG)

Probably it would have to be modified a bit, I'm not sure at this moment how much.

2. Second one:
- maybe new mixin should be created only for lists in abstract
- it could change the container to `<ol>` and add some class (if it is needed)
- what kind of output we want to get then? Similar to output from the mixin above:

![stepwise](https://user-images.githubusercontent.com/40228854/53956716-a7942f00-40dc-11e9-80fb-f09e8554f99b.JPG)

So `<ol>` with `os-class` and `<li>` with additional `<span>'s` inside?

openstaxalina:
**Decision:** Try to target the list by its place within the Learning Objectives.  If not, may need to add a class in the content.

MartaSztuka:
Issue is fixed
![Screenshot from 2019-03-29 14-37-08](https://user-images.githubusercontent.com/47241114/55236742-eecf9480-5230-11e9-9882-9197904e6937.png)
[...]

PR #940, opened 2019-03-14
Numbered list in abstract

wilkus27:
https://github.com/openstax/cnx-recipes/issues/888 Change bullet list in abstract to list with three level numbering: #chapter.#section.#item


wilkus27:
Now lists look like that:

abstract

Automatic list numbering will be removed by webview.

wilkus27:
@helenemccarron I tried to modify existing mixin related with custom lists. Those changes rebuild output in dev-math and chemistry. Should I apply release flag in that case? 

About calculus, at this moment those changes don't work there, because list in abstract should be `<ol>` not `<ul>`. Do we want to do this modification in content or should I create some mixin in recipe, which do this change?

MartaSztuka:
Issue is fixed
Screenshot from 2019-03-29 14-37-08
Screenshot from 2019-03-29 14-38-37
Screenshot from 2019-03-29 14-39-11

wilkus27:
@helenemccarron in a139ec5 I was trying to reduce repeating code as we discussed before. Please, give me your thoughts about it.

brittany-johnson:
@wilkus27 Hey! This looks really good! 
We suggest using one selector, instead of an `if else` for `className` and `dataTypeName` and to package this with a release flag

wilkus27:
@brittany-johnson To be sure, you mean to use one variable called `selector` inside the `$Config_CustomList`, so in calculus it will be:
```
$Config_CustomList: (
  (selector: '[data-type="abstract"] ul', beforeLabel: counter(chapter) "." counter(section) ".", afterLabel: ". "),
);
```
and in dev-math it will be:
```
$Config_CustomList: (
  (selector: '.stepwise', beforeLabel: "Step ", afterLabel: ". "),
);
```
that's right? If yes, with such solution there is a problem with that part of code:
```
#{$selector} {
        counter-reset: #{$counterName};
        class: os-#{$name};
        > li {
          &::inside {
            container: span;
            class: "os-#{$name}-content";
          }
          &::before {
            container: span;
            class: "os-#{$name}-token";
            counter-increment: #{$counterName};
            content: $beforeLabel counter(#{$counterName}) $afterLabel;
            move-to: listBucket;
          }
          &:deferred {
            content: pending(listBucket) content();
          }
        }
      }
```
The variable `$name` have to be one word string like `abstract` or `stepwise` because we are using it to create the names of classes of some containers. So maybe I should add this variable to config? 
Then it will look like that:
```
$Config_CustomList: (
  (selector: '[data-type="abstract"] ul', name: "abstract", beforeLabel: counter(chapter) "." counter(section) ".", afterLabel: ". "),
);
```

tomjw64:
@wilkus27 Really great thoughts. I think that we should go with the `selector` + `name` keys, so that:

1. All calls to this mixin look the same (i.e. less logical branching)
2. The arguments are easier to validate, since they would not be optional (right now use of `className` and `dataTypeName` must be included XOR to work as expected)
3. The name in the selector and class name we give to the baked content can differ if we so choose

Nice work reducing code duplication, by the way!

wilkus27:
@brittany-johnson @tomjw64 thank you for your kind words! :) 

In 7a73e94 I modified mixin and I added new keys to `$Config_CustomList` in **Chemistry** and **Dev-math** (and of course in **Calculus** ).

I'm wondering about release flag. If I want to enable in this PR all of those changes only for **Calculus** I have to create separate mixin, because earlier `mixin modify_lists` was included in generator file in `:pass(10)`.  But I had to move it to `:pass(9)`, because in **Calculus** in `beforeLabel` I'm using counters (`counter(chapter).counter(section)`) and they don't work in `:pass(10)`. So this change causes differences in output files for **Chemistry** and **Dev-math**.

So, should I do this or I can live this as-is? Or maybe I should create a release flag and enable it for all of three books (chemistry, dev-math,calculus)?

tomjw64:
@wilkus27 Since this change is for Calculus, go ahead and make a release flag for this and enable it for all the books except Chemistry and Dev-Math. This will isolate our change but prevent us from doing meaningless rollout in the future.

Just to tack on one more thing: Be sure you validate the type on `$selector`

wilkus27:
Thank you! :)

PR #948, opened 2019-03-19
chore: lint the styles

philschatz:
uses [stylelint](https://github.com/stylelint/stylelint) and [stylelint-no-px](https://github.com/meowtec/stylelint-no-px) to find all the cases where `px` is used.

Depends on #961 being merged first


tomjw64:
The question that we couldn't discuss over a call:
Why did we leave base font size, since rem assumes the user sets the base font size for themselves?

philschatz:
@tomjw64 I am not quite sure whether you are referring to the `base-font-size` field in the settings or the idea of including a base font size on the `<body>`.

- I think that in order for rems to work, they still need a `body { font-size: ##px; }` set. When a user zooms in/out the browser changes this value.
- I left the `base-font-size` variable in the settings because others did not weigh in on whether switching from `px` to `rem` was a good idea. Also, because of the previous bullet

helenemccarron:
I agree that the base-font-size (I think it should be applied to the body/book) should be left as px, for the reasons Phil mentioned.

PR #949, opened 2019-03-19
Generate platform files

helenemccarron:
Refs https://github.com/openstax/unified/issues/169

A basic script to generate two output files.

Changes the import statement for the platform files.

TODO: 
 - [ ] handle sass error to abort the script (currently stops and leaves the variable to whatever it was set when it errored
- [ ] move the platform array out of the build script

Few additional notes:
we could add and import a config for each platform (before the theme) and use the `settings` file:
  ```
    platform: PDF,
    scope: ".myclass",
```

and scope general mixins as such:

```
#{settings(scope)} {
  @include tables(settings(supported-tables));
}
```

### CODE REVIEW

File: script/build-styles [hunk 1]
```diff
@@ -3,7 +3,7 @@ cd "$(dirname "$0")/.." || exit 111
 source ./script/bootstrap || exit 111
 
 arg1=$1
-
+platforms=(PDF unified)
```
philschatz:
```suggestion
platforms=(pdf web)
```

- What is the benefit of using uppercase for some of the platform names but not others? it seems a bit confusing that the most of the path and filename is lowercase except for `PDF`
- Any thoughts on using the output rather than the project for the rex/web/unified platform?

script/build-styles
```diff
@@ -20,13 +20,24 @@ if [[ ${#STYLE_NAMES[@]} == 0 ]]; then
...
-  
+  read -r style_name _ <<< "${style_config}" 
+
   sass_file="${styles_root}/${style_name}/book.scss"
-  css_file="${output_dir}/${style_name}.css"
 
-  do_progress "Generating ${css_file}" \
-    "${sass_bin}" --source-map true --output-style nested "${sass_file}" "${css_file}"
-  
+  for platform in "${platforms[@]}"; do
```
philschatz: Can you also include this loop in the `./script/test-styles` script to make sure the CSS selectors are covered for both platforms?

script/build-styles
```diff
@@ -20,13 +20,24 @@ if [[ ${#STYLE_NAMES[@]} == 0 ]]; then
...
+
+    css_file="${output_dir}/${style_name}-${platform}.css"
+
+    sed -i 's/PFname/'${platform}'/g' ${sass_file}
```
philschatz: Could this be done using an external SASS function ([stackoverflow](https://stackoverflow.com/questions/15153613/provide-sass-variables-definitions-on-command-line))?

If not, then rather than commit the file into git, could it be a SCSS file that is ignored (in `.gitignore`) but has a `{filename}.example` file that is checked in? (Example: the README in [dotenv-safe](https://www.npmjs.com/package/dotenv-safe) )

That way there's documentation of what the file should contain but a script can just write out a new file without having to carefully replace what is currently in the file.


I'm partial to `environment.scss` in case there are more externally-defined values that would need to be set

script/build-styles
helenemccarron: I didn't put too much thought into the naming convention at this point, so it works for me!

script/build-styles
helenemccarron: If I'm reading the link you provided carefully, I'm not sure it would work? They are targeting variables, not `@import` statements, where the use of variables or functions aren't allowed, it seems to me, but I haven't tested it.

The issue I can see with the `environment.scss` is that the platform has to be imported at a specific time (see https://github.com/openstax/cnx-recipes/pull/950/files#diff-4daa36f27a476c08cdb66f82d672a2aaR5) but I can play with it and see.

PR #953, opened 2019-03-20
Build script w/ nodejs

tomjw64:
refs https://github.com/openstax/unified/issues/169

based on the build script changes in #949 by @helenemccarron. Instead of using the CLI for node-sass, this uses the nodejs script import and allow environment variables to be used to specify the platform. It also uses includePaths in place of using sed. Additionally, it injects a PLATFORM() function which returns the current platform. I used @debug to print it in each platform to ensure it's working.
This is mostly an exercise so we can see what options we have to mess with platform specifics.


script/build-styles
```diff
@@ -22,11 +21,15 @@ fi
...
-  css_file="${output_dir}/${style_name}.css"
 
-  do_progress "Generating ${css_file}" \
-    "${sass_bin}" --source-map true --output-style nested "${sass_file}" "${css_file}"
+  for platform in "${PLATFORMS[@]}"; do
+    css_file="${output_dir}/${style_name}-${platform}.css"
+
+    do_progress "Generating ${css_file}"
+    PLATFORM="${platform}" \
+      node js/build-style/build-style.js "${sass_file}" "${css_file}"
```
philschatz: It seems like if this command fails then the script will happily continue, is that correct? I usually prefix commands with `try` in order to fail (& print the exact command that was attempted)

tomjw64: Oops! I meant to include the commabd in the `do_progress` of the previous line! Good catch!

tomjw64: I can't seem to figure out how to get `do_progress` to work with the following command (when on multiple lines). Could you help me with that @philschatz since you wrote the function?

philschatz: This should work (or add a commandline arg to the node script):

```sh
    PLATFORM="${platform}" do_progress "Generating ${css_file}" \
      node js/build-style/build-style.js "${sass_file}" "${css_file}"
```

Issue #904, opened 2019-03-08
Figure numbering


MartaSztuka:
In PDF Figure numbering reset every subsection #.#. and starts with Figure 1.
In webview Figure numbering starts over every chapter.
https://katalyst01.cnx.org/contents/mwjClAV_@8.1:95eK2O0n@11/1-4-Polynomials

![domki](https://user-images.githubusercontent.com/47241114/54023626-b7237e80-4195-11e9-9508-724c019cd23f.jpeg)
![figi](https://user-images.githubusercontent.com/47241114/54283276-f854cd80-459d-11e9-9ca2-cc61638bc9ef.jpeg)

https://katalyst01.cnx.org/contents/_VPq4foj@8.1:IkN3_myo@9/1-6-Absolute-Value-Functions

https://katalyst01.cnx.org/contents/E6wQevFf@11.1:rqFRa9y4@12/10-1-Non-right-Triangles-Law-of-Sines

MartaSztuka:
It's good now
https://katalyst01.cnx.org/contents/mwjClAV_@8.2:Sc8taR03@9/1-5-Factoring-Polynomials
https://katalyst01.cnx.org/contents/_VPq4foj@8.1:IkN3_myo@9/1-6-Absolute-Value-Functions
https://katalyst01.cnx.org/contents/E6wQevFf@11.1:83kZ17SW@10/3-3-Rates-of-Change-and-Behavior-of-Graphs

![figr](https://user-images.githubusercontent.com/47241114/54824740-16ac7e80-4cac-11e9-80d0-ab5e5a13b25c.jpeg)

MartaSztuka:
Fixed in all 3 books
![aaaaaaaaaa](https://user-images.githubusercontent.com/47241114/55245889-622ed180-5244-11e9-82e0-937f38a4d148.jpeg)

uskka:
Numbering of figures is now reseted in every chapter, but not in the EOC. Figure numbering in EOC needs to be reseted too in order to match PDF numbering.

PR #957, opened 2019-03-21
changed numbering of figures within modules

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/904


recipes/books/precalculus/_config.scss
```diff
@@ -142,8 +143,19 @@ $Config_UnnumberedExercises: (
   (contextSelector: '[data-type="example"]', childSelector: '[data-type="exercise"]'),
 );
 
-$Config_TargetLabels: append($Config_TargetLabels,
-  (selector: '[data-type="page"] :not([data-type="example"]) > [data-type="exercise"]:not(.unnumbered)', labelText: "Exercise " counter(chapter) "." counter(exerciseMaybeInContent))
+$Config_TargetLabels: (
```
helenemccarron: Is it necessary to call the entire list here? I think this causes changes that might be undesirable. See https://github.com/openstax/cnx-recipes/pull/957/files#diff-02cc61e37c215d4a262f299a2d1d6bbbL1306, it looks like the example numbering is completely gone. Is that normal?

Adrianltd: SASS allows to append `one` element to the `list` using `append`. I needed to add few elements to that list so I've decided to copy-paste the whole entry from `common-stuff` and add them manually. I could create a new list and use `join` to merge 2 lists, but that wouldn't be a better solution IMO. But you're right, even though the config is the same as was in the `common-stuff` file, the output is different :/

PR #958, opened 2019-03-21
Fix off by one error

tomjw64:
parent content declaration occurred before the move-to for the child

### DIFF
--- a/recipes/books/_release.scss
+++ b/recipes/books/_release.scss
@@ -58,7 +58,8 @@ $RELEASE_FLAGS: (
     "intro-business",
     "business-ethics"),
 
-  EOCsection-links: enable-for(),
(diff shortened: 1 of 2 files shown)


recipes/books/_release.scss
```diff
@@ -58,7 +58,8 @@ $RELEASE_FLAGS: (
     "intro-business",
     "business-ethics"),
 
-  EOCsection-links: enable-for(),
+  EOCsection-links: enable-for(
```
helenemccarron: Shouldn't we enable this for at least one book to verify it's working?

tomjw64: Enabled for two books in https://github.com/openstax/cnx-recipes/pull/963. I put this PR as blocking for that one, so that we can test this for econ and biology.

Issue #889, opened 2019-02-27
Headline in Examples aren't bold


MartaSztuka:
In webwiev in Example section headline is written same style as text below. In PDF it's it's thicken.
https://katalyst01.cnx.org/contents/i4nRcikn@3.7:MM9PEdsz@4/5-1-Approximating-Areas

PDF:
![screenshot from 2019-02-27 13-06-01](https://user-images.githubusercontent.com/47241114/53491990-0b44a980-3a98-11e9-8766-6184c5dacb4f.png)
 webview:
![screenshot from 2019-02-27 13-06-10](https://user-images.githubusercontent.com/47241114/53492012-1566a800-3a98-11e9-8345-74c8dcfe3929.png)

Problem occures in whole book

openstaxalina:
If it's possible to target them -- likely in a title element-  headers should be bold as @MartaSztuka suggested.  cc @wilkus27

openstaxalina:
It seems like these can be targeted using `data-type="title":`
![Screen Shot 2019-03-25 at 1 19 53 PM](https://user-images.githubusercontent.com/7863733/54944098-b82c0e00-4f00-11e9-98e3-ac5c5dce6f51.png)

openstaxalina:
If markup is inconsistent, we need to standardize the markup for Example titles.

uskka:
Looks good. Tested on katalyst01.

<img width="958" alt="Screenshot 2019-04-03 at 12 23 11" src="https://user-images.githubusercontent.com/40228918/55471973-50b24480-560b-11e9-9719-329b2d74fed3.png">

PR #965, opened 2019-03-26
Titles in example

wilkus27:
Change titles in example note to `<h4>` headers. 
Related with issues: https://github.com/openstax/cnx-recipes/issues/889, https://github.com/openstax/webview/issues/2218 and https://github.com/openstax/webview/issues/2203.


wilkus27:
I created one flag, which change all titles inside the example note, although there is separate issue for each title. Mixin which I modified has impact on elements with `[data-type"example"]` but also with `[data-type="exercise"]`.  Because of that, this modification change also header **Hint** in Checkpoint note in Calculus, but this part will be trashed. 

So, please tell me your thoughts. Maybe better solution is to create separate flag for each case (`[data-type="problem"]`, `[data-type=solution]`, `[data-type=commentary]`) and only for `[data-type="example"]` not also for `[data-type="exercise"]`.

wilkus27:
This solution is too general. It has impact on captions of tables inside exercises:

table_in_exercise

table_title

Issue #914, opened 2019-03-11
Hint in Checkpoint note


wilkus27:
On katalyst01 (https://katalyst01.cnx.org/contents/i4nRcikn@3.7:fP4FUrwS@7/1-1-Review-of-Functions) there are hints in Checkpoint notes:

![checkpoint](https://user-images.githubusercontent.com/40228854/54134045-82265e80-4417-11e9-8f9a-a9a7c3895178.JPG)

In PDF there is no such part inside that note:

![checkpoint_pdf](https://user-images.githubusercontent.com/40228854/54134134-a71ad180-4417-11e9-9524-4a2cb39168ad.JPG)

I checked if it appears in PDF in Answer Key, but it's not:

![answer_pdf](https://user-images.githubusercontent.com/40228854/54134705-bc443000-4418-11e9-870f-1ae4632b0eee.JPG)

I also didn't find it in any other place in PDF.

wilkus27:
@openstaxalina should hints on webview be trashed? Or maybe they will be added to PDF?

openstaxalina:
@wilkus27 thanks for noticing this - hints should be trashed as we do not have plans to add these to the PDF.

MartaSztuka:
No Hints in Checkpoints anymore, all three books checked
![Screenshot from 2019-03-29 13-22-30](https://user-images.githubusercontent.com/47241114/55232446-e2920a00-5225-11e9-988e-e627ab5722e5.png)
![Screenshot from 2019-03-29 13-21-41](https://user-images.githubusercontent.com/47241114/55232450-e58cfa80-5225-11e9-9578-56f92edaefc2.png)
![Screenshot from 2019-03-29 13-21-52](https://user-images.githubusercontent.com/47241114/55232454-e756be00-5225-11e9-8f46-bdf5a6bf0960.png)

uskka:
Looks good. Tested on katalyst01.

PR #967, opened 2019-03-26
914 hint in checkpoint

wilkus27:
Related with https://github.com/openstax/cnx-recipes/issues/914.

I created new config to trash Hint element from Checkpoint note. It's just my proposition. 
I was also thinking about creating release flag. Please let me know, which solution is better.

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -476,7 +476,7 @@
 
   @include modify_retitleHeaders($Config_hasCompositeChapter); // Update the h2 h3 ... levels to make sense
 
-  @include modify_trashCustomElement($Config_TrashSelectedElement);
+  @include modify_trashElement($Config_TrashSelectedElement);
 }
 :pass(10) {
   //No Answer Key, make sure that solutions that are still present in the content are discarded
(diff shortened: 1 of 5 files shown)


recipes/books/_generator.scss
```diff
@@ -475,6 +475,8 @@
   }
 
   @include modify_retitleHeaders($Config_hasCompositeChapter); // Update the h2 h3 ... levels to make sense
+
+  @include modify_trashCustomElement($Config_TrashSelectedElement);
```
helenemccarron: I like the idea of a general trash config option, it appears to be missing. Can we name it `modify_trashElement` for concision?

recipes/books/calculus/_config.scss
```diff
@@ -150,6 +150,10 @@ $Config_UnnumberedExercises: (
     (contextSelector: '.checkpoint', childSelector: '[data-type="exercise"]'),
 );
 
+$Config_TrashSelectedElement: (
```
helenemccarron: Why is this a list? Could it be a simple mixin with a variable that would simply use `@include modify_trash($selector);`?

MartaSztuka:
No Hints in Checkpoints anymore.
Screenshot from 2019-03-29 13-22-30
Screenshot from 2019-03-29 13-21-41
Screenshot from 2019-03-29 13-21-52

PR #970, opened 2019-03-27
chore: Update the intro-business min-snippet

philschatz:
This updates the min-snippet file that was introduced in #843 and updates the list of selectors that do not exist in the book but do exist in the CSS. 

Should these selectors be removed from the CSS file since they do not exist in the book?


tomjw64:
The tables are certainly something in other books of the business series. I don't know what to say about that link to learning though... I think we should add this to allow tests to pass, but also we should remove most of the things in this `generic.xhtml` file once we drop theme3 and start working with the framework from scratch.

helenemccarron:
I agree, I am not even sure `link-to-learning` will exist as is in the initial theme, I would remove it from here (and from the empty theme in #960 as well)

philschatz:
I'll make a separate PR to remove the link-to-learning since this PR is just to update the min-snippet and there are several unused items in the `generic.xhtml` file. I'll rename the file in this PR to `unused.xhtml` in order to help clarify what is in it.

Issue #908, opened 2019-03-08
Subsection Number Section Exercises


MartaSztuka:
In PDF Section Exercises has Section numbers #.#, in webview it's missing.
https://katalyst01.cnx.org/contents/_VPq4foj@8.1:IkN3_myo@9/1-6-Absolute-Value-Functions

![exerc](https://user-images.githubusercontent.com/47241114/54034197-93246500-41b6-11e9-9b02-781b819f6e18.jpeg)

https://katalyst01.cnx.org/contents/E6wQevFf@11.1:RZUxkOcn@9/1-6-Rational-Expressions

https://katalyst01.cnx.org/contents/mwjClAV_@8.1:xJt0U6vR@11/2-3-Models-and-Applications

openstaxalina:
I know that this was a challenge; however, I think that we were able to do something similar for DevMath: https://cnx.org/contents/yqV9q0HH@17.2:gUzCHLwU@17/1-1-Introduction-to-Whole-Numbers.

MartaSztuka:
@openstaxalina issue is fixed, tested on katalyst01
![Screenshot from 2019-04-08 13-40-43](https://user-images.githubusercontent.com/47241114/55721591-09112b80-5a04-11e9-8dee-568391b24465.png)

openstaxalina:
@MartaSztuka I agree we can close.  Thank you!

Issue #905, opened 2019-03-08
Section Exercises numbering 


MartaSztuka:
In PDF at the end of every subsection in Section Exercises numbering starts with 1.
In webview Section Exercises numbering resets every chapter.
https://katalyst01.cnx.org/contents/mwjClAV_@8.1:95eK2O0n@11/1-4-Polynomials

![exercises](https://user-images.githubusercontent.com/47241114/54024172-3ebdbd00-4197-11e9-9f3a-f543e2e8c7be.jpeg)

https://katalyst01.cnx.org/contents/E6wQevFf@11.1:RZEC2R_o@6/9-4-Sum-to-Product-and-Product-to-Sum-Formulas
https://katalyst01.cnx.org/contents/_VPq4foj@8.1:bVz-NYId@12/2-2-Graphs-of-Linear-Functions

MartaSztuka:
@openstaxalina 
Issue is fixed, tested on katalyst01
May I close?
![Screenshot from 2019-04-08 09-46-21](https://user-images.githubusercontent.com/47241114/55707373-be7fb700-59e3-11e9-8c90-efaaef2bf8c3.png)

openstaxalina:
@MartaSztuka I agree it can be closed.  Thank you!

PR #972, opened 2019-03-27
fixed section exercises numbering, added RESET_SECTION option

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/905 and https://github.com/openstax/cnx-recipes/issues/908

Not a huge change, but should be reviewed and tested carefully. I'm not sure if that's the best approach to fix that issue, because a part of that code already existed and I didn't want to break anything that already worked fine in Dev Math.


helenemccarron:
I think you need to merge master into your branch in order for travis to work again. This said, does your change affect the dev-math recipe? Looking at your code, it looks ok but I am having a hard time seeing how it would affect (if at all) dev math.

PR #983, opened 2019-04-01
Change script/fetch-html to not use ssh

karenc:
- Change cnx-recipes to use python 3.6.8

  The baked-pdf CLI is in nebuchadnezzar which requires python >= 3.5.  To
  prepare for changing the scripts to use neb, we need to change the
  python version:
  
  - Change `.python-version` which is used for installing the correct
    python version in pyenv;
  
  - Change `Jenkinsfile` to use python 3 and create a universal wheel file
    so it's possible to install cnx-recipes in python 2 and python 3.
    Python 2 support is essential for cnx-deploy.
  
  - Change `script/bootstrap` to check the virtualenv python version and
    recreate the virtualenv if necessary.
  
  - Install cnx-easybake>=1.2.5 for python 3 compatibility fixes

- Change script/fetch-html to not use ssh

  script/fetch-html was ssh'ing into servers and running scripts on the
  server side to generate a single html for a book and downloading it.
  
  This is not ideal for the long term because it means that textbook devs
  need to have ssh access to servers for them to download a book.
  
  As part of the baked pdf CLI, we changed this to use "neb get" to
  download the book with collection.xml and index.cnxml, which is all
  public via https.  Then use "neb assemble" to create the single html.
  
  ```
  ./script/fetch-html biology
  ```
  
  will download the book in `data/biology/raw` and the single html is
  stored in `data/biology/collection.assembled.xhtml`.
  
  ```
  ./script/fetch-html --with-resources biology
  ```
  
  will download the book as above but also the resources (e.g. images).

- Change script/bake-book to use outputs from fetch-html

  The fetch-html script now generates the book html in the format
  `./data/{book_name}/collection.assembled.xhtml`.  Adjust the bake-book
  script to use this as the input file and change the output to
  `./data/{book_name}/collection.baked.xhtml`.

- Require the latest released version of Nebuchadnezzar

- Update the README to reflect changes to fetch-html


pumazi:
The fetch portion of this works for me. I tried it with both intro-business and psychology.

I'm having an issue baking though. The issues doesn't appear to be related to the changes in the PR. Perhaps rebasing will correct that, as there are newer changes on `master`.

pumazi:
I'm seeing a difference between `data/psychology/collection.baked.xhtml` and `data/psychology-baked.xhtml`. These should be the same no?

philschatz:
I added a couple commits to: 

- update the other scripts to use the `/data/{book_name}/collection.baked.xhtml` format.
- update the documentation to refer to the format.
- updated the `intro-business` min-snippet since it seems to have changed because the method of fetching the book changed (or just that the content has changed recently)
    - the HTML now pretty-prints so it will be easier to see what changed.

I can move the last bullet item into a separate Pull Request since it only tangentially pertains to changing `script/fetch-html`

karenc:
@philschatz Thanks so much for updating the scripts.  I was only testing `./script/fetch-html` and `./script/bake-book` so :sweat_smile:  I just updated some commit messages.

Issue #921, opened 2019-03-12
Wrong name of the links to the Theorem notes


wilkus27:
On katalyst01(e.g. https://katalyst01.cnx.org/contents/i4nRcikn@3.7:dKCfyV9u@5/2-2-The-Limit-of-a-Function) links to the Theorem notes take name of the other elements (like figure or table). But they lead to the proper places.

webview:

![theorem_link](https://user-images.githubusercontent.com/40228854/54196819-9ecba080-44c2-11e9-88ec-0929b8199de3.JPG)

PDF:

![theorem_link_pdf](https://user-images.githubusercontent.com/40228854/54196830-a7bc7200-44c2-11e9-9bdf-0b100d34f71e.JPG)

MartaSztuka:
Issue occurs in all 3 books
https://katalyst01.cnx.org/contents/HTmjSAcf@3.4:rn4DsBBU@7/1-3-The-Fundamental-Theorem-of-Calculus#fs-id1170571704350
https://katalyst01.cnx.org/contents/oxzXkyFi@3.7:1TskJjp2@5/2-4-The-Cross-Product#fs-id1163724130388

tomjw64:
@wilkus27 You will need to append to the targetlabel list (inside the calculus config file) with an entry for theorem notes. For example, something similar to:
```
$Config_TargetLabels: append($Config_TargetLabels,
  (selector: 'div[data-type="chapter"] [data-type="note"].theorem', labelText: "Theorem" counter(chapter} "." counter(theorem))
);
```
Once you have that line in your config, if it is not working properly, confirm if the theorem counter is initialized on the targetlabel pass, as well as if the selector value is indeed targeting your theorem notes. Happy hunting!

MartaSztuka:
Fixed in all 3 books. Tested on katalyst01
![Screenshot from 2019-05-21 13-30-21](https://user-images.githubusercontent.com/47241114/58092566-b6af5700-7bcc-11e9-8a95-a433e26c183f.png)

MartaSztuka:
Issue fixed, tested on katalyst01
![Screenshot from 2019-05-21 13-59-00](https://user-images.githubusercontent.com/47241114/58094302-a39e8600-7bd0-11e9-9833-ff21a35a798e.png)

PR #986, opened 2019-04-03
#921 link to theorem

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/921

Links to theorem notes should be named by titles of those theorems.


wilkus27:
@helenemccarron  At this moment it is not working properly. Links are named by titles from one theorem before. 
For example:
there are `Theorem 2.1 Two Important Limits`, `Theorem 2.2 Relating One-Sided and Two-Sided Limits`, `Theorem 2.3 Infinite Limits from Positive Integers` and all links to theorem 2.2 are now named by `Two Important Limits`, to theorem 2.3 by `Relating One-Sided and Two-Sided Limits` etc.

I couldn't find out why it is working in that way.

recipes/mixins/_link.scss
```diff
@@ -39,7 +39,18 @@
...
     $selector: map-get($target, selector);
+    $titleSelector: map-get($target, titleSelector);
     $labelText: map-get($target, labelText);
+
+    @if $titleSelector {
+      #{$selector} {
+        #{$titleSelector} {
+          string-set: title content();
+        }
+        $labelText: string(title);
```
tomjw64: This line will yield unexpected behaviour. It is not possible to set SCSS variable with cnx-easybake functions.

wilkus27:
@helenemccarron @tomjw64 in c1c140f I added all changes which we have done with Helene on her visit in our office. At this moment a few things are broken. We wanted to check what impact have some parts of code on links to theorem notes. Still we can't figure out why titles are passed by one.

wilkus27:
In f278cc0 I added ID of theorem notes to its subtitle. The code now doesn't look good, but it's working (locally). The links to theorems are labeled by corresponding  subtitles. Now they are not moved by one as previously.

image

With changes from da88d98, the link from above was named by the Theorem 2.1 subtitle, thas is **Two Important Limits**.

wilkus27:
@tomjw64 In 67d55e3 I've tried to use reference mixin for the subtitle string. At this moment all links to theorems are called by subtitle of the last theorem from the book: 

image

We discussed it with Helene and probably you mean something else saying about track the titles the same way we track the counters. Can you give a small example of infrastructure which you were talking about? Then we could be sure we are thinking about the same things.

tomjw64:
@wilkus27 You took a big step in the right direction with this PR. This is more like I had imagined as the solution for this. That being said, I'm not surprised that each one is being titled with the last theorem in the book, due to difference in the passes.
To solve this, we will need an extra step in generator. Start by throwing something like this into `pass(9)`:
```scss
@each $note in $Config_Notes {
  $className: map-get($note, className);
  @include reference_refNoteSubtitleStringAs(#{$className}SubtitleString, #{$className});
}
```

tomjw64:
@wilkus27 It looks like we are getting closer. I think you may be running into one of the less fortunate characteristics of cnx-easybake. I'm going to have to investigate this branch a little bit before I can point you in the right direction.

tomjw64:
### The problem
So to describe the issue we are running into here, we have to go into how easybake iterates through elements. This is what we can imagine we have. The numbers represent both iteration order and an identifier for the element.
```
note-0
└── title-1
note-2
└── title-3
```
When we call `target-string(attr(href), target-label)` we are asking for the value of the string `target-label` at the place which we pointed to with our href. So we can mark up what this string is at each point in the current code, keeping in mind we set the `target-label` string on each instance of a `title`
```
note-0  # (empty)
└── title-1  # title-1
note-2  # title-1
└── title-3  # title-3
```
So if we have a link pointing to `note-2`, it will read the value `title-1` from the target-label, which is why we are getting our off-by-one problem. Normally we to fix this sort of thing we could use something like `note:deferred` to get the note on a post-order iteration, but we cant use `:deferred` on an href. This means we can't use our usual tricks here, and we have to do some trickery.
### The Solution
Here is a picture of what we need to do:
Initial State:
```
note-0
└── title-1
note-2
└── title-3
```
Step 1 (temp elements):
```
title-1-copy
note-0
└── title-1
title-3-copy
note-2
└── title-3
```
Step 2 (do target label):
```
title-1-copy  # title-1
note-0  # title-1
└── title-1  # title-1
title-3-copy  # title-3
note-2  # title-3
└── title-3  # title-3
```
Step 3 (clean up to initial state):
```
note-0
└── title-1
note-2
└── title-3
```

The hard part about this is going to be making it extendable and reusable to other types of references, and not hardcoding it for a single type. Figuring out these situations in which we require extra passes is one of the more challenging aspects of easybake. Let me know if you have questions during the implementation of this.

wilkus27:
@tomjw64 This is my proposition for solve this problem: 1d6fe7c. I've tried to paste temporary title in many places, e.g as a first child, last child of the theorem note, outside of the note, I've included all of those changes in different passes in `generator` file and nothing worked. So I've created new attribute in container with theorem class, which in one pass in `generator` file it contains content of the subtitle, one pass later it is used for `target-label` and one more pass later it should be removed, but I couldn't find a way how to do it, so I just changed the value from content of subtitle to "true".

tomjw64:
@wilkus27 I think that some symbols may get messed up when they are in an element attribute, but this was a good way to go. Let's move forward assuming we aren't going to get any weird characters inside our subtitles.
I think we are almost done. I think that we can make this work inside the standard targetLabel infrastructure, rather than having to modify our setTargetLabel mixin.

wilkus27:
@tomjw64 Yes, it's working :) I pushed the change.

tomjw64:
Excellent! Approved.

Issue #923, opened 2019-03-12
Numbering in Review Exercises and Practice Test


MartaSztuka:
https://katalyst01.cnx.org/contents/_VPq4foj@8.1:FzKp--aR@8.1/Practice-Test
https://katalyst01.cnx.org/contents/E6wQevFf@11.1:XU_EPhae@11.1/Practice-Test
https://katalyst01.cnx.org/contents/mwjClAV_@8.1:3nhouCF9@8.1/Practice-Test
![pt](https://user-images.githubusercontent.com/47241114/54200984-d8a1a480-44cc-11e9-8aa8-d49567e27db5.jpeg)

Issue is in all books, also effects Answer Key.

MartaSztuka:
Figure Numbering doesn't match PDF in **EOC**
https://katalyst01.cnx.org/contents/_VPq4foj@8.4:Q0dD29cN@8.4/Review-Exercises
![asdfghjkl](https://user-images.githubusercontent.com/47241114/55404432-294e6f80-5558-11e9-825f-bfc57c1bc551.jpeg)

MartaSztuka:
Exercises numbering is fixed & tested in all 3 books.
@openstaxalina can I close the issue?
![sdgSRrhbeh](https://user-images.githubusercontent.com/47241114/55551043-1a8fc600-56da-11e9-80ed-3c13fd154655.jpeg)

openstaxalina:
@MartaSztuka I do not see the correct (restarted) numbering in the Review Questions or Practice Test on webview.  Here are screenshots from _Precalculus_:
<img width="1405" alt="Screen Shot 2019-04-04 at 11 58 19 AM" src="https://user-images.githubusercontent.com/7863733/55574032-009db580-56d1-11e9-809a-976169d5fd25.png">
<img width="1440" alt="Screen Shot 2019-04-04 at 11 54 30 AM" src="https://user-images.githubusercontent.com/7863733/55573922-c9c79f80-56d0-11e9-9663-6e1b72378175.png">

I tried testing with cleared browser cache, but I still see the numbering mismatch.

Would you mind rebaking and retesting this?

MartaSztuka:
@openstaxalina I'll do it now and let you know ;)

MartaSztuka:
@openstaxalina after rebaking numbers are restarting properly
[...]

PR #987, opened 2019-04-03
fixed EOC exercise numbering to be reseted at composite page

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/923

### DIFF
--- a/recipes/mixins/_count.scss
+++ b/recipes/mixins/_count.scss
@@ -61,7 +61,7 @@
   @include _count(
     $name,
     '.os-eos [data-type="exercise"]',
-    '#{$resetSelector}, [data-type="composite-page"], .appendix'
+    '#{$resetSelector}, .appendix'
   );
 }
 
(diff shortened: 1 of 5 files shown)


recipes/mixins/_count.scss
```diff
@@ -61,7 +61,7 @@
   @include _count(
     $name,
     '.os-eos [data-type="exercise"]',
-    '#{$resetSelector}, .appendix'
+    '#{$resetSelector}, [data-type="composite-page"], .appendix'
```
tomjw64: Instead of hard-coding `[data-type="composite-page"]` in, it might be more prudent to pass it as the `$resetSelector` argument instead

Adrianltd: Good idea. It improved the code so it does not affect `dev-math` books anymore.
There seems to be duplicate code in the output, but that's because two mixins (`count_countEOCExercises` and `count_countEOSExercises`) works the same way for these books, it should still reset the exercises the same way. I'll make sure that testers verify that. Ping me if it's good to merge, please.

tomjw64: @Adrianltd Looks good. Approved

Issue #993, opened 2019-04-08
Example links 


uskka:
Example links numbering looks like "1.1" it should be only "1". 
https://katalyst01.cnx.org/contents/_VPq4foj@8.4:WWmnx_o-@8.4/Key-Concepts

Webview

<img width="566" alt="Screenshot 2019-04-08 at 15 17 52" src="https://user-images.githubusercontent.com/40228918/55727536-7c219e80-5a12-11e9-9c52-2394e8860aa2.png">

PDF

<img width="1209" alt="Screenshot 2019-04-08 at 15 22 38" src="https://user-images.githubusercontent.com/40228918/55727537-7c219e80-5a12-11e9-8d28-e32daefbd691.png">

uskka:
Looks good. Tested on katalyst01.
<img width="1283" alt="Screenshot 2019-04-08 at 17 39 00" src="https://user-images.githubusercontent.com/40228918/55737368-58b41f00-5a25-11e9-8c25-52f0a1eb8fee.png">

uskka:
There should be additional numbering reset after each section in key concepts. Look at 1.2, examples numbering is not reseted but it's incremented. https://katalyst01.cnx.org/contents/_VPq4foj@8.4:WWmnx_o-@8.4/Key-Concepts

Webview

<img width="892" alt="Screenshot 2019-04-11 at 17 54 33" src="https://user-images.githubusercontent.com/40228918/55972584-12103000-5c84-11e9-8aec-4f73b6bc1ab2.png">

PDF

<img width="760" alt="Screenshot 2019-04-11 at 17 54 24" src="https://user-images.githubusercontent.com/40228918/55972583-12103000-5c84-11e9-96c2-315cd520576d.png">

uskka:
Now it looks good in all of the volumes. Tested on katalyst01.
<img width="1201" alt="Screenshot 2019-04-17 at 11 26 20" src="https://user-images.githubusercontent.com/40228918/56276881-a571b700-6103-11e9-8384-b3cee1ab2df6.png">

PR #994, opened 2019-04-08
changed the numbering of links to examples

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/993

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -459,7 +459,13 @@
     } @else { @error 'BUG: Invalid enum'; }
   }
 
-  @include count_countExamples(example);
+  $resetAt: map-get($Config_PartType_Example, resetAt);
+  @if $resetAt == moduleReset {
+    @include count_countSectionExamples(exampleInSection);
+  } @else {
+    @include count_countExamples(example);
+  }
+
   @include count_countEquations(equation);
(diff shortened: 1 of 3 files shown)


Adrianltd:
@tomjw64 According to the tests (https://github.com/openstax/cnx-recipes/issues/993#issuecomment-480885724) numbering is consistent. 
The Precalculus books have different kind of numbering than other books (everything is numbered with reset at module and the numbering does not take chapters into consideration so for example it's `5` instead of `1.5`).

Ping me of it's good to merge.

tomjw64:
@Adrianltd My concern was that the verification test was not complete in that it did not check the numbering on examples in any chapters/modules after Chapter 1 - module 1. Your changes do not imply that the examples now reset every module.

Adrianltd:
@tomjw64 you were right, after retesting it came out that the examples did not reset properly. It's not assosiated with this PR, but good catch. I'll fix it and let you know when it's ready.

Adrianltd:
@tomjw64 after few tests it turned out that it was associated with this PR, not the unproper reset of examples' numbering. They did reset at correct places - examples had proper numbering, but the numbering of the links didn't use the proper count-mixin and counter name.
Now it should be good, tested locally and looks fine. QA will verify tomorrow.

Thank you once again for your good eye!

PR #1014, opened 2019-04-17
Trash selected element

wilkus27:
Apply new config created in https://github.com/openstax/cnx-recipes/pull/967 for trashing selected element in **Anatomy** and **U-Physics**

Issue for Anatomy: https://github.com/openstax/cnx-recipes/issues/1031.

Issue for University Physics: https://github.com/openstax/cnx-recipes/issues/1032


wilkus27:
Well, in https://github.com/openstax/cnx-recipes/pull/967/files I've created `modify_trashElement` mixin and `TrashSelectedElement` config to remove in **Calculus** all commentary parts with hints from Checkpoint notes. Then Helene proposed to use this solution to old issues from other books in which some parts of book were removed. We found similar problems in **Anatomy** and **U-Physics**. 

Because changes from Calculus are not merged at this moment, I had to add those new mixin and config separately in this PR, but in the same places in files as it is now in `calculus-recipe-631` branch . So I had to decide if I want to change output in Calculus to keep logic from Anatomy or vice versa. 

About **U-Physics**, as you can see, in `_config` file there is a note:
```
Temporary fix while we wait for content to be fixed.
  We are waiting for these <strong> Check Your Understanding titles to either be removed or
  given a proper 'title' class/data-type.
  Currently, these aren't removed by other note modification operations, creating duplicate titles for these notes

:pass(0) {
  @include modify_trash('[data-type="note"].check-understanding p:first-child strong:first-child')
}
```
So there was an idea to resolve it in content. Helene just told me about this problem, so I tried to resolve it in this new way, but we didn't have time to discuss it. Maybe we should all talk about it on the call, also with @openstaxalina .

wilkus27:
Because after discussion we wanted to be able to add more then one selector, in 076e2ce I changed TrashSelectedElements config. Now we can add elements to trash in this way:
```
$Config_TrashSelectedElements: (
  (selector: '.preface [data-type="abstract"]'),
  (selector: '[data-type="page"] [data-type="title"]'),
  (selector: '...'),
);
```
But, because here there is just one key maybe it will be better to do this in this way, with simple list:
```
$Config_TrashSelectedElements: ('.preface [data-type="abstract"]', '[data-type="page"] [data-type="title"]', '...');
```
or create map with few selectors
```
$Config_TrashSelectedElements: (selector1: '.preface [data-type="abstract"]', selector2:  '[data-type="page"] [data-type="title"]', selector3: '...');
```

Also I'm wondering what should I do with rest of code? At this moment I removed mixin which I've created before and putted everything to `generator` file. Is it ok, or should I put this to another file? I mean this part:
```
@if $Config_TrashSelectedElements {
    @each $trash in $Config_TrashSelectedElements {
      $selector: map-get($trash, selector);
      @include modify_trash($selector);
    }
```

Please let me now what you are thinking about it @helenemccarron @tomjw64.

tomjw64:
@wilkus27 I think you should go with the simple list option, since the key seems to be meaningless :)

wilkus27:
@tomjw64 Done in d543631

PR #1015, opened 2019-04-17
Cluster notes in Answer Key by sections

helenemccarron:


Adrianltd:
@openstaxalina We've had a conversation with @helenemccarron about the titles of modules in the Answer Key. We think that using the full title of the section rather than using hardcoded `Section ` + number of the section in the `Try It` notes in `Answer Key` as in the PDF might be a better solution, because the code could be reused in other places this way. What do you think about that? Should we stick to the PDF or leave it as it is for now in webview?
webview-try-it
pdf-try-it

openstaxalina:
@Adrianltd thank you for looping me in!  As long as it *also* has the number (i.e. `1.1`), I think this looks great.

To confirm:

:white_check_mark: Desired: `1.1 Functions and Function Notation`
:x: Undesired: `Functions and Function Notation`

Does this make sense?

Adrianltd:
@openstaxalina Yes, so it it actually like that `desired` version right now, with the numbers. Thanks for the fast respond!

openstaxalina:
@Adrianltd thank you!  I thought so based on your screenshot, but wanted to confirm.  Thanks again!

brittany-johnson:
@Adrianltd According to the issue, I don't think we want to affect `u-physics`. Maybe we should apply a release flag?

Adrianltd:
@brittany-johnson release-flag added, it no longer affects `dev-math` nor `u-physics`

recipes/mixins/_compose.scss
```diff
@@ -624,7 +624,45 @@
...
+              move-to: note-#{$className}-header;
+            }
+            &::after {
+              class: "os-tryit-solution-area";
+              content: pending(note-#{$className}-header)  pending(note-#{$solutionClassName}-SECTIONS);
+              move-to: note-#{$solutionClassName}-GETCHAPTER;
+            }
+          }
+          @else {
+          [data-type="note"].#{$className} [data-type="solution"] {
```
tomjw64: The formatting and indentation in this addition make it difficult to read and thus difficult to review. Can this be adjusted?

recipes/mixins/_compose.scss
```diff
@@ -624,7 +624,45 @@
...
+              content: $noteTitle;
+              move-to: note-#{$className}-title;
+            }
+            &::before {
+              container: h3;
+              content: pending(note-#{$className}-title);
+              move-to: note-#{$className}-header;
+            }
+            &::after {
+              class: "os-tryit-solution-area";
```
tomjw64: this line implies that the only kind of moved note is a 'Try It', is that and will that always be the case?

recipes/mixins/_compose.scss
```diff
@@ -638,12 +676,14 @@
...
-          container: div;
-          class: "os-solution-area";
-          content: pending(note-#{$className}-header) pending(note-#{$className}-#{$solutionClassName}-GETSECTION);
-          move-to: note-#{$solutionClassName}-GETCHAPTER;
+          &::after {
+            container: div;
+            class: "os-solution-area";
+            content: pending(note-#{$className}-header) pending(note-#{$className}-#{$solutionClassName}-GETSECTION);
+            move-to: note-#{$solutionClassName}-GETCHAPTER;
+          }
```
tomjw64: It seems the only change to this block is that it was indented, but it does not seem to be indented correctly.

recipes/mixins/_compose.scss
Adrianltd: 1. Yes.
2. You're right, there might be some other cases in the future - changed that in the latest commit.

recipes/mixins/_compose.scss
Adrianltd: I think it's correct now.

recipes/mixins/_compose.scss
Adrianltd: @tomjw64 should be correct now

PR #1016, opened 2019-04-17
Styling framework with subspaces

tomjw64:
Here's a sample of the styling framework with design spaces, and as a bonus, a fairly complete suite of unit tests to go along with it.

WIP:
- [x] Better nesting on superspace creation
- [x] ~Transfer over the color-scheme logic~ (settings can do this now?)
- [x] Transfer over platform logic
- [x] Use the same importers on building the styles as were used for the unit tests
- [x] Repoint build scripts to the right places
- [ ] Repoint test scripts to the right places
- [ ] Make a sample book, template, and design directory
- [ ] Documentation of public API
- [ ] Undo temporarily commented-out code in `./script/test`
- [ ] Polyfill for property resets on components (optional)
- [x] Framework tutorial


philschatz:
1. I was looking for sass code coverage but did not find anything (I’ll try harder… maybe the Dart SASS project outputs coverage via an API?)
2. `sass-test` works by generating a CSS file and then parsing the CSS comments (see snippet below)
3. Nothing in the public API (`create_superspace`, `add_settings_functions`, and `use`) is SASS-specific per-se which indicates a good code-smell to me.

styles/build/build.js
```diff
@@ -0,0 +1,46 @@
+const fs = require('fs')
+
+const sass = require('node-sass');
```
brittany-johnson: Should we remove this semicolon since you're not using them anywhere else in the file?

styles/build/build.js
```diff
@@ -0,0 +1,46 @@
...
+const { String: SassString } = sass.types
+
+const platform = (process.env.PLATFORM || "NONE").toLowerCase()
+const platformIncludesPath = `${__dirname}/../platforms/${platform}`
+if (!fs.existsSync(platformIncludesPath)) {
+  throw new Error(`Specified platform '${platform}' does not have an associated directory.`)
+}
+
+const inputFile = process.argv[2]
+const outputFile = process.argv[3]
```
brittany-johnson: Is the platform argument passed after the output file argument? Also is it optional?

brittany-johnson: Answer: Pass an env var

styles/build/build.js
brittany-johnson: Answer: Thomas agrees

tomjw64:
>     1. I was looking for sass code coverage but did not find anything (I’ll try harder… maybe [sass/dart-sass](https://github.com/sass/dart-sass) outputs coverage via an API?)
> 
>     2. `sass-test` works by generating a CSS file and then parsing the CSS comments (see snippet below)
> 
>     3. Nothing in the public API (`create_superspace`, `add_settings_functions`, and `use`) is SASS-specific per-se which indicates a good code-smell to me.
> 
> 
> RE 3: I would like to try to use the handy `dump_*` functions to create a JSON file and then hack up an implementation in another language to test that theory. If that is true, then I really like how portable the files made using the framework are. If, in the future we wanted to switch to something else to generate the CSS files then it would be relatively simple to swap.
> Click me to see the **sass-true CSS file**

I am still looking for coverage options as well with not much luck :thinking: . I'm glad there are good code good smell :smiley: - would like to hear more about this portability of which you speak.

Issue #1019, opened 2019-04-18
Question numbering restarting each chapter after EOC nesting changes - no longer matching PDF, etc.


openstaxalina:
@yblock commented on [Thu Apr 18 2019](https://github.com/openstax/rhaptos.cnxmlutils/issues/177)

Book: Astronomy 
Server: Production 
CNX link: https://cnx.org/contents/LnN76Opl@20.2:55W6DXyy@20.2/Review-Questions

![Screen Shot 2019-04-18 at 8.32.23 AM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/f03c17ee-252e-4000-b05d-e63cfa1093ba)

openstaxalina:
![AstroNumberingIssue](https://user-images.githubusercontent.com/7863733/56369341-cb0fc680-61be-11e9-9235-e1ac532f1081.png)

openstaxalina:
The numbering should be **continuous** through the following EOC exercise sections:
Review Exercises
Thought Questions
Figuring for Yourself

Stackblocks:
Verified in https://github.com/openstax/content-managers/issues/765

PR #1020, opened 2019-04-18
Issue #1019, astronomy hotfix

tomjw64:
As suspected, this was not a difficult fix. I don't suspect this was a regression from 3-level nesting, since the git blame indicates this config option has been this way since the astronomy recipe has been created.


Stackblocks:
Displaying as specified with updated version of content / code.

image

PR #1021, opened 2019-04-19
Add support for creating multiple indexes

### LINKED ISSUE #1010: Enable multiple indices types
The Polish recipe for University Physics needs more than one kind of Index: names index, foreign terms, etc. in addition to our current terms. 

This would be a useful feature for all recipes and should be integrated in our code base. The current work for it has been started here: https://github.com/openstax/cnx-recipes/pull/377/files

It should be extracted and combined with the recipes to enable multiple indices for all books.

aiwenar:
Resolves #1010.

This PR adds support for creating multiple _partial indexes_, that is indexes which contain only a part of all terms defined in a book.

Partial indexes are composite pages with `specialPageType` set to `$PAGE_INDEX_PARTIAL`. This means that books which do not explicitly opt-in into this behaviour are unchanged. Partial indexes take the same configuration option as normal indexes (`$PAGE_INDEX`), with two additions:

- `indexType` determines which terms will be displayed in this index (more on this later)

- `defaultIndex` marks a partial index as the default index. All terms which were not assigned an index will be sorted into the default one.

As an example compare _University Physics_ with its Polish adaptation, _Fizyka dla szkół wyższych_. _University Physics_ defines only a single [index](https://cnx.org/contents/1Q9uMg_a@13.4:pvcKRWwr@13.4/Index), using this configuration:

```sass
(className: "index",        clusterBy: $CLUSTER_NONE, specialPageType: $PAGE_INDEX, name: "Index"),
```

_Fizyka dla szkół wyższych_ uses three indexes: [Skorowidz nazwisk](https://cnx.org/contents/TqqPA4io@1.126:JXYkqYBo@1.126/Skorowidz-nazwisk) (index of surnames), [Skorowidz rzeczowy](https://cnx.org/contents/TqqPA4io@1.126:lyD6ZIjV@1.126/Skorowidz-rzeczowy) (index of terms), and [Skorowidz terminów obcojęzycznych](https://cnx.org/contents/TqqPA4io@1.126:eX5qwRnb@1.126/Skorowidz-termin%C3%B3w-obcoj%C4%99zycznych) (index of foreign terms), using this configuration:

```sass
(className: "index-name",     clusterBy: $CLUSTER_NONE, specialPageType: $PAGE_INDEX_PARTIAL, indexType: 'name',    name: "Index of names"),
(className: "index-term",     clusterBy: $CLUSTER_NONE, specialPageType: $PAGE_INDEX_PARTIAL, indexType: 'term',    name: "Index of terms",         defaultIndex: true),
(className: "index-foreign",  clusterBy: $CLUSTER_NONE, specialPageType: $PAGE_INDEX_PARTIAL, indexType: 'foreign', name: "Index of foreign terms"),
```

Terms are assigned to a particular index using attribute `index` from namespace `http://katalysteducation.org/cxlxt/1.0`. If this attribute is defined on a `<term>` element with a value matching `indexType` of one of the defined partial indexes it will be sorted into it. Terms without this attribute will be sorted into partial index marked as `defaultIndex`. Terms with `index` set to a value which doesn't match any defined index will be ignored.


aiwenar:
They function slightly differently. `PAGE_INDEX` collects all terms, regardless of value of `cxlxt:index` while `PAGE_INDEX_PARTIAL` with `isDefault = true` collects only terms _without_ `cxlxt:index`. I aggree that it would be best to unify them, especially since now terms whose `clxst:index` doesn't match any index partial are removed, but the only way of doing it I can think of would involve a significant change to how indexes are assembled.

PR #1025, opened 2019-04-22
feat: lint the JS files

philschatz:
This lints the JS files. I like to add `--fix` to the command so that I don't have to wait until Travis runs to tell me that I have a lint problem.


philschatz:
Most of the work in here was getting Travis & Concourse to use the new node version. https://github.com/openstax/ci-images was updated & will need to be updated each time a repo needs to use a newer version of node or python.

PR #1030, opened 2019-04-24
feat: set framework settings without creating a function

philschatz:
This allows a setting to reference another setting using a `(_ref: targetSetting)` rather than rely on being inside a function closure. It also updates the `debug_settings()` to show when a setting has been overridden.

# Example

```sass
@include add_settings((

  primaryColor: blue,

  ChapterNote: (
    _selectors: ('.chapter > '),
    'container': (
      color: (_ref: primaryColor),
    )
  )
));
```


tomjw64:
I would like to merge this asap, but I also might want to apply the new invalid reference error to subspace groups

PR #1044, opened 2019-05-01
Fix release chart

tomjw64:
The update to the latest node version broke the release flag chart because `order` now is `'0'` rather than `0` and no longer passes the the if statement


js/check-release-flag-enabled/extract-chart.js
```diff
@@ -9,7 +9,7 @@ getStdin().then(recipeFeaturesScss => {
     const recipeName = rendered.vars['global']['$RECIPE_NAME']['value']
     const features = rendered.vars['global']['$RELEASE_FLAGS']['value']
 
-    if (order === 0) {
+    if (order === '0') {
```
philschatz: Not sure if it's worth adding but I think you can also do:
```js
const order = Number.parseInt(process.argv[2])

...

if (Number.isNaN()) { yell and fail }
```

PR #1045, opened 2019-05-03
Allow for platform specifics, repoint build scripts

tomjw64:
Also: reuse the custom import-once importer for both testing and building

### DIFF
--- a/styles/build/build.js
+++ b/styles/build/build.js
@@ -4,7 +4,7 @@ const sass = require('node-sass')
 const { String: SassString } = sass.types
 
 const platform = (process.env.PLATFORM || 'NONE').toLowerCase()
-const platformIncludesPath = `${__dirname}/../platforms/${platform}`
+const platformIncludesPath = `${__dirname}/../framework/platform/${platform}`
 if (!fs.existsSync(platformIncludesPath)) {
   throw new Error(`Specified platform '${platform}' does not have an associated directory.`)
 }
--- a/styles/build/build.js
+++ b/styles/build/build.js
@@ -1,6 +1,7 @@
 const fs = require('fs')
 const sass = require('node-sass')
 const toConstantCase = require('to-constant-case')
+const importOnce = require('./../../js/node-sass-importer/import-once')
 
 const { String: SassString } = sass.types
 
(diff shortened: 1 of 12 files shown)


styles/framework/enum/_init.scss
```diff
@@ -37,3 +37,7 @@ $FRAMEWORK_ENUM_GLOBAL: ();
   }
   @return map-deep-get($FRAMEWORK_ENUM_GLOBAL, $name);
 }
+
+@function is_enum($value) {
+  @return (str-index($value, "ENUM__") == 1);
```
philschatz: Can/should this check `$FRAMEWORK_ENUM_GLOBAL` so that someone does not accidentally write `PLATFORM=web ./build.js` ?

styles/test/test-framework.scss
```diff
@@ -968,3 +988,45 @@ $test_note_grouped: (
...
+          color: green;
+        }
+      } @include expect {
+        platform: pdf;
+        color: blue;
+      }
+    }
+  }
+  @include it('allows function-based platform branching') {
+    @include assert-equal(
```
philschatz:
```suggestion
    @include assert-equal(
      if_platform(enum('Platform:::REX_WEB'), "hello", "goodbye"),
      "goodbye"
    );
    @include assert-equal(
```

I did not see the else case being tested

styles/framework/enum/_init.scss
tomjw64: It is not necessary, because `enum($string)` function checks to see if the key exists in `$FRAMEWORK_ENUM_GLOBAL` already, so as long as the value has the form of an enum, we can be sufficiently sure that the key was checked for existence.

styles/build/build.js
```diff
@@ -1,29 +1,37 @@
...
+
+const platform = (process.env.PLATFORM || 'none')
+const platformIncludesPath = `${stylesRoot}/framework/platform/${platform}`
+if (!fs.existsSync(platformIncludesPath)) {
+  throw new Error(`Specified platform '${platform}' does not have an associated directory.`)
+}
+
 const getPlatform = () => {
-  return SassString(platform)
+  return SassString(toConstantCase(platform))
```
philschatz: Since specifying the platform is done inside scripts is this necessary? It may also obscure if names change.

PR #1052, opened 2019-05-06
Add framework tutorial

tomjw64:
# [View Rendered document here](https://github.com/openstax/cnx-recipes/blob/framework-tutorial/styles/README.md)


styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+`NOT_NULL` - Pick this option when the property must be present, but also must be determined by the style author. For example, we know that notes `type-one`, `type-two`, and `type-five` are outlined the same way, but with a different color. Thus the color property must be present, but it is up to the style author as to what the color should be.
+
+`GROUPED` - Pick this option if it is desired for properties to share the same values. The `_groups` key of the superset can define a key and value, the key of which is the name for the group, and the value of which is the shared property value. For example, if we were to define our note border colors seperately for some reason, we could group border-left|right|top|bottom-color together to `(enum('ValueSet:::GROUPED'), note-box-color)` and in `_groups` we add the entry `note-box-color: enum('ValueSet:::NOT_NULL')`
+
+### Writing the superspaces
+
+Open the file we just created in your favorite editor to get started writing our design superspaces. We will write our components to variables to that they can be reused.
```
brittany-johnson:
```suggestion
Open the file we just created in your favorite editor to get started writing our design superspaces. We will write our components to variables so that they can be reused.
```

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+This is sufficient to remove the previous error, but a new one arises again. Let's skip ahead to the point where we have resolved each error after a series of failed compilations to the point where we get a successful one. The final result should be:
+```scss
+// book.scss
+
+// ...imports above
+
+@include add_settings((
+  TypeOne: (
+    _selectors: (".page > .type-one"),
+    container: (
```
brittany-johnson: I've been adding settings in this format, but I'm confused on when to write them with the namespace indicator like on line 263 above. 
`'container:::border-color': grey,` vs `container: ( border-color: grey,)`
@tomjw64

tomjw64: The two are completely interchangeable (more specifically, the second one is converted into the first one), so it's up to you which you like better and would like to use in a particular circumstance!

brittany-johnson: Ahhhhh, okay cool. Thanks!

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+    _selectors: (".page"),
+  )
+));
+
+// ...uses below
+```
+
+### Finishing up
+
+The preceding book file will yield some CSS, but we aren't yet finished. We have to handle the values which we deemed NULLABLE, such as the font weight for our type-one notes. This means to finish our styles, we must add the following to our settings:
```
philschatz:
```suggestion
The preceding book file will yield some CSS, but we aren't yet finished. We have to handle the values which we deemed `NULLABLE`, such as the font weight for our type-one notes. This means to finish our styles, we must add the following to our settings:
```

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+To add settings, we can call the framework provided mixin `add_settings`. `add_settings` takes a variable number of arguments, but each must be a map of settings. Let's call the `add_settings` mixin to resolve the compiler error we received earlier.
+
+```scss
+// book.scss
+
+// ...imports above
+
+@include add_settings((
+  'TypeOne:::_selectors': (
+    ".page > .type-one"
```
philschatz: When are double-quotes used vs single-quotes? Is it to clarify that a string is a selector? I ask because it seems like GitHub's syntax highlighting stops when it sees double-quotes, but maybe it stops for a different reason.

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+mkdir -p ./styles/books/accounting && touch ./styles/books/accounting/book.scss
+```
+
+In the blank book.scss file, let's import our framework and the design we have just created.
+```scss
+// book.scss
+@import 'framework/framework';
+@import 'design/example/design';
+```
+
```
helenemccarron: It might be useful to add a step to create the platform folder so the script doesn't error: `mkdir -p ./styles/framework/platform/pdf`

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+);
+$boxed_note_title: (
+  _name: 'title',
+  _subselector: ' > .title',
+  _properties: (
+    font-weight: enum('ValueSet:::NULLABLE')
+  )
+);
+$bg_note_container: (
+  _name: 'container',
```
helenemccarron: It might useful to describe what this refers to. By using a generic term such as "container" (which I assumed to be a keyword like in recipes), I didn't realize that this is used to "name" the component and that I'd be calling it later.

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+  _subselector: '.note',
```
helenemccarron: Similarly, this is set but not directly accessed later. It feels like those three fields are distinct if I understand correctly:
`_name` creates an "ID" to access the `_properties`
` _subselector` is set but cannot be accessed or modified in the settings later
`_properties` are accessed in the settings  via `_name`

However, at this point, as a framework user, I have no way of telling them apart. Later, when accessing them, I have to remember that `_selectors`, which is not defined here but elsewhere must be defined as well, that I must ignore `_subselector` and add values for each entry in `_properties`, which I do not reference directly. I feel that a bit more guidance on what the user should keep in mind when writing their first design space would help crystallize their understanding faster.

styles/README.md
philschatz: I was also initially confused by this term.

One thing that was discussed was to use a convention for strings: no-quotes for framework fields (like `_name` on this line) and single-quotes for user-defined things. The framework can enforce this convention by using a SASS function to check what type of string was used.

[str-is-quoted](https://github.com/sass/sass/issues/1291#issuecomment-393700250) should be available in SASS 4 but the link is to a polyfill just-in-case.

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+  _properties: (
+    color: (enum('ValueSet:::NULLABLE'), white),
```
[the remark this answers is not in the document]
tomjw64: It is passing the default. I'm not sure how used they will be. Remains to be seen... As for the ambiguity, as far as Sass's type system is concerned, when we pass `white` it is the same thing as passing a space-separated list with only a single value of `white`. Adding a space separated list with more values in it should not be concerning.

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+```scss
+// book.scss
+@import 'framework/framework';
+@import 'design/example/design';
+```
+
+Let's try to build our new book.scss. If there is an entry added to books.txt for your book style you can use
+```bash
+# bash
+./script/build-styles # This will output css in the ./styles/output directory
```
tomjw64:
```suggestion
./script/build-styles accounting # This will output css in the ./styles/output directory
```

styles/README.md
```diff
@@ -0,0 +1,323 @@
...
+```bash
+PLATFORM=pdf node ./styles/build/build.js ./styles/books/accounting/book.scss
+# => Error: Error [ENUM__Error:::INDEX_ERROR]: Unknown setting `TypeOne:::_selectors`.
+```
+
+We got an error. The framework has told us we need to specify selectors in our settings. This is a good opportunity to go over how settings work.
+
+### Settings
+
+To add settings, we can call the framework provided mixin `add_settings`. `add_settings` takes a variable number of arguments, but each must be a map of settings. Let's call the `add_settings` mixin to resolve the compiler error we received earlier.
```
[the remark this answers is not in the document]
tomjw64: It's a relic of adding functions as settings files. I haven't used it except for testing. Think it should be removed?

styles/README.md
tomjw64: From what I've seen, many syntax highlighters don't know how to handle the `:::`, even when its in a string, so that might be it.

styles/framework/settings/_init.scss
```diff
@@ -60,6 +86,19 @@ $FRAMEWORK_USED_SETTINGS: ();
   @include remove_settings($settings...);
 }
 
+/// This function is used to get values from the globally-accessible settings.
+///
+/// @param {String} $name - The key for the desired setting
+/// @param {String} $default [enum('Object:::NONE')] - The default value to use if a setting for the key is not found
+///
+/// @group publicapi
```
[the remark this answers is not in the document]
tomjw64: Oops! Great point! I'd just been used to calling it so much. This is why it's good to have fresh eyes!

tomjw64:
```suggestion
/// @group internal
```

styles/README.md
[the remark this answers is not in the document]
philschatz: hmm, in that case if defaults are possible then would the following be more legible since fields with defaults are always nullable:

```suggestion
    color: default(white),
```

styles/README.md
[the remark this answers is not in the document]
philschatz: (⚠️ bikeshedding) Ah, for testing you might want to add multiple settings within a single scope. In that case it seems like allowing multiple arguments to `scoped_add_settings` makes sense. And `add_settings` could still support multiple settings but the multiple arguments could be de-emphasized in the documentation or even disabled in the code (up to you).

It just seems more readable to me to have each setting as a separate call.


Another option could be that `add_settings` takes 1 argument: either a map or a list of maps

styles/README.md
philschatz: ah, ok. That's probably the culprit.

styles/README.md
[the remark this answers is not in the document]
tomjw64: fields with defaults are not necessarily nullable. See `$test_subspace_with_default` in the framework tests and the test that uses it `allows ValueSets to include a default value if setting is missing`

PR #1053, opened 2019-05-09
[WIP] Test chemistry styles

brittany-johnson:
Excuse the mess, very much a WIP


styles/templates/template1/template1.scss
```diff
@@ -0,0 +1,7 @@
+@import './styles/designs/carnival/notes';
+@import './styles/designs/carnival/intro';
+
+@include use('icon-feature-2', 'icon-feature');
```
philschatz: Rather than have `add_settings` and `use`, would it be useful to support an optional 3rd argument to use which allows you to provide settings. So something like this:

```sass
@include use('icon-feature-2', 'icon-feature', (
    _selectors: ('.link-to-learning'),
    'container': (
        border-bottom: '1.5pt solid #{$G3}',
    ),
));
```

styles/designs/carnival/notes.scss
```diff
@@ -0,0 +1,113 @@
+//will RM later, just testing
+@import url('https://fonts.googleapis.com/css?family=IBM+Plex+Sans');
```
philschatz: Does/should the framework need to support a way of adding fonts? This line is an example of stuff that's is unknown to the style compiler & therefore it might be useful to somehow include it in the framework.

styles/framework/function/_color-scheme.scss
```diff
@@ -0,0 +1,36 @@
...
+
+@mixin update-color-scheme-config() {
+  $color-scheme: openstax-color-scheme() !global;
+  @each $scheme-get-function in $color-scheme-manifest {
+    $color-scheme: map-merge($color-scheme, call($scheme-get-function)) !global;
+  }
+}
+
+// Retrieves color value from color-scheme
+@function color-scheme($name) {
```
philschatz: Is this function used? How is `color-scheme('foo')` different from `settings('foo')`? (or `settings('ColorScheme:::foo')` if you prefer to have a namespace for the mapping.


As a more concrete example, wouldn't the following:


```sass
@include add_settings((
    something: color-scheme('foo')
));
```

be the same as:

```sass
@include add_settings((
    something: ref('foo')     // or (_ref: 'foo') which is the notation for how it works now
));
```

See https://github.com/openstax/cnx-recipes/pull/1030

styles/designs/carnival/notes.scss
tomjw64: What are you envisioning? I don't see any other reasonable way to get the fonts included except for an import statement.

philschatz: Oh, just that rather that writing something like 

```sass
@import url('https://....');
...
@include add_setting((
    'whatever...:::font-family': 'IBM Plex Sans'
));
```

we could write:

```sass
@include uses_font('IBM Plex Sans', 'https://...');   // this just calls @import and remembers the font name
...
@include add_setting((
    'whatever...:::font-family': font('IBM Plex Sans')
));
```

By using `font(...)` the framework could error when the font URL is not provided.

Also, instead of relying on people remembering to use the `font(...)` function, the `use` function could special-case `font-family` and check to see if the fonts are defined. This last suggestion is kind of hacky and might not always work so it might not be so useful.

PR #1061, opened 2019-05-13
Documentation for CLI

edwoodward:
Started with Karen's PR documentation and added some additional details about parameters.

### DIFF
--- a/script/README.md
+++ b/script/README.md
@@ -123,7 +123,7 @@ CLI Usage
    1. If you want to mathify all of the books, use the --all flag. If a book in books.txt is not found, the process will fail:
 
         ```
-        docker-compose run --rm bake-book --all
+        docker-compose run --rm mathify-book --all
         ```
 
     2. If mathifying a single book, a short book name must be used as detailed in books.txt. An example is `intro-business` in the first example.
--- a/script/README.md
+++ b/script/README.md
@@ -1,0 +1,167 @@
+
+Baked Output CLI Documentation
+==============================
+
+Setup
+-----
+
+1. If you're concerned about messing up your work directory, feel free to clone this in a different directory:
+
+   ```
+   git clone -b cli-master https://github.com/openstax/cnx-recipes.git cnx-recipes-cli
+   cd cnx-recipes-cli
+   ```
+
+1. Install docker and docker-compose (docker-compose may already be bundled with docker):
+
+   ```
+   ./script/install-docker
+   ```
+
+   This should work for people using ubuntu and brew on Mac OS. For everyone else, you probably need to download docker here: https://www.docker.com/products/docker-desktop
+
+   There is no need to run other setup scripts.
+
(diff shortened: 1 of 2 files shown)


script/README.md
```diff
@@ -0,0 +1,167 @@
...
+   1. If you want to download from a specific host:
+
+      ```
+      docker-compose run --rm -e HOST=katalyst01.cnx.org fetch-book --with-resources intro-business
+      ```
+
+    2. To download all of the books in books.txt, use the --all flag:
+
+        ```
+        docker-compose run --rm fetch-book --all
```
karenc: This doesn't work because there are some books in books.txt that aren't online it seems -- principles-management and entrepreneurship:

https://github.com/openstax/cnx-recipes/blob/master/books.txt#L51-L52

Do you think we should still include this?

script/README.md
```diff
@@ -0,0 +1,167 @@
...
+        
+        ```
+    3. If downloading a single book, a short book name must be used as detailed in books.txt. An example is `into-business` in the first example.
+
+    4. The `--rm` flag is used to remove the previous docker image if there is one.
```
karenc: The `--rm` flag is used to remove the container after running the command.  It doesn't remove any images.  From the official help text `docker-compose run --help`:

```
--rm                  Remove container after run. Ignored in detached mode.
```

script/README.md
```diff
@@ -0,0 +1,167 @@
...
+   ```
+   docker-compose run --rm mathify-book intro-business
+   ```
+
+   This creates `./data/intro-business/collection.mathified.xhtml`.
+
+   1. If you want to mathify all of the books, use the --all flag
+
+        ```
+        docker-compose run --rm bake-book --all
```
karenc: This should be `mathify-book` instead of `bake-book`.

script/README.md
```diff
@@ -0,0 +1,167 @@
...
+      ```
+      docker-compose run --rm -e STYLE=recipes/output/my-style.css build-pdf intro-business
+      ```
+    2. If you want to generate pdfs for all of the books, use the --all flag
+
+        ```
+        docker-compose run --rm build-pdf --all
+        ```
+
+    3. If generating a pdf for a single book, a short book name must be used as detailed in books.txt. An example is `into-business` in the first example.
```
karenc: Typo `into-business` <- missing `r`

script/README.md
edwoodward: @karenc I can add "If a book in books.txt is not on the host, it will fail"

karenc: I think the existing script will fetch all the books up to the first book that fails and exit.  I don't know how long it's been since people used `--all`.

If you have the script set up, may be you can try doing `fetch-book --all` and see if it actually works...?

script/README.md
edwoodward: @karenc Fixed

script/README.md
edwoodward: @karenc Fixed

Issue #1060, opened 2019-05-13
Use Config_UnnumberedElements


tomjw64:
Convert the library to use `$Config_UnnumberedElements` rather than `$Config_UnnumberedExercises` and `$Config_UnnumberedEquations`

openstaxalina:
Discussed with @Adrianltd and @tomjw64 -- the CSS output should not change.  Therefore, there should be no need for regression testing.

Adrianltd:
We previously had `unnumbered equations` being called in the generator at pass 1 and `unnumbered exercises` being called in the generator at pass 3.
During the `precalculus recipe` creation I needed to add `unnumbered figures`, so instead - as we have discussed in https://github.com/openstax/cnx-recipes/pull/836 - I've unified all that 3 options to be `unnumbered elements`.
@tomjw64 we still might need to do some regression testing. Because there's less uses of unnumbered equations than unnumbered exercises, I think it's safer to put that `$Config_UnnumberedElements` to pass 3. Then there's only `physics` and `ap-physics` to be tested. Although I don't see any possible issues caused by that, there's nothing happening with equations between pass 1 and 3.

Adrianltd:
Solution is merged, I'm closing the issue.

PR #1062, opened 2019-05-15
1060 use unnumbered elements

Adrianltd:
https://github.com/openstax/cnx-recipes/issues/1060


tomjw64:
It just occurred to me I can automate the regression tests for this one. Standby...

tomjw64:
I have diffed the books before and after and have confirmed that no regressions occurred. Merge away!

PR #1128, opened 2019-06-11
Add not_implemented function

tomjw64:
To be used temporarily in place of spacing and rhythm functions, which are not yet implemented. Can be repurposed for other not yet implemented features as well.


styles/framework/function/_not-implemented.scss
```diff
@@ -0,0 +1,5 @@
+/// @group publicapi
+@function not_implemented($value) {
+  @warn "A not yet implemented function was meant to be called.";
+  @return $value;
+}
```
tomjw64:
```suggestion
}

```

Issue #1077, opened 2019-05-20
carnival theme: learning objectives


lwtchu:
![Screen Shot 2019-05-20 at 3 58 23 PM](https://user-images.githubusercontent.com/11299125/58051567-26173f00-7b18-11e9-872c-e1c7470939cd.png)

"By the end of this section, you will be able to" styling
IBM Plex Sans Medium, italicized (color is book specific)
List font: Muli, regular, base font size
Bullets match the book specific color

openstaxalina:
The attached PR modifies the new `styles` files.  It only creates files inside of the `styles` folder.  No product that we are using at this time uses those files.  

Therefore, we will include this in today's (June 21) code freeze.

openstaxalina:
Devs and Alan need to meet next to go over the test procedure.

brittweinstein:
@Stackblocks based on the meta info above, I'm thinking this is ready to be retested? Please let me know if that is not the case.

Stackblocks:
Displaying as specified with updated version of content / code.
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/a8561575-fbbf-4adc-a440-d5bea24f1809)

Stackblocks:
Closing this issue and spinning off an issue for the font fix.

helenemccarron:
Updated the font size to be  font-scale(1) for the learning objective titles and list items (originally set to values bigger than base font size but not in the scale)

Stackblocks:
Something broke this since the closing of the issue. Re-opening for fix.

## State as of 2019-08-15 T 12:00-06
![image](https://user-images.githubusercontent.com/12678235/63111865-4f4b3e00-bf54-11e9-8d98-a1394e5b70d5.png)

Stackblocks:
Fonts were being imported incorrectly and the styling was spilling over into the list text. Resolved.

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/4eac75ba-dd1a-47da-a097-5dd30b43c46f)

PR #1130, opened 2019-06-11
Learning Objectives with Thomas' comments added

clee1060:
Added Thomas' comments. Take another look though, my old git branch was messed up so everything should be cleaned up now.
https://github.com/openstax/cnx-recipes/issues/1077


styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+
+@include add_settings((
+    LearningObjectiveTitle: (
+        _selectors: ('[data-type=\'page\']'),
+        title: (
+            color: $B2,
+            font-family: 'IBM Plex Sans',
+            font-style: italic,
+            font-weight: 900,
+            font-size: 13px,
```
helenemccarron: can you please replace hard coded values with the new function `not_implemented()`, this will help refactoring the files when the file vertical rhythm function will be available.

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+        container: (
+            margin-bottom: 20px,
+            padding-left: 13px,
+        ),
+        listItem: (
+            color: black,
+            font-family: 'Muli',
+            font-style: normal,
+            font-weight: normal,
+            font-size: 12px,
```
helenemccarron: Please use `not_implemented()`  (see above) - we will use a font scale in the future

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+        container: (
+            margin-top: 0px,
+        ),
+    ),
+
+    // Actual list
+    LearningObjectiveWrapper: (
+        _selectors: ('[data-type=\'abstract\']'),
+        container: (
+            margin-bottom: 20px,
```
helenemccarron: Please use `not_implemented()`  (see above)

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+            font-size: 12px,
+            padding-bottom: 3px,
+            list-style: none,
+        ),
+        listDot: (
+            color: $B2,
+            content: "\"\2022\"",
+            display: inline-block,
+            width: 1em,
+            margin-left: -1em,
```
helenemccarron: Please use `not_implemented()`  (see above). Also please note that the unit we are enforcing is `rem`s

styles/books/chemistry/book.scss
tomjw64: ^^ more specifically `13px` becomes `not_implemented(13px)`

styles/designs/carnival/parts/_learning-components.scss
```diff
@@ -0,0 +1,52 @@
+// Learning Objectives Components
```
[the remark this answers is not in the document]
tomjw64: I think it might be useful. It will certainly be easy to change if we don't like it :)

styles/designs/carnival/parts/_learning-components.scss
```diff
@@ -0,0 +1,52 @@
...
+$LearningLearningTitle: (
```
tomjw64: This is a bit of an odd variable name. Was it supposed to be `LearningObjectivesTitle`? Make sure you recompile to make sure it's still working after the change.

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+        _selectors: ('[data-type=\'page\']'),
+        title: (
+            color: $B2,
+            font-family: 'IBM Plex Sans',
+            font-style: italic,
+            font-weight: 900,
+            font-size: 13px,
+        ),
+        container: (
+            margin-top: 0px,
```
tomjw64: Doesn't look like this declaration is making it into your compiled CSS. Is it necessary?

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1,51 @@
...
+@import '../../designs/carnival/parts/learning-spaces';
+@import url('https://fonts.googleapis.com/css?family=IBM+Plex+Sans');
+@import url('https://fonts.googleapis.com/css?family=Muli');
+
+// Colors
+$B2: #3172B6; 
+
+@include add_settings((
+    LearningObjectiveTitle: (
+        _selectors: ('[data-type=\'page\']'),
```
tomjw64: This would imply that the element selected by `[data-type='page']` represents the LearningObjectiveTitle, which I'm not sure I would agree with. We've decided in previous discussions (of which you were not a part) that leaving the top level component subselectors blank is good practice. Perhaps this value should be `[data-type='page'] > ul` and the `_subselector` field for the space should be empty rather than `> ul`

tomjw64: With that said, I think you _should_ be able to fit this into a single superspace, rather than two. I've been wrong before, though.

tomjw64: @clee1060 Don't know how this got marked as resolved, so I'll ping you here.

styles/designs/carnival/parts/_learning-components.scss
tomjw64: @clee1060 Don't know how this got marked as resolved, so I'll ping you here.

clee1060: @tomjw64 I resolved it because I changed it, did the changes not get pushed?

styles/books/chemistry/book.scss
clee1060: @tomjw64 I resolved it because I changed it, did the changes not get pushed?

styles/designs/carnival/_learning-components.scss
```diff
@@ -0,0 +1,50 @@
+// Learning Objectives Components
+$LearningObjectiveTitle: (
+  _name: "title",
+  _subselector: '',
```
tomjw64: Is there any subselector for the title? Perhaps a `[data-type="title"]` or an `h2` or `h3`. Doubt it is plaintext as a child of the abstract

tomjw64: Giving this a selector would allow you to use the WrapperComponent as the (now appropriately named) top-level component and then you may not need so many `font-*` declarations.

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,38 @@
...
+            margin-left: not_implemented(13px),
+        ),
+        listItem: (
+            color: (_ref: 'colorMap:::fontBodyColor'),
+            font-family: (_ref: 'typography:::serifFont'),
+            font-style: normal,
+            font-weight: normal,
+            font-size: not_implemented(12px),
+            padding-bottom: not_implemented(3px),
+            list-style: none,
```
tomjw64: Since this component relies heavily on the `ListDot` component, the `list-style: none` declaration may be worth hardcoding into the design for now.

styles/designs/carnival/_learning-spaces.scss
```diff
@@ -0,0 +1,23 @@
+@import 'learning-components';
+
+// Superspaces
+//Learning global superspace
+@include create_superspace('LearningObjectiveSpace', (
+  _components: (
+    map-merge($LearningObjectiveTitle, (
+      _components: (
+        map-merge($LearningWrapperComponent, (
```
tomjw64: The title should probably be a child of the wrapper, not the other way around.

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,38 @@
...
+        ),
+        listDot: (
+            color: (_ref: 'colorMap:::tocTitleColor'),
+            content: "\"\2022\"",
+            display: inline-block,
```
tomjw64: Since the purpose of this component is to serve as the dot and to do so it must also take up space, it would be implied that this element is `inline-block`. Perhaps this is a declaration best left in the design rather than in the settings since it is essentially a design requirement (and would not be configured)

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,38 @@
...
+            width: not_implemented(1em),
+            margin-left: not_implemented(-1em),
```
tomjw64: Will width and margin-left always be equal but opposite values? Maybe we can set some variable or set some setting to be referenced by both of these so that only one value need be changed for alignment.

styles/designs/carnival/_learning-components.scss
```diff
@@ -0,0 +1,50 @@
+// Learning Objectives Components
```
brittany-johnson: On K.K branch there's a `parts` folder, do we want to place component and space file in that folder? Or just keep them outside of the folder?
@helenemccarron @tomjw64

brittany-johnson:
Besides the question I listed above, everything else looks good!

styles/designs/carnival/_learning-components.scss
```diff
@@ -0,0 +1,50 @@
...
+$LearningListComponent: (
+  _name: 'listItem',
+  _subselector: '> li',
+  _properties: (
+    color: enum('ValueSet:::NOT_NULL'),
+    font-family: enum('ValueSet:::NOT_NULL'),
+    font-style: enum('ValueSet:::NULLABLE'),
+    font-weight: enum('ValueSet:::NULLABLE'),
+    font-size: enum('ValueSet:::NOT_NULL'),
+    padding-bottom: enum('ValueSet:::NOT_NULL'),
```
tomjw64: It might be better to make this NULLABLE like the other margin/spacing declarations in this space, because i doubt this padding really needs to be asked for in all cases. However, I don't think it's worth blocking this PR for if there was some level of reasoning behind picking NOT_NULL for this.

styles/books/chemistry/book.scss
```diff
@@ -0,0 +1 @@
+@import 'designs/carnival/learning-settings'
```
```suggestion
@import 'designs/carnival/learning-settings'

```

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,36 @@
...
+        listDot: (
+            color: (_ref: 'colorMap:::tocTitleColor'),
+            content: "\"\2022\"",
+            width: not_implemented(1em),
+            margin-left: not_implemented(-1em),
+        ),
+    ),
+));
+
+@include use('LearningObjective', 'LearningObjectiveSpace');
```
tomjw64:
```suggestion
@include use('LearningObjective', 'LearningObjectiveSpace');

```

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,36 @@
+@import 'framework/framework';
```
brittany-johnson: Let's import the framework in the book file instead of the settings file. :D

styles/designs/carnival/_learning-components.scss
clee1060: When I asked @helenemccarron and @tomjw64 about this before, they said to keep them in the carnival folder, so I'm gonna resolve this for now.

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,35 @@
+@import 'learning-spaces';
```
helenemccarron: This file should be moved to the template folder.

styles/designs/carnival/_learning-settings.scss
```diff
@@ -0,0 +1,35 @@
...
+@import 'templates/generic/_colors';
```
helenemccarron: Please use the carnival template colors and typography

Issue #1105, opened 2019-05-30
Precal trio footnotes restart by module in webview


oscryan:
# What happened?

Footnote numbering does not match the pdf because it restarts at every module.

# Reproducibility Details

- **Does it also happen in an incognito window?**: yes
- **Does this happen in one place, a few places, or everywhere?**: every footnote

# Environment:

- **Book(s)**: College Algebra
- **URL/Server**: https://staging.cnx.org/contents/mwjClAV_@12.1:dGtL5139@12/6-3-Logarithmic-Functions for example
- **Location (page number, chapter/section)**: footnotes
- **Suggested nearby searchable text**: earthquake.usgs
- **Browser**: Firefox

# Screenshots
<img width="1116" alt="Screen Shot 2019-05-30 at 11 25 22 AM" src="https://user-images.githubusercontent.com/1895432/58647732-cbc36e80-82cd-11e9-9ff6-d007ed820ea2.png">
<img width="1108" alt="Screen Shot 2019-05-30 at 11 25 12 AM" src="https://user-images.githubusercontent.com/1895432/58647733-cbc36e80-82cd-11e9-870c-d27f94a4233d.png">

rosemaryyang:
On PDF version: 
Preface has a footnote that is in Roman numerals. 
![image](https://user-images.githubusercontent.com/50595077/58821596-09d8cf00-85fb-11e9-8fc7-569d033b31bd.png)

Footnote numbering does not restart with every chapter.
![image](https://user-images.githubusercontent.com/50595077/58822536-61783a00-85fd-11e9-8642-4b166b23e05d.png)

openstaxalina:
@rosemaryyang thank you!  So footnote numbering is continuous throughout the entire book (after the preface)?  Could you confirm it never restarts?

rosemaryyang:
Yes, the numbering never restarts.

wilkus27:
@openstaxalina doing the investigation I noticed some differences between precal trio. 

In **Algebra and Trigonometry** on webview exist footnote like in PDF and I can change the numbering for that in recipe.

webview

![image](https://user-images.githubusercontent.com/40228854/59275126-04fbc700-8c5c-11e9-82e5-5bfe2add7ee3.png)

![image](https://user-images.githubusercontent.com/40228854/59275161-11801f80-8c5c-11e9-86b0-6b484322ef31.png)

PDF

[...]

PR #1135, opened 2019-06-14
#1105 Custom footnote numbering

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1105

Numbering of footnotes in precal trio should reset after Preface and then after whole book. Type of counters in Preface should be lower-roman.


recipes/mixins/_modify.scss
```diff
@@ -839,3 +839,18 @@
     class: "chapter-content-module"
   }
 }
+
+@mixin modify_customFootnotes($footnoteConfig) {
+  @if $footnoteConfig {
```
tomjw64: It seems like `$footnoteConfig` will at least be a list, even if empty, in which case the `@each` will suffice to not run the block. Is your output identical if you remove this `@if` check?

wilkus27:
@tomjw64 Removing `@if` condition didn't change output of any book, when I checked it locally running `build-recipes` script, but some checks failed, I'm not sure why.

Issue #1011, opened 2019-04-15
Cases for various element cases in recipes + internationalization work


helenemccarron:
The Polish recipe for University Physics needs tables to support cases in order to be merged. This is the ground work needed to convert our recipes to support different languages. We should implement some of the logic and file structure that would support future work in enabling a more comprehensive approach to localization and different languages support. Some of that work has been started in https://github.com/openstax/cnx-recipes/pull/377, should be extracted and resolved in a separate PR.

openstaxalina:
The goal is to implement such that only the Polish recipe is affected, **none of the other files**.  If this will impact other files, we need to discuss the test load.

brittweinstein:
We need @aiwenar's documentation before work can begin.

openstaxalina:
@paulinaszczucinska will discuss with @aiwenar.  If we could receive at least the first part of the documentation, that could allow Natalia and others to work on getting the Polish University Physics recipe merged.

helenemccarron:
We should focus on removing any hard coded values and order assumptions from the mixins first. This will allow for the Polish recipe to be completed. 

Later, we can reassess the config files's structure and consider moving variables but we should complete the mixin's refactor first.

brittweinstein:
Natalia's Documentation: https://docs.google.com/document/d/1O14g4C2PIOp1M3SFis5bxE867VcVvg6z8J_4d2VwhJY/edit#heading=h.fejamw5qncxb

openstaxalina:
Went with the bare minimum of changes, to recipes only and should not have any impact on currently baked content (all OpenStax books).  cc @helenemccarron and @Stackblocks in case I misstated anything.

brittweinstein:
@wilkus27 this seems to be ready to be merged and go to testing. Is this the case?

wilkus27:
@brittweinstein Yes, I've just merged it.

wilkus27:
**Summary**: problem with order assumption is resolved by https://github.com/openstax/cnx-recipes/pull/1187. There are created new mixins which allow to change order of elements in titles and captions of figures and tables. This solution is an option for future textbooks, so is not implemented to any existing book at this moment.

Here is an example how it could be implemented in future: https://github.com/openstax/cnx-recipes/pull/1187/commits/54127417eedced091724fa0b5e26e5f33469de09.

MartaSztuka:
[...]

PR #1159, opened 2019-06-24
Internationalization

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1011

### DIFF
--- a/recipes/books/_common-stuff.scss
+++ b/recipes/books/_common-stuff.scss
@@ -12,7 +12,11 @@ $Config_hasAbstract: true;
 $Config_abstractTitle: null;
 $Config_PartType_Chapter_TitleContent   : (os-number: counter(chapter), os-divider: ' ' );
 $Config_PartType_Appendix_TitleContent  : (os-number: counter(appendix, upper-alpha), os-divider: ' | ');
-$Config_Section_TitleContent   : (os-number: counter(chapter) "." counter(section), os-divider: ' ', os-text: existingContent);
+$Config_PartType_Section_TitleContent   : (os-number: counter(chapter) "." counter(section), os-divider: ' ' );
+
+$Config_Chapter_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
+$Config_Appendix_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
+$Config_Section_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
 
 $Config_PartType_Table_CaptionContent   : (os-title-label: "Table ", os-number: counter(chapter) "." counter(table), os-divider: " "); //TODO: Make a counter for both chapter and appendix, and automate the switch between standard and upper alpha counting?
 $Config_PartType_Table_CaptionContentAp : (os-title-label: "Table ", os-number: counter(appendix, upper-alpha)  counter(table), os-divider: " ");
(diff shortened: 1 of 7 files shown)


wilkus27:
@helenemccarron I started with removing order assumption from mixin related with sections titles. 

First, I wanted to create some config in which we could decide about order of all spans with specific classes which create title:

```
$Config_SectionTitle(os-number: ... , os-divider: ' ', os-text: content());
```
But at this moment class `os-text` is added by mixin:

```
@mixin modify_spanWrapTitles() {
  :not([data-type="metadata"]) > [data-type="document-title"] {
    content: '';
    &::after {
      content: content();
      container: span;
      class: "os-text";
    }
  }
}
```
which has to be included in some pass before that one in mixin `number_numberSections` is included.

In 72d1990 I'm trying to create config in which we can decide of order of two parts: existing one - created in content and new one - created by recipe. 

Mixin which I created (`utils_titlePartsOrder`) is changing output:

image

At this moment it's creating content for each key. Probably it's easy, but I couldn't figure out how can I take contents from each loop, and then combine them to one.

wilkus27:
@helenemccarron in 72da88c there is my second proposition. It's not so complex as first one and it doesn't change any output. I aplied those changes only for section titles. I want to be sure we want to go in this way before I will try to apply it for other titles.

wilkus27:
I start to analyze captions of tables and figures. Their content is created in that way:

image

At this moment only first part is defined by config.
So I need to find solution similar to the first one, which I pushed. Because in tables captions there are a few parts I will create some function. In this case it's not a good idea to use  previous solution:

image

wilkus27:
@helenemccarron  @tomjw64 in 6eea7e8 I created new file with order functions. For this moment only for figure and table captions and for chapter, appendix and section titles. I put it in mixins directory, but maybe I should move it somewhere else. Sorry for the mess in this PR, I wanted to try different approaches. Only last one commit is important. Please give me your thoughts about this idea. I want to know your opinion before I start working on functions for other elements.

recipes/mixins/_order.scss
```diff
@@ -0,0 +1,39 @@
...
+    $contentValue: null;
+
+    @each $titlePart in map-keys($documentTitleConfig) {
+        $titleValue: map-get($documentTitleConfig, $titlePart);
+        $value: null;
+
+        @if $titleValue == 'partTypeTitleConfig' {
+            $value: pending(b#{$element}Label);
+        }
+        @if $titleValue == 'existingContent' {
```
tomjw64:
```suggestion
        @else if $titleValue == 'existingContent' {
```

`$value` is not guaranteed to be not null after this block, but null is also not an acceptable value for `$value` given how it is used.

recipes/books/_common-stuff.scss
```diff
@@ -14,13 +14,22 @@ $Config_PartType_Chapter_TitleContent   : (os-number: counter(chapter), os-divid
 $Config_PartType_Appendix_TitleContent  : (os-number: counter(appendix, upper-alpha), os-divider: ' | ');
 $Config_PartType_Section_TitleContent   : (os-number: counter(chapter) "." counter(section), os-divider: ' ' );
 
+$Config_Chapter_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
+$Config_Appendix_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
+$Config_Section_TitleContentOrder: (Label: partTypeTitleConfig, Content: existingContent);
```
tomjw64: It looks like the only thing this gives us the ability to do is to put the label before or after the existing title. Is this enough? What if something like a circumfix is necessary around the title?

recipes/mixins/_order.scss
tomjw64: Perhaps also an `@else` block that errors when an acceptable `$titleValue` is not given would be nice as well.

recipes/mixins/_order.scss
```diff
@@ -0,0 +1,39 @@
...
+            $value: content();
+        }
+
+        $contentValue: $contentValue $value;
+    }
+
+    @return $contentValue;
+}
+
+@function order_figTableCaption($figTableConfig) {
```
tomjw64: The logic of this and `order_documentTitle` are extremely similar. Is there a chance they could be merged somehow?

wilkus27:
In 4679792 I created one function for titles and captions but with assumption that structure of `$Config`s is the same like previously. 

I was thinking also about this comment: https://github.com/openstax/cnx-recipes/pull/1159#discussion_r301180631.

And at this moment I think it is enough because recipes don't support other elements.
In titles there are only spans with classes: `os-number, os-divider, os-text`. 
In captions there are: `os-title-label, os-number, os-divider, os-title, os-divider, os-caption`.
If someone wants to add other element he can do it in two ways:

- by creating its content by himself

Define it in some : 
`$Config_CustomPart: (classNameofSpan: contentOfSpan, classNameofSpan: contentOfSpan);`

Then include utils_title mixin in mixin related with title, like for section titles:

image

- by moving some existing content

For example, like in setCaptionHelper mixin:

image

In both cases he has to modify in some way mixins and add new names for a buckets where content will be moved. Because in order function I'm using names of buckets, either I have to change function to be independent of buckets (what will be hard) or I have to think of the names of the future backets and support them in function.

wilkus27:
**Decision:** this proposition will not be used as a solution for https://github.com/openstax/cnx-recipes/issues/1011. This PR could be closed.

PR #1181, opened 2019-07-02
Warn user if style not defined in bakery CLI

karenc:
Previously `bake-book` and `build-pdf` only warned user if they provided
an explicit `STYLE` argument or if style is defined in books.txt but
it's not found.  Add a new warning to let user know when style is not
defined in books.txt.

These are the screenshots of the warnings:

bake-book-warning

build-pdf-warning


tomjw64:
Travis is hanging on this PR, but it looks like the build passed: https://travis-ci.org/openstax/cnx-recipes/builds/555864218. Manually merging.

PR #1182, opened 2019-07-03
Change order of spans in captions

wilkus27:
We want to be able to decide in `config.scss` or `common-stuff.scss` file about order of all spans creating captions of figures and tables: 

image

At this moment we can change order only for `os-title-label`,`os-number` and `os-divider`.

### DIFF
--- a/recipes/books/_common-stuff.scss
+++ b/recipes/books/_common-stuff.scss
@@ -21,6 +21,13 @@ $Config_PartType_Figure_CaptionContent  : (os-title-label: "Figure ", os-number:
 $Config_PartType_Figure_CaptionContentAp: (os-title-label: "Figure ", os-number: counter(appendix, upper-alpha) counter(figure), os-divider: " "); //TODO: Make a counter for both chapter and appendix, and automate the switch between standard and upper alpha counting?
 $Config_PartType_Figure_CaptionContentPr: (os-title-label: "Figure ", os-number: counter(figure), os-divider: " ");
 
+// Here we can determine order of spans creating captions of figures and tabels.
+// bCaption - contains all spans defined in $Config_PartType_Table(Figure)_CaptionContent(Ap)(Pr),
+// bCaptionTitle - contains span with default title of figure or table
+// bCaptionDivider - contains span with blank space
+// bCaptionContent - contains span with default content of caption
+$Config_FigTable_CaptionContentOrder: (bCaption, bCaptionTitle, bCaptionDivider, bCaptionContent);
+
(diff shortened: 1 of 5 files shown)


recipes/mixins/_number.scss
```diff
@@ -348,10 +348,16 @@
       @error "BUG: Invalid captionType '#{$captionType}'";
     }
   }
+
+  $partOne: nth($figTableConfig, 1);
+  $partTwo: nth($figTableConfig, 2);
+  $partThree: nth($figTableConfig, 3);
+  $partFour: nth($figTableConfig, 4);
```
tomjw64: Not a fan of having to manually access each index like this. Should probably use a loop for this.

recipes/books/_common-stuff.scss
```diff
@@ -21,6 +21,13 @@ $Config_PartType_Figure_CaptionContent  : (os-title-label: "Figure ", os-number:
 $Config_PartType_Figure_CaptionContentAp: (os-title-label: "Figure ", os-number: counter(appendix, upper-alpha) counter(figure), os-divider: " "); //TODO: Make a counter for both chapter and appendix, and automate the switch between standard and upper alpha counting?
 $Config_PartType_Figure_CaptionContentPr: (os-title-label: "Figure ", os-number: counter(figure), os-divider: " ");
 
+// Here we can determine order of spans creating captions of figures and tabels.
+// bCaption - contains all spans defined in $Config_PartType_Table(Figure)_CaptionContent(Ap)(Pr),
+// bCaptionTitle - contains span with default title of figure or table
+// bCaptionDivider - contains span with blank space
+// bCaptionContent - contains span with default content of caption
+$Config_FigTable_CaptionContentOrder: (bCaption, bCaptionTitle, bCaptionDivider, bCaptionContent);
```
tomjw64: Interesting idea here. One drawback might be that it forces a config file user to know the names of the buckets already.

recipes/mixins/_order.scss
```diff
@@ -0,0 +1,10 @@
+@function contentOrder($orderConfig, $bucketsNumber) {
+    $value: null;
+
+    @for $i from 1 to $bucketsNumber +1 {
```
tomjw64: This can be an `@each` loop. This would allow you to drop the `4` from the arguments of content order.

wilkus27:
@helenemccarron @tomjw64 In 018eeee I modified function to use `@each` loop and to be useful also for document titles.

One check failed. I don't have account in Concourse to be able to check what is wrong.

wilkus27:
**Decision:** this proposition will not be used as a solution for #1011. This PR could be closed.

PR #1187, opened 2019-07-09
Order mixins

wilkus27:
We want to be able to change order of spans which create titles and captions. 

For an example I changed orders of spans in **American Government**. After some decisions about such solution I will remove changes for this book.


wilkus27:
Assumption for this solution:
1. changing order of spans is optional
2. new recipe (in other language) has to be created in similar way like English recipes (with order assumptions defined in mixins) then someone who is creating configuration file for new book can check what parts contains title or caption and decide of their order
3. orders of spans are defined in `config.scss` file for each book - for this solution is better to do this in that way, not to defined orders in `common-stuff.scss`, because depending on the book selectors or parts of some element can be different
e.g. in some books in caption of table there is one more span then here `$_tableCaption: (os-caption, os-divider, os-title-label, os-number, os-divider);` - with class `os-title`.
4. in English recipe order of spans of titles is the same for a chapter, section and appendix, similarly - order of spans of captions is the same for tables and figures in the whole book. With this solution we can define different order for titles and captions depending on place where elements are (in preface/chapter/appendix).

Advantages of 1. and 2.:
- question is how many languages are using different order? If most of them have the same structure of titles like in English, maybe it is good to make changing of parts of titles optional.

Disadvantages of 1. and 2.:
- person who is creating `common-stuff` and `config` file has to know that in some `$Config_PartType_(...)` order of keys is important and it has to be the same like in English recipe, so for example, first someone has to put 
`$Config_PartType_Table_CaptionContent   : (os-title-label: "Table ", os-number: counter(chapter) "." counter(table), os-divider: " ");`
 in **common-stuff.scss** and then in **config.scss** file of some book 
`$_tableCaption: (os-caption, os-divider, os-title-label, os-number, os-divider);`.
 It could be confusing. 

Advantages of 4.:
- there is a lot of options for titling and captioning, in different parts of book we can define different structure, we can decide what we want to change and if we want to.

Disadvantages of 4.:
- is it necesarry to have so many options? Maybe it will be better to assume that someone who decide that order in titles should be `(os-text, os-divider, os-number);` wants to have this structure for all of the titles, so in that case there should be one mixin changing all titles (from chapters, sections in TOC etc.).

recipes/mixins/_modify.scss
```diff
@@ -852,3 +852,19 @@
     @include number_customFootnotes($incrementParentSelector, #{$counterName}, #{$counterName}Link, $counterType);
   }
 }
+
+@mixin modify_classOrder($classNameList) {
+  @each $class in $classNameList {
+    .#{$class} {
+      string-set: #{$class}Span content();
```
tomjw64: Using `string-set` will not preserve any existing attributes or nesting inside each element. Using `node-set` would be safer if it is possible.

recipes/mixins/_order.scss
```diff
@@ -0,0 +1,39 @@
...
+
+        #{$parentSelector} {
+            > [data-type="document-title"] {
+                @include modify_classOrder($classNameOrder);
+            }
+        }
+    }
+}
+
+@mixin order_usedExistingTitles($usedTitleConfig) {
```
tomjw64: This mixin and `order_documentTitle` and `order_figTableCaption` are exactly the same except for the intermediate selector. It's probably worth it to just tack that on to the selectors and just use this mixin for everything to stay more dry. If you really don't want to repeat the intermediate selectors, it might be a better idea to move the logic to before the mixin is used.

wilkus27:
@helenemccarron  @tomjw64 I minimalized three mixins to one. I was also trying to modify `classOrder` mixin in this way:

image (6)

image (7)

but it's adding extra divs 

image (8)

I need to spend more time to figure out how to change it to make it work properly.

wilkus27:
@helenemccarron @tomjw64 in 30fa268 I changed `classOrder` mixin to use `node-set` not `string-set`. 

There is problem with changing order of parts of section titles in TOC. At this moment it looks like that:

image

Using this selector:

`.os-unit .os-chapter > li > a`

numbers and dividers are also added to titles wich contain only text (like **Introduction** or **Key Terms**).

tomjw64:
@wilkus27 Rather than modifying the ToC code, why not just move the reordering step to before the ToC step (pass < 11) or moving the ToC step to after the reordering step?

wilkus27:
Ok, thank you @tomjw64 for suggestion to problem with TOC. I will try this.

Unfortunately I have found another problem with captions.

In American Government captions have different structure. Some of them don't have a text, only numbered label:

image

At this moment in such case **classOrder** mixin add text from caption before:

image

image

tomjw64:
@wilkus27 I just explored it a little and it seems that it may also possible to avoid the duplicate IDs by using `move-to` with `pending` as opposed to `node-set` with `nodes`

wilkus27:
@helenemccarron @tomjw64 changes proposed by Thomas have solved all problems :tada: Locally everything is displaying properly and after run diff script there are no differences between html on master and this branch. Also after run build-recipes script there are no changes in other outputs.

recipes/mixins/_toc.scss
```diff
@@ -251,7 +251,7 @@
             attr-cnx-archive-shortid: string(page-shortid);
             container: li;
           }
-        }
+        }
```
helenemccarron: You may want to delete the extra whitespaces here

recipes/mixins/_modify.scss
```diff
@@ -852,3 +852,41 @@
     @include number_customFootnotes($incrementParentSelector, #{$counterName}, #{$counterName}Link, $counterType);
   }
 }
+
+@mixin modify_classOrder($classNameList) {
+  @each $class in $classNameList {
+    .#{$class} {
+      move-to: #{$class}Span;
+    }
+    &:deferred {
```
tomjw64: Having a `:deferred` generated for each iteration of this list is going to create a lot of selectors and more matching iterations for easybake, which will mean a longer baking time. It might be better to construct the content value string through the iterations but only actually apply it with a `:deferred` once. I'm imagining something similar to this (but I freehanded this so it's probably not exactly right):
```scss
@mixin modify_classOrder($classNameList) {
  $contentList: #{"content()"};  // Interpolation may not be necessary
  @each $class in $classNameList {
    $contentList: $contentList pending(#{$class}Span);
    .#{$class} {
      move-to: #{$class}Span;
    }
  }
  &:deferred {
    content: $contentList;
  }
}
```

recipes/mixins/_modify.scss
```diff
@@ -852,3 +852,41 @@
...
+@mixin modify_elementsOrder($orderConfig, $mainSelector: null) {
+  @each $order in $orderConfig {
+    $parentSelector: map-get($order, parentSelector);
+    $childSelector: map-get($order, childSelector);
+    $selector: map-get($order, selector);
+    $classNameOrder: map-get($order, classNameOrder);
+
+    @if $parentSelector {
+      #{$parentSelector} {
+        #{$mainSelector} {
```
tomjw64: It seems that `$parentSelector` and `$mainSelector` are strictly prefixes to the `$childSelector`/`$selector`, so why are they necessary. Why can't it all just be included into one single `$selector` variable? Even if one of the two is necessary, it seems like there's at least one unnecessary variable here. This mixin is a bit of a confusing read because of this. Maybe we should push this through and make an issue for a refactor once we use the mixin once or twice?

wilkus27: When I tested changing the order of elements I chose many different selectors. I was thinking that it may be useful to group elements depending on repeating parts of selector. For example for figure captions the main selector is `.os-figure .os-caption-container` but we can choose many different parent selectors and define different order for figures in Preface, Introduction page, regular page, composite page, appendix etc. So with such solution selector in `$Config_Order_(...)` can be shorter. On the other hand, I think that probability that someone may want to define different order for the same element depending on in which part of book it is, is small. Rather he would like to define the same order in the whole book. So in b6d6216 I modified mixin to use only `selector`.

wilkus27:
@helenemccarron @tomjw64 In b6d6216 I made requested changes and tested them locally on American Government changing the order of parts of different elements. Everything was working properly. After that I removed those changes and ran `build-recipes` script. Like previously, there is no change in any output.

PR #1197, opened 2019-07-11
modify base font

helenemccarron:
Set the base font to 10px and include it to the chemistry book so measurements in rems are accurate


styles/books/chemistry/book.scss
```diff
@@ -2,6 +2,8 @@
 @import '../../templates/template1/design.scss';
 @import '../../templates/template1/settings.scss';
 
+@include use('CommonStyles', "CommonBookStuff");
```
```suggestion
@include use('RootTypography', "common:::RootTypography");
```
mostly to accomodate the namespaced convention

PR #1206, opened 2019-07-15
Exercise label order

wilkus27:
In **common-stuff** and **config** files parts of exercises label are defined in oposite order then they are displayed. 

For example:

image

image

We want to define spans forming a label in the same order they are displaying.


wilkus27:
At this moment those changes are enabled only for calculus.

I couldn't find any book which is using label `Exercise #.` . In most of books all content exercises are moved either to EOS or EOC and for such type of exercises labels are created separately. So I couldn't test 2a18259 locally.

wilkus27:
In 0558a02 I made similar change for equations. I didn't put it in flag because this doesn't change any output. There is no book using such case. If equations are labeled, those labels are displaying after equation in every book. 

I tested this locally changing `numberAfterEq` to `false` in `$Config_PartType_Equation` :

image

And it was working properly.

wilkus27:
I ran the `./script/diff-book-prepare` on master and `./script/diff-book` on this branch. In e4392ba there is a file with differences between `./data/calculus-vol1/collection.prepared.xhtml` and  `./data/calculus-vol1/collection.baked.xhtml`.

wilkus27:
**Decision:** this proposition will be considered as a part of work on internationalization framework for other languages than English and Polish.

PR #1212, opened 2019-07-17
Update readme with run command fo docker documentation

helenemccarron:
Just so we don't have to dig through PRs to remember this exists!

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,23 @@
 1. Run `docker-compose run --rm mathify-book intro-business` to convert all the math to svg.
 1. Run `docker-compose run --rm build-pdf intro-business` to create the pdf.
 
+## Run scripts
+You can run commands in Docker:
+
+`docker-compose run --rm command ...`
+
+To run something like:
+
+`PLATFORM=pdf node ./styles/build/build.js ./styles/books/anatomy/book.scss`
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,23 @@
...
+
+`docker-compose run --rm command ...`
+
+To run something like:
+
+`PLATFORM=pdf node ./styles/build/build.js ./styles/books/anatomy/book.scss`
+
+just prepend with ./script/run:
+
+`./script/run PLATFORM=pdf node ./styles/build/build.js ./styles/books/anatomy/book.scss`
```
tomjw64: Does this actually work?
`docker-compose run --rm PLATFORM=pdf node ./styles/build/build.js ./styles/books/anatomy/book.scss` doesn't really make sense to me.

tomjw64: Does this actually work?
`docker-compose run --rm PLATFORM=pdf node ./styles/build/build.js ./styles/books/anatomy/book.scss` doesn't really make sense to me unless `node` is a docker-compose service.

helenemccarron: That's the example Karen provided in her PR, I did not test it, however. I am fine with removing it from here, since `./script/run ./script/build-styles` does the same thing

PR #1240, opened 2019-07-24
Polish University Physics

wilkus27:
Create recipe for Polish University Physics.


wilkus27:
@helenemccarron in cb102a1  I used function from https://github.com/openstax/cnx-recipes/pull/377/files and put it in the `_link.scss` file. I also created new file `_cases.scss` in locale directory with all variables with cases of words. 

I've also found that in Polish book title `Footnote` is not translated. Previously (in https://github.com/openstax/cnx-recipes/pull/377/files) it was resolved in this way:

image

In 9f9cc2f I created new mixin for that.
After ran `./script/build-recipes` there were no changes in outputs of other books.

Issue #1092, opened 2019-05-21
anatomy & physiology: boxed features


lwtchu:
See #1087 for styling specifications

Title: Diseases of the... 
Treatment: Diseases
Class: `anatomy diseases`
Color: #00819A

Title: Disorders of the...
Treatment: Disorders
Class: `anatomy disorders`
Color: #B10069

Title: Aging and the...
Treatment: Aging
Class: `anatomy aging`
Color: #B05E36

Title: Everyday Connection
Treatment: Everyday
Class: `anatomy everyday`
Color: #00819A

clee1060:
<img width="794" alt="Screen Shot 2019-07-25 at 11 01 06 AM" src="https://user-images.githubusercontent.com/16073667/61889758-a452e180-aecb-11e9-9261-2847a17b7c6a.png">
<img width="797" alt="Screen Shot 2019-07-25 at 11 01 33 AM" src="https://user-images.githubusercontent.com/16073667/61889759-a452e180-aecb-11e9-9583-dffe675dd507.png">
<img width="804" alt="Screen Shot 2019-07-25 at 11 01 42 AM" src="https://user-images.githubusercontent.com/16073667/61889760-a452e180-aecb-11e9-8372-b6dd55d2fd97.png">
<img width="799" alt="Screen Shot 2019-07-25 at 11 01 56 AM" src="https://user-images.githubusercontent.com/16073667/61889761-a452e180-aecb-11e9-8ad2-34c49855c260.png">

Stackblocks:
Coloration of the box and title displaying correctly, but getting some intermittent breakages in the stroke on the bottom left/right of the boxed content and a couple of instances of bad page breaks in the boxed content. These can be be broken out into separate issues if needs be.

# Diseases of the...
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/ed39de23-2533-4f99-828f-ea1277168838)

# Disorders of the...
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/c537e478-91e0-4926-a11e-ff9cdebde34b)

# Aging and...
## Closeup
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/8400ad92-071d-4097-9903-ed160321fe00)

## Broken Box Example
[...]

PR #1242, opened 2019-07-24
Anatomy boxed-features

clee1060:
https://github.com/openstax/cnx-recipes/issues/1092


styles/templates/template1/parts/_aging-settings.scss
```diff
@@ -0,0 +1,55 @@
+@import 'templates/template1/colors';
+@import 'templates/template1/typography';
+
+@include add_settings((
```
helenemccarron: Is it possible to reuse those settings? It feels like those two notes should be using sharing a lot of those settings. @tomjw64 I do not remember how we decided we would handle notes specifically but the idea behind the shapes, if I remember correctly was to avoid such repetition, can you provide guidance on how to reuse features without too much repetition, down to the settings? Thank you!

PR #1250, opened 2019-07-29
Make sure cnx-recipes use the latest dependencies

karenc:
We recently made changes to the transforms code so cnx-recipes need to
be updated so the transforms and bake-book results are accurate.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,7 @@
 requests
 pytest
 PyICU==1.9.8
+rhaptos.cnxmlutils>=1.7.0
 cnx-easybake>=1.2.5
+cnx-transforms>=1.2.0
 nebuchadnezzar>=9.1.0


requirements.txt
```diff
@@ -1,5 +1,7 @@
 requests
 pytest
 PyICU==1.9.8
+rhaptos.cnxmlutils>=1.7.0
 cnx-easybake>=1.2.5
+cnx-transforms>=1.2.0
```
tomjw64: Maybe it's because I'm more familiar with how npm dep trees work, but I'm wondering why this is being updated here instead of in the neb dependency: https://github.com/openstax/nebuchadnezzar/blob/master/requirements/main.txt

requirements.txt
```diff
@@ -1,5 +1,7 @@
 requests
 pytest
 PyICU==1.9.8
+rhaptos.cnxmlutils>=1.7.0
```
tomjw64: This wasn't a dependency before. Why is it now?

karenc: This is a dependency of one of the dependencies in [nebuchadnezzar](https://github.com/openstax/nebuchadnezzar/blob/master/requirements/main.txt): [cnx-transforms](https://github.com/openstax/cnx-transforms/blob/master/requirements/main.txt).

It's necessary for transforming cnxml to html.  There are a lot of packages that are not in requirements.txt but are installed as dependencies of the other packages.  I'm trying to make sure we use the latest packages in cnx-recipes...  Not sure if this is the best way.

requirements.txt
karenc: Yeah this is a bit of a problem.  Python doesn't automatically update packages if they're already installed (I think).  As we release newer versions of the different packages, e.g. cnx-transforms, are we supposed to release a new version of nebuchadnezzar just to update the minimum version of a dependency? :/

tomjw64: I think it would be reasonable to have a new version of neb with the change being updating the minimum supported cnx-transforms version, especially if these changes between those versions are breaking in some way. Or if anything else that uses neb might want this update for themselves and we don't want different systems producing different results.

I feel like there's a strong argument that Neb should be the source of truth about what dependencies (and which versions) it needs to do its job well, not it's dependents. On the other hand, I'm fairly green by OpenStax developer standards. Would there be any other devs that might have opinions or would want to weigh in on this?

pumazi: It might be best to pin to a specific version. If we see cnx-recipes as the Baking application then it should pin the version, so essentially addressing what we are trying to do with RERO.

therealmarv: Maybe a bit of overload but can neb be self updating certain packages like cnx-transforms on itself on run? Python does not update on normal `pip install` but `pip install -U`?

Is this too much work?

karenc:
I'll go the `pip install -U` path. New PR soon...

PR #1253, opened 2019-08-01
Issue 1157

richelleyy:
border feature for `.everyday`


styles/templates/template1/_settings.scss
```diff
@@ -14,9 +14,10 @@
 @import 'templates/template1/parts/appendix-settings.scss';
 @import 'templates/template1/parts/index-settings.scss';
 @import 'templates/template1/parts/preface-settings';
-@import 'templates/template1/parts/_eoc-settings.scss';
+@import 'templates/template1/parts/eoc-settings.scss';
```
helenemccarron: I wonder why this is necessary in this PR, was it not working before?

richelleyy: @helenemccarron when I was trying to open these settings files, it shows unable to read or the file doesn't exist. I'm not sure why it goes wrong but it can compile successfully and display designs on webview.

PR #1257, opened 2019-08-02
Always update python dependencies when running scripts

karenc:
cnx-recipes should be using the latest versions of all the packages.
This changes `pip install -r requirements.txt` to
`pip install -U -r requirements.txt` so it'll update all the outdated
packages.

### DIFF
--- a/script/assemble-book
+++ b/script/assemble-book
@@ -85,7 +85,7 @@ for book_config in "${BOOK_CONFIGS[@]}"; do
   rm -rf "${raw_file/$OUTPUT_DIR/.\/data}"
 
   do_progress "Creating HTML file for ${book_name}" \
-    docker-compose run --rm neb /bin/bash -c "neb assemble \"${raw_dir}\" \"${book_dir}\" && chown -R $(stat -c '%u:%g' ./data) \"${book_dir}\""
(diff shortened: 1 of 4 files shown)


script/assemble-book
```diff
@@ -85,7 +85,7 @@ for book_config in "${BOOK_CONFIGS[@]}"; do
   rm -rf "${raw_file/$OUTPUT_DIR/.\/data}"
 
   do_progress "Creating HTML file for ${book_name}" \
-    docker-compose run --rm neb /bin/bash -c "neb assemble \"${raw_dir}\" \"${book_dir}\" && chown -R $(stat -c '%u:%g' ./data) \"${book_dir}\""
+    docker-compose run --rm neb /bin/bash -c "pip install -U -r requirements/main.txt && neb assemble \"${raw_dir}\" \"${book_dir}\" && chown -R $(stat -c '%u:%g' ./data) \"${book_dir}\""
```
pumazi: This is fine and more than acceptable. I just wanted to mention that we may not want to duplicate efforts of requirement mentioning here and in Nebuchadnezzar.

karenc: What do you mean?

PR #1264, opened 2019-08-05
Set dpi at 200 for all books

helenemccarron:
No more huge images!


styles/templates/template1/_settings.scss
```diff
@@ -33,14 +31,16 @@
...
-    RootTypography: (
+    BookRoot: (
       font-family: (_ref: 'typography:::sansFont'), //book default font
       font-size: (_ref: 'typography:::baseFontSize'), //base font size
       line-height: (_ref: 'typography:::baseLineHeight'), //base font height
-      color: (_ref: "colorMap:::fontBodyColor")
-    )
+      color: (_ref: "colorMap:::fontBodyColor"),
+      prince-image-resolution: "auto, 200dpi",
+      prince-background-image-resolution: "200dpi"
```
tomjw64:
```suggestion
      prince-background-image-resolution: 200dpi
```

styles/templates/template1/_settings.scss
```diff
@@ -33,14 +31,16 @@
...
+    ),
   ),
-));
+))
```
tomjw64:
```suggestion
));
```

styles/templates/template1/_settings.scss
tomjw64: Not required. Just more "css-y"

styles/templates/template1/_settings.scss
tomjw64: Should probably include the semicolon, just in case a statement is added after this one.

PR #1266, opened 2019-08-06
Allow pdf docker-compose service to find styles/icons

tomjw64:
Hopefully once this gets merged, devs and QA should be able to use `docker-compose run --rm build-pdf [book]` and get consistent results.


tomjw64:
@karenc That might be nice, since this is turning out to be a not-so-nice kind of patch.

karenc:
@tomjw64 Let me try that out and see how that looks :+1:

PR #1267, opened 2019-08-06
Tag ToC pages for styling

tomjw64:


recipes/mixins/_toc.scss
```diff
@@ -186,12 +223,31 @@
...
+          move-to: eob-toc;
+          class: "os-toc-chapter";
+        }
+        &.os-solution-container::after {
+          content: pending(eoc-toc);
+          attr-cnx-archive-uri: attr(id);
+          attr-cnx-archive-shortid: string(chap-shortid);
+          container: li;
+          move-to: eob-toc;
+          class: "os-toc-solutions";
```
helenemccarron: Is what is unique about the solutions the fact they are a composite chapter? If so, should the class be `os-toc-composite-chapter`?

PR #1273, opened 2019-08-07
Horizontal rhythm

helenemccarron:
A function to align things horizontally using the horizontal unit decided upon by UX.


styles/templates/template1/parts/_index-settings.scss
```diff
@@ -6,7 +6,7 @@
         _selectors: (".os-index-container"),
         Container: (
             column-count: 3,
-            column-gap: not-implemented(5rem),
+            column-gap: grid(1),
```
tomjw64: Where is this function defined? Is it a sass builtin?

helenemccarron: oops forgot to commit the new file 🤦‍♀

Issue #437, opened 2018-07-02
carnival theme: footnotes


damondemar:
![Screen Shot 2019-08-06 at 4 58 46 PM](https://user-images.githubusercontent.com/11299125/62580297-7c6c6200-b86b-11e9-84a9-a1fc40ac86e5.png)

Should be Alegreya Regular 1.2rem (9pt)
Width of line: 110 px (11rem)
Line stroke weight: 0.5pt 
Line is 1 line height above the text

** All footnotes across all titles so be NO SMALLER than 1.2rem (9pt)
Footnotes are generally a superscript number BUT should be able to support symbols as well

lwtchu:
@wilkus27 i added more specs. let me know if you need any clarification!

wilkus27:
Thank you! :smiley:

Stackblocks:
Please take a look at the footnotes, they're interacting weirdly with their neighbors. 

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/ea821edd-05f4-4d3a-baea-6e77fa516a3c)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/2316ca9d-534d-4cd3-80b1-e4fcd5bbd155)

wilkus27:
About first screenshot, this is the part of content, regular list, not footnote.

About second one, generally footnotes should display at the bottom of the page with last paragraph of module which contains footnote call. After discussion with Helene: the future goal will be to prevent the titles and learning objectives to be breaking apart.

Stackblocks:
Following review by Helene and Natalia, marking as closed. We will handle the content breaking issue in a separate tracker.

PR #1275, opened 2019-08-08
#437 - footnotes

wilkus27:
Issue https://github.com/openstax/cnx-recipes/issues/437.


wilkus27:
@helenemccarron This is how it displays currently:

image

I need to removed space between text and border. It's so big because in the content footnotes have title:

image

At this moment I removed it by 

image

But maybe better way will be to trash it by recipes.

wilkus27:
Should I also style footnote calls?

image

Usually, it could be done by prince properties.
But in this case footnotes are created in the content.

This is the markup from `collection.assemble.xhtml`

image

So that's why I change content to `none` for `::fotnote-call` and `::footnote-marker`

image

Please let me know if I wrong understand it.

helenemccarron:
@wilkus27 Regarding the spacing, have you tried `display: none`, this should remove the entire element from the DOM as opposed to removing its content. You may want to research the difference between `display: none` and `visibility: none`, one removes it altogether from the dom, the other one just hides the content. `content:none` is rarely, if not never used in CSS to hide content. 

For your second question, the footnotes should be moved as part of the baking, isn't that the case? (your screenshot shows collection.assembled, not collection.baked. And yes, you should style the footnote marker (the link in the content) as well.

wilkus27:
@helenemccarron Thank you for a hint with `display: none`. That's what I was looking for :) 

About second problem, I think I asked the wrong question. What I want to know is, is it a good solution that I removed footnote call and marker created by prince?

I was confused, because usually there is such markup:

```
Some text <span class="some-class">Content of footnote</span> some text.
```
And using `float: footnote` for class `some-class` we get footnote area at the end of the page with automatically created footnote markers and content of the footnote and also footnote call is created in the place where is the span with this class.

In our case footnote markers and calls are already created:

image

image

image

And content of footnotes is at the end of the module, so when I'm using `float: footnote` for a `[data-type="footnote-refs"]` prince moved the whole container to the footnote area, but it generates footnote call at the end of the module not in the text.

image


So as I said, I removed them and styled the existing ones. 
Should I remove existing calls and markers and try to find a way to generate them by prince in proper place?

styles/templates/template1/parts/_footnotes-settings.scss
```diff
@@ -0,0 +1,51 @@
+@import 'templates/template1/colors';
+@import 'templates/template1/typography';
+
+@include add_settings ((
+    Footnotes: (
+        _selectors: ('[data-type = "footnote-refs"]'),
+        Container: (
+            color: (_ref: 'colorMap:::fontBodyColor'),
+            line-height: line-height(1),
+            font-family: (_ref: 'typography:::baseFont'),
```
helenemccarron: I this necessary, the basefont should be inherited

styles/templates/template1/parts/_footnotes-settings.scss
```diff
@@ -0,0 +1,51 @@
...
+            font-size: font-scale(1),
+        ),
+        Border: (
+            border-top-color: (_ref: 'colorMap:::fontBodyColor'),
+            border-top-width: not-implemented(0.5pt),
+            width: not_implemented(11rem),
```
helenemccarron: can you use the newly implemented grid() function? One grid unit = 10px

styles/templates/template1/parts/_footnotes-settings.scss
```diff
@@ -0,0 +1,51 @@
...
+            padding-bottom: not_implemented(1.8rem),
```
helenemccarron: Can you use the newly implemented vertical-rhythm() function?

styles/templates/template1/parts/_footnotes-settings.scss
```diff
@@ -0,0 +1,51 @@
...
+        ),
+        Title: (
+            display: none,
+        ),
+        List: (
+            list-style: none,
+            margin: not-implemented(0),
```
helenemccarron: Can you please use the spacing functions? Also, in general, we avoid setting top margins and top padding, so it could be avoided, it would be great (this is a special occurance, though, since it's very unlikely a footnote would end up on top of page) but I thought I'd let you know for later :D

styles/templates/template1/parts/_footnotes-settings.scss
wilkus27: Inherited font is IBM Plex Sans. In specs from https://github.com/openstax/cnx-recipes/issues/437 font in footnotes should be Alegreya.

styles/templates/template1/parts/_footnotes-settings.scss
wilkus27: After merged new changes I implemented h-grid.

styles/templates/template1/parts/_footnotes-settings.scss
wilkus27: After merged new changes I implemented v-grid.

styles/templates/template1/parts/_footnotes-settings.scss
wilkus27: I wasn't sure which functions should I implement here. You meant h-grid and v-grid? 

I'm also wondering that maybe this could be removed after merge https://github.com/openstax/cnx-recipes/pull/1284/commits/9ce72737ee1e37e5204ea3c1491143a368d047bd. There margin and padding are specified for `ul` in `_reset.scss`.

PR #1291, opened 2019-08-13
Add EOC border bars

tomjw64:


styles/books/chemistry/book.scss
```diff
@@ -30,6 +30,82 @@
...
+
+.os-eoc.os-glossary-container {
+  page: eoc;
+  prince-page-group: start;
+}
+.os-eoc {
+  page: eoc;
+}
+
+$bandColor: settings('scheme:::B2');
```
helenemccarron: Note for later: it looks like not using variables isn't completely realistic (as proven here ^_^)

Issue #1290, opened 2019-08-13
Carnival Theme: Page Title Styling


Stackblocks:
![2019-08-13_11-26-54](https://user-images.githubusercontent.com/12678235/62959856-1332a480-bdbf-11e9-94a6-293b42b8ae8b.png)
---
![2019-08-13_11-28-16](https://user-images.githubusercontent.com/12678235/62960089-905e1980-bdbf-11e9-92f1-e3dd0353da87.png)
---
![2019-08-13_11-28-47](https://user-images.githubusercontent.com/12678235/62959898-2776a180-bdbf-11e9-82a8-f3391f125d3f.png)
---
![2019-08-13_11-29-53](https://user-images.githubusercontent.com/12678235/62960062-81776700-bdbf-11e9-956c-975b13cc47ea.png)
---

Please go through the style and make sure that page titles are styled consistently. 

Styling for page title: `IBM Plex Light; 3.583rem; 0.75rem padding under`

Please also make sure that the stroke underneath is `1px`

Stackblocks:
Additional information (and other things you can check / work on if time permits) available here.

https://github.com/openstax/cnx-recipes/issues/1072

Stackblocks:
Page titles showing up as Muli #00819A instead of IBM Plex Sans #000000

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/8aae09e7-e2d8-4b65-945f-b693ec3b3415)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/becb9fd9-111b-4f7e-8121-a3c987c74417)

Stackblocks:
Displaying as specified with updated version of content / code.

![2019-08-15_16-08-43.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/8b636411-84ea-49ac-859d-3e577deebd66)
---
![2019-08-15_16-09-49.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/d0a71130-38d8-464b-9463-544137d46e5e)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/94acf465-fc23-44d8-acf6-f3e4a380687d)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/a8b8b085-5774-4b40-bd1b-ccbed549f393)

PR #1296, opened 2019-08-14
#1290 page title styling

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1290

Here is the document with discrepancies: https://docs.google.com/document/d/13jXVqidA7ztSeK8JvH9J4IGJ2iF5N64_gIH66GiBVmY/edit#

Decisions:

```
The things I can confirm for the page titles are they should all be:
- black (#000000)
- Uppercase
- IBM Plex Sans Light,
- font size 3.583 rem.

For font weight we can keep it bolded for now but the key is that the elements are uniform.
```


styles/templates/template1/parts/_preface-settings.scss
```diff
@@ -3,12 +3,15 @@
         _selectors: (".preface"),
 
         Header:(
-            font-family: (_ref: 'typography:::sansFont'),
+            color: (_ref: 'colorMap:::pagetitleColor'),
+            font-family: (_ref: 'typography:::sansFontLight'),
```
tomjw64: Light font + font-weight: bold = ???

helenemccarron: I believe that this was implemented prior to this PR and kept for consistency. @Stackblocks  we may want to open another issue to remove the bold?

PR #1319, opened 2019-08-26
Issue 1305

brittany-johnson:


styles/output/chemistry-pdf.css
```diff
@@ -186,11 +186,6 @@ div[data-type="page"].preface {
   prince-image-resolution: auto, 200dpi;
   prince-background-image-resolution: 200dpi; }
 
-[data-type='chapter'] > [data-type='page'] .os-figure > figure {
```
brittany-johnson: This style doesn't appear to be doing anything so that is why I removed it and didn't place it somewhere else

PR #1342, opened 2019-09-03
Bake-book to use book instead of template for style injection

tomjw64:


wilkus27:
Tested locally. `bake-book` script working properly :tada: 

But similar problem occurs also for `build-pdf` script:

image

tomjw64:
@wilkus27 Thanks for the good eye. I've patched that as well!

wilkus27:
You're welcome. Working perfectly :confetti_ball:

PR #1372, opened 2019-09-09
Fix sup and sub scripts line height

helenemccarron:


styles/output/apbiology-rex-web.css
```diff
@@ -54,6 +54,18 @@
...
   margin-bottom: 1.5rem; }
 
+sub, sup {
+  font-size: 75%;
+  line-height: 0;
+  position: relative;
+  vertical-align: baseline; }
+
+sup {
+  top: -0.5rem; }
```
tomjw64: If aligned to the baseline but then shifted by a constant amount, will this appear correctly inside large text, such as a title?

helenemccarron: That's a good point, the original uses ems, I will take it back to the drawing board and test it some more

Issue #1360, opened 2019-09-06
library wide: links/url styling


lwtchu:
I don't think there's a card that capture this. 

Color: #027eb5
Treatment: bottom border

In a shaded note (only Business Series!): #026aa1 with bottom border

MartaSztuka:
Link's colour changed.
![Screenshot from 2019-09-20 16-18-51](https://user-images.githubusercontent.com/47241114/65334281-6966e480-dbc2-11e9-8d5c-d5beb95e1fdd.png)

PR #1388, opened 2019-09-13
Style regular links

wilkus27:
Issue https://github.com/openstax/cnx-recipes/issues/1360


wilkus27:
I created `link-components.scss` and `link-shapes.scss` in `carnival>parts` directory. In issue there is also info that in shaded notes in Business series color of links should be different:

image

I'm wondering if I should style this now. It shouldn't be put to `carnival` directory, so maybe I should style links in `common` directory and there style two types of links - links in shaded notes and links in every other place? 

Other idea is to not style this now, but when style for Business series will be created. Then link could be styled as a component of shaded note. 

@helenemccarron @tomjw64 @brittany-johnson Please, let me know what do you think about it.

Issue #1367, opened 2019-09-09
Clean up Learning Objectives


helenemccarron:
- [x] Rename: `$LearningObjectiveTitleContainer: ( _name: "Title",...`
- [x] Rename `$LearningWrapperComponent: ( _name: 'Container',...`
- [x] Rename all components to learning-objective (rather than just learning)

MartaSztuka:
No changes detected in pdf.

PR #1391, opened 2019-09-13
Clean up Learning Objectives

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1367


styles/templates/template1/parts/_learning-settings.scss
```diff
@@ -3,21 +3,15 @@
...
             font-family: (_ref: 'typography:::sansFont'),
-            font-style: italic,
-            font-size: font-scale(1),
-            line-height: line-height(1)
         ),
-        Container: (
+        List: (
             margin-bottom:v-grid(2),
             margin-left: h-grid(2),
             padding-left: 0,
```
wilkus27: Should values of margins and paddings be harcoded? Now they are set to `OPTIONAL`.

tomjw64: yes, they should be hardcoded.

Issue #1396, opened 2019-09-17
anatomy: add style for module headers


wilkus27:
Module headers in Anatomy should be styled with specs from https://github.com/openstax/cnx-recipes/issues/1286

Stackblocks:
Displaying as specified with updated version of content / code.

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/18f0692c-35ef-4656-8222-f1ecdcdc1643)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/0e760db2-929e-4ca2-8979-d76f5def8afe)

PR #1397, opened 2019-09-17
Add style to module headers in anatomy

wilkus27:
Issue https://github.com/openstax/cnx-recipes/issues/1396


wilkus27:
In this case I move settings to `book.scc` files. 

@tomjw64 I was also trying to use `create_subshape` mixin, as you advised me, but this solution is good when we want to change the explicit value. It doesn't work when value is a map.

image (10)

tomjw64:
@wilkus27 was your `use` statement before your settings? It's hard to tell what went wrong without seeing the code

wilkus27:
@tomjw64 I included `create_subshape` once in `module-settings` file, another time in `module-shapes` file. In each case in output file I get implicit value `color: (_ref: "colorMap:::anatomyHeaderColor"),`. I will try to push such changes in another PR. It will be easier to find cause of this.

wilkus27:
I try this here: https://github.com/openstax/cnx-recipes/pull/1399/commits/795336381f3dc1d25d7763291ad236ce1b575759

PR #1402, opened 2019-09-18
Create files with book specific colors

wilkus27:
Some parts of book have the same structure and almost the same settings in each of the book  from Carnival Theme. The only one difference is in colors (e.g color of headers). 

A good example of this is Key Terms. To avoid repeating code and creating separate settings for each book I created files with color specific for book.


styles/books/biology/_colors.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../templates/template1/colors';
+
+@include add_settings ((
```
helenemccarron: I wonder if we need a separate file? Is this going to be a very long list, do you think? If so, it makes sense. I like the extra setting, since it's local to the book. Thoughts, @brittany-johnson @tomjw64 ?

wilkus27: Well base on palette color, each book has more or less 5 colors specific for itself. It depends on how we want to define colors for elements. We can name each of specific color in some way so then there will be 5 elements in list. But if we want to define colors with name of elements, as currently, it could be a long list. For example, now colors for headers, Chapter Oultine, Toc, Learning Obejctives are defined separately even all of those elements use the same color.

wilkus27:
**Decision:** keep color settings in one place as previously, all specific settings move to book file.

Thus, I'm closing this PR.

PR #1403, opened 2019-09-18
Add biology unit toc

tomjw64:


styles/designs/carnival/parts/_toc-components.scss
```diff
@@ -204,4 +204,42 @@ $TocContainer: (
...
+  )
+);
+
+$TocUnitLink: (
+  _name: 'TocUnitLink',
+  _subselector: ' > a',
+  _properties: (
+    display: block,
+    font-family: enum('ValueSet:::REQUIRED'),
+    font-size: enum('ValueSet:::REQUIRED'),
```
helenemccarron: Are the font size, line height and font weight going to change per book? Should they be hardcoded?

PR #1410, opened 2019-09-24
Add ap-everyday boxed note

tomjw64:


brittweinstein:
@tomjw64 I think we can go ahead and merge?

PR #1411, opened 2019-09-24
Icon note tweaks, add apbio note

tomjw64:


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -88,7 +89,7 @@ $IconNoteMainTitleNumberComponent: (
   _properties: (
     display: inline-block,
     word-wrap: normal,
-    max-height: 2rem,
+    // max-height: 2rem,
```
helenemccarron: can we just delete this?

Issue #1394, opened 2019-09-16
anatomy: add style for ToC with Units


wilkus27:
ToC in Anatomy should be styled with specs for ToC with Units from https://github.com/openstax/cnx-recipes/issues/1070

PR #1421, opened 2019-09-26
Add style to Toc in Anatomy

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1394


styles/books/anatomy/book.scss
```diff
@@ -2,6 +2,115 @@
 @import 'templates/template1/design';
 @import 'templates/template1/settings';
 
+@include add_settings((
```
helenemccarron: Would it make sense to have most of these in a toc-with-units-settings file of some sort and only define the book specific color settings at the book level? This seems like a huge chunk of potential copy and paste for all the books.

styles/templates/template1/_settings.scss
```diff
@@ -15,7 +15,6 @@
 @import 'templates/template1/parts/references-settings';
 @import 'templates/template1/parts/answerkey-settings';
 @import 'templates/template1/parts/chapterintro-settings';
-@import 'templates/template1/parts/toc-settings';
```
helenemccarron: See my comment above, I am a bit worried about move this to the book level.

wilkus27:
@helenemccarron requested changes are ready for review :slightly_smiling_face:

Issue #1427, opened 2019-10-01
Anatomy: remove Unit titles from book


wilkus27:
Unit titles which appear in the text should be removed:
![image](https://user-images.githubusercontent.com/40228854/65970362-b1112a00-e466-11e9-9cc8-2a2d0ba8ceae.png)

MartaSztuka:
Unit title is removed from the text.
![Screenshot from 2019-10-10 13-18-20](https://user-images.githubusercontent.com/47241114/66564658-d259d080-eb60-11e9-880d-9775b9100309.png)

PR #1436, opened 2019-10-04
Remove unit title

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1427


wilkus27:
@helenemccarron @tomjw64 I was trying to trash unit title by recipes, in different passes, but each time I got such error:

image (11)

so I decided to use `display: none` property in styles. Please let me know what do you think about such solution.

styles/designs/carnival/parts/_unit-components.scss
```diff
@@ -0,0 +1,15 @@
...
+    _properties: (
+
+    )
+);
+
+$UnitHeader: (
+    _name: "Header",
+    _subselector: " > h1",
+    _properties: (
+        display: none,
```
helenemccarron: I am bit on the fence on this one but I wonder if we should delete this kind of title in the recipes or hide it. This should not stop moving forward in this case, but I thought I'd at least ask the question for future reference. cc @brittany-johnson and @tomjw64

styles/books/anatomy/book.scss
```diff
@@ -94,6 +94,7 @@
 
 @include use('BookRoot', "common:::BookRoot");
 @include use('Preface','carnival:::UnitPreface');
+@include use('Unit', 'carnival:::UnitShape');
```
tomjw64: Since all that is happening is hiding, consider using `common:::Hidden`

tomjw64:
@wilkus27 Your error in easybake seems strange. Do you have a branch?

wilkus27:
@tomjw64 I use `common:::Hidden` as you requested. It's ready for review. Also I created a PR with trashing unit titles by recipes. Here is the link https://github.com/openstax/cnx-recipes/pull/1442

PR #1442, opened 2019-10-09
Trash unit title

wilkus27:
This is a PR with other way of removing unit title from the books - by recipes.
@tomjw64 as I wrote in https://github.com/openstax/cnx-recipes/pull/1436 such solution cause problems:

image

But PDF is generated properly.

(I removed in this PR release flag `use-title-nodes` from Anatomy because it removes unit and chapter titles from TOC. I wanted to be sure that this change is not caused by trashing unit title from the book)


tomjw64:
I'm guessing this error may be caused because deletion of the Unit header removes it from the Table of Contents? If this is the case, then hiding with styles may very well be the correct way to go.

tomjw64:
I've looked through cnx-epub and it looks like it's looking for the unit document-title, which mean that the content should have the document title, which also means that hiding it via styles looks to be the better way of doing this. I would be comfortable pushing that change through. Does that work for you Natalia? Feel free to close this issue if you agree

wilkus27:
Ok :+1: Thank you for checking that :relaxed: So I'm going to close this PR.

Issue #1438, opened 2019-10-08
Carnival Theme: Adding Color bands to EOB


ajluk:
EOB section should have added Color Bands regarding the specification placed in https://github.com/openstax/cnx-recipes/issues/1074

MartaSztuka:
Colour bands added.
![Screenshot from 2019-11-06 13-14-38](https://user-images.githubusercontent.com/47241114/68298354-6bfd9a80-0099-11ea-925a-4d4ed6905077.png)

PR #1443, opened 2019-10-09
Add color bands to EOB Appendix in chemistry 

ajluk:
Issue: #1438


styles/books/chemistry/book.scss
```diff
@@ -167,6 +186,83 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
   }
 }
 
+@mixin setAppendixPageBorder($type, $side) {
```
tomjw64: Since this mixin is almost exactly the same as above, why not change the above to accept an interpolated page type?

styles/books/chemistry/book.scss
```diff
@@ -167,6 +186,83 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
...
+        @content;
+      }
+      @right-top {
+        @content;
+      }
+    }
+  }
+}
+
+@mixin setEobPageBorder($type, $side) {
```
tomjw64: Since this mixin is almost exactly the same as above, why not change the above to accept an interpolated page type?

styles/books/chemistry/book.scss
```diff
@@ -148,46 +147,8 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
...
-        @content;
-      }
-    }
-  }
-}
-
-@mixin setAppendixPageBorder($type, $side) {
-  @page appendix:#{$type} {
+@mixin setPageBorder($type, $side) {
+  @page #{$page-type}:#{$type} {
```
ajluk: Mixin refactored to use interpolated page type. Is it done correctly?

styles/books/chemistry/book.scss
ajluk: I've apply this and above changes in the second commit and are ready for review.

styles/books/chemistry/book.scss
tomjw64: This does not look like it will work. Did you check the PDF? Keep in mind SCSS variables are not available during the "runtime" of the stylesheet.

What I expected is that you'd have one additional argument to `setPageBorder` with `$page-type` rather than trying to set it on the fly.

styles/books/chemistry/book.scss
```diff
@@ -129,8 +147,8 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
   margin-#{$side}: -1 * $fullBleedWidth;
 }
 
-@mixin setEocPageBorder($type, $side) {
-  @page eoc:#{$type} {
+@mixin setPageBorder($paget-type, $type, $side) {
```
```suggestion
@mixin setPageBorder($page-type, $type, $side) {
```

Was this working? When I switch branches, I get a compilation error when I try to build the styles myself because of this typo.

styles/books/chemistry/book.scss
```diff
@@ -167,19 +185,19 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
   }
 }
 
-@include setEocPageBorder("left", "left") {
+@include setPageBorder("page", "left", "left") {
```
tomjw64: This will produce `@page page:left` but this will probably not select anything.

styles/books/chemistry/book.scss
ajluk: I fixed the typo.

styles/books/chemistry/book.scss
ajluk: I have refactored the code above and the changes are ready for review.

Issue #1446, opened 2019-10-10
anatomy: add style to figures


wilkus27:
Add style to figures with specs from https://github.com/openstax/cnx-recipes/issues/1081.

MartaSztuka:
Styled.
![Screenshot from 2019-10-24 13-34-45](https://user-images.githubusercontent.com/47241114/67481929-1c63ab80-f663-11e9-9388-76ea51b9045c.png)

PR #1441, opened 2019-10-09
Style Anatomy Introduction page

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1428

PR #1456, opened 2019-10-14
Style figures in Anatomy

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1446


wilkus27:
**Waiting** for https://github.com/openstax/cnx-recipes/pull/1441. Title of figure in Introduction page also should be styled.

styles/designs/carnival/parts/_figureimage-shapes.scss
```diff
@@ -8,7 +8,25 @@
...
+                        $FigureLabel
+                    )
+                )),
+                $PicCaption
+            ),
+        )),
+    ))
+);
+
+@include create_shape('carnival:::FigureWithTitle',(
```
wilkus27: I created new shape for figures with titles. But maybe it will be better to add component related with title to previous shape even if not all books have figures with titles. Please let me know what do you think.

brittany-johnson: Hmmm. I'm thinking of a situation where we have a book that has figures with and without titles. But, I don't know if that is a real possibility. If it is we can still just include both shapes. 

I don't really see any major advantages or disadvantages to either approach. So whatever you feel is best!

helenemccarron: I liked how you cleaned up the difference between a title and a label. Regarding the shapes, we had a quick discussion and came to the conclusion using the same shape would be better. The current default structure is as follow: 
`<caption-container> <number> <title> <caption></caption-container>`.
Everything in the `<caption-container>` is optional. You can have a number but no title or caption, no title but a caption etc. Per Larissa, there is no way to be sure that a book will never have titles or captions for all the images in it so allowing for all three element within one shape makes sense.

wilkus27:
@helenemccarron I created one shape for figures as requested. Changes are ready for review.

Issue #1341, opened 2019-08-30
microbiology: iconned features


lwtchu:
See #1088 for specifications

Title: Micro Connections
Class: `microbiology micro-connection`
Color: #0DC0DC
Icon: https://drive.google.com/open?id=1z3nm9jswlV-pNb5v7UOlFwXZIAY1SKbZ

Microbiology also has Link to Learning (`microbiology link-to-learning`) & Check Your Understanding (`microbiology check-your-understanding`) which should be the same as it is in the other carnival theme titles. 

@helenemccarron no answer needed now - but let us know when you're ready for us to remove the words microbiology from the class in the Link to Learning and Check Your Understanding features.

brittany-johnson:
<img width="1792" alt="Screen Shot 2019-09-25 at 1 00 33 PM" src="https://user-images.githubusercontent.com/24909007/65627222-828ddd80-df94-11e9-8785-ef9d9c336d83.png">

lwtchu:
TWO SMOL THINGS:
Can it be
CHECK YOUR
UNDERSTANDING 
the your with the check

and thent he list inside check your understanding, the bullet should be within the feature.

brittany-johnson:
<img width="741" alt="Screen Shot 2019-10-14 at 1 09 54 PM" src="https://user-images.githubusercontent.com/24909007/66773126-f7fa1900-ee83-11e9-95cd-585d6f40c64e.png">

Stackblocks:
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/1dc4c99e-b44f-477f-bf82-a8a1335206cb)

Getting the bullet within the feature and the "Check Your" grouped together over "Understanding", but the icon and color have changed for the feature. Was that intentional, @brittany-johnson ?

PR #1458, opened 2019-10-14
issue-1341 pt 2

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/1341


helenemccarron:
It looks like `display: flex` does the job but would it make more sense to make the `max-width` required instead and adjust the with of the element based on the text length?

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -194,7 +202,7 @@ $IconCondensedTitleTextComponent: (
   _subselector: ' > .os-title-label',
   _properties: (
     display: inline-block,
-    max-width: h-grid(10),
+    max-width: h-grid(15),
```
tomjw64: Does this not cause regressions in the other IconCondensed Notes? Double check Link to Learning and Check Your Understanding in Chem. I expected this to change to `max-width: enum('ValueSet:::REQUIRED')` so that we could tailor width as needed per note class.

helenemccarron: Hah, you beat me to it!

Issue #1447, opened 2019-10-10
Style Fake Titled Tables


wilkus27:
Style tables in Anatomy based on informations from https://github.com/openstax/cnx-recipes/issues/1080, https://github.com/openstax/cnx-recipes/issues/1241, https://github.com/openstax/cnx-recipes/issues/1174.

Edit: style Fake Titled Tables in Anatomy and Chemistry (for Chemistry some changes in recipes are needed to have a proper markup to style).

MartaSztuka:
Fake titled tables are styled.
Anatomy:
![Screenshot from 2020-01-09 14-51-13](https://user-images.githubusercontent.com/47241114/72073820-0031eb00-32f1-11ea-9730-60d2b2165337.png)
Chemistry 2e:
![Screenshot from 2020-01-09 15-33-18](https://user-images.githubusercontent.com/47241114/72076155-6587db00-32f5-11ea-9bcb-e86eacd51536.png)

PR #1463, opened 2019-10-16
Style table in Anatomy

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1447


wilkus27:
Here is how it looks now:

image

styles/books/anatomy/book.scss
```diff
@@ -134,6 +142,7 @@
 @include use('Module', 'carnival:::ModuleShape');
 @include use('LearningObjectiveAbstract', 'carnival:::LearningObjectiveAbstractShape');
 @include use('Term', 'carnival:::Term');
+@include use('TopTitledTableSecondType', 'carnival:::TopTitledTableSecondTypeShape');
```
helenemccarron: I am concerned this might be confusing. We clarified the table types in https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/1241. Can you please rename this accordingly? We might not need a new shape. Please let us know if you need any help with this.

wilkus27: @helenemccarron in Anatomy there is another type of table. I didn't find it in issue which you posted. It has class `top-titled` but title is not in the header row. It's in div with class `os-table-title` before `<table>` tag:

table

wilkus27:
@helenemccarron I added changes in recipes for chemistry and created one type of top titled tabels (fake titled) for Anatomy and Chemistry. There are two problems at this moment:

1. After added changes in recipes for Chemistry occurs errors in baking process:

image

I need to spend more time to investigate this. But PDF is generating properly.

2. On the screenshot in  https://github.com/openstax/cnx-recipes/pull/1463#issuecomment-543656767 table looks good but this one is very wide. I found smaller tables and for them styling doesn't look good:

image

What would be better? To align title to left or to centerized whole table?

wilkus27:
@helenemccarron I resolved problem with errors, but I don't really like it :smiley: 
It has two disadvantages. New `div` has attributes typical for tables and in table after moving title there is an empty `<tr>`.
image

I had an idea to move only `<sup>` first and then use `string-set` to copy title content, but it didn't work. I was trying to use `:deferred` and to make changes in two passes, but each of those solutions cause another problems. So I decided to apply this https://github.com/openstax/cnx-recipes/pull/1463/commits/ac3d49368f0ed6da60301ea51391262779442afe not perfect solution. Please let me know what do you think about it.

styles/books/chemistry/book.scss
```diff
@@ -74,11 +74,11 @@
     ),
 ));
 @include add_settings((
-    TopTitledTable: (
-    _selectors: (".os-table"),
-    'TableHeadTitle:::color': (_ref:"colorMap:::chemistryHeaderColor"),
-    'LastTableRow:::border-bottom-color': (_ref:"colorMap:::chemistryHeaderColor"),
-    ),
+    FakeHeaderTitle: (
```
brittany-johnson: In PR https://github.com/openstax/cnx-recipes/pull/1563/files I've already changed the name from `TopTitledTable` to `FakeTitleTable`.  Do you prefer the name `FakeHeaderTitle` over `FakeTitleTable`? cc: @helenemccarron

wilkus27: I think that `FakeTitleTable` sounds better :relaxed:  When I merge U.U to resolve conflicts I will leave the name created by you :)

recipes/mixins/_number.scss
```diff
@@ -199,9 +199,17 @@
...
           &.top-titled {
             //this will have to be rewritten after all the content has been updated to contain the title within the title instead of the hacky first header in the table.
-            thead tr:first-child {
-              string-set: TopTitleContent content();
-              move-to: trash;
+            @if release-flag-enabled(table-move-fake-title) {
+              thead tr:first-child th {
+                container: div;
+                class: "os-table-title";
+                move-to: tableTopTitle;
```
helenemccarron: Why use move in this case? Wouldn't using `node-set` instead of `string-set` do the same thing (copy all the node to the title then trash them)?

wilkus27: I was trying to use `node-set` but it changes the id e.g. from `auto_m68846_footnote-ref3` to `auto_m68846_footnote-ref3-copy`. `move-to` doesn't cause any changes in attributs. But, as I write here: https://github.com/openstax/cnx-recipes/pull/1463#issuecomment-549815945 it's not a perfect solution.

wilkus27:
@helenemccarron as we talked, I added changes in other passes which clean up fake title. It's ready for review.

recipes/mixins/_number.scss
```diff
@@ -224,9 +232,17 @@
       @if ($hasTopTitle and release-flag-enabled(appendix-top-title-copy)) {
         &.top-titled {
           //this will have to be rewritten after all the content has been updated to contain the title within the title instead of the hacky first header in the table.
-          thead tr:first-child {
-            string-set: TopTitleContent content();
-            move-to: trash;
+          @if release-flag-enabled(table-move-fake-title) {
```
helenemccarron: I think it's not a good idea to nest a release-flag inside another one.

PR #1468, opened 2019-10-17
use-title-nodes with regression fix

tomjw64:


brittweinstein:
@tomjw64 we're good to merge!

Issue #1473, opened 2019-10-18
college physics: add style to Preface


wilkus27:
Apply style for Preface in College Physics with the same specs as in Anatomy (https://github.com/openstax/cnx-recipes/issues/1392).

Color pallete for this book: https://github.com/openstax/cnx-recipes/issues/1163

MartaSztuka:
Styled properly.
![Screenshot from 2019-10-22 12-59-56](https://user-images.githubusercontent.com/47241114/67279884-3a4ad800-f4cc-11e9-940f-ee2c57ee593e.png)

PR #1474, opened 2019-10-18
Style Preface in Physics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1473


styles/designs/carnival/parts/_preface-shapes.scss
```diff
@@ -12,7 +12,7 @@
     ))
 );
 
-@include create-shape('carnival:::UnitPreface',(
+@include create-shape('carnival:::PrefaceWithH5',(
```
helenemccarron: It seems to me that a more meaningful refactor of this shape would be to consolidate the two shapes into one with H5 in `Preface`. Or is there a reason for having both of them?

wilkus27: @helenemccarron This shape was created because in part of books there is no h5 header in Preface. But I also think it will be better to have just one shape for Preface even there is such difference between books. Changes are ready for review. 

I also moved back part of settings to `setting.scss` file and kept in `book.scss` only book specific settings (colors).

PR #1509, opened 2019-10-24
chemistry chapter page class

wilkus27:
Enable chapter-page-class flag for chemistry


wilkus27:
@helenemccarron I'm not sure what affect could have enabling this flag. It adds class to pages from chapters but I can't find any place where this class is used later for something.

PR #1513, opened 2019-10-24
Chemistry EOCsection-links

wilkus27:
Enable EOCsection-links flag in Chemistry


MartaSztuka:
EOC chapters are blue links.
Screenshot from 2019-10-25 11-49-49
Screenshot from 2019-10-25 11-53-25

Issue #1495, opened 2019-10-23
Enable all feature flags for Anatomy


helenemccarron:
All release flags should be enabled. Please list here which flags were enabled to they can be tested.

helenemccarron:
All flags were enabled except for the `give-composites-ids` flag, which is broken.

wilkus27:
And except `appendix-has-numbered-examples`. Some changes in mixin are needed to be able to enable this flag in Anatomy.

PR #1515, opened 2019-10-24
Enable appendix-has-numbered-examples in Anatomy

ajluk:
Issue: #1495


ajluk:
Enabling this flag generate errors while building ./recipes/output//anatomy.css (screenshot above). There is a need to make changes in the code after enabling this flag in Anatomy.
Screenshot from 2019-10-25 11-41-46

wilkus27:
Here is the condition for this flag:

image

Errors appear because in Anatomy in `$Config_PartType_Example` key `apTitleContent` is not defined.

So either we have to add it, even there are no appendix in Anatomy or modify in some way condition from screenshot above.

openstaxalina:
@wilkus27 could you create an issue + PR to resolve what you are describing above?  Please modify the mixin to allow for books without appendices.

wilkus27:
@openstaxalina Thanks for a respond. Here is the issue: https://github.com/openstax/cnx-recipes/issues/1539

recipes/output/accounting.css
```diff
@@ -1556,7 +1556,7 @@
 :pass(9) [data-type="chapter"], :pass(9) .appendix {
   counter-reset: example; }
 
-:pass(9) div[data-type="chapter"] [data-type="example"], :pass(9) .appendix [data-type="example"] {
+:pass(9) div[data-type="chapter"] [data-type="example"] {
```
wilkus27: Modyfications which I made are causing such changes in output of some books. I checked that those books don't even have `$Config_PartType_Example` in their config files, so I think it shouldn't make any regression. 
But if you think it would be better to avoid changes in outputs I will try to modify code in a different way.

There is one exception. In Biology `$Config_PartType_Example` is defined in that way:

image

Do you want me to do investigation about that? To find out what was the final decision and how book looks now (whether there are or not examples in Biology)? Maybe I should create issue about that? Let me know what do you think about it.

recipes/books/_generator.scss
```diff
@@ -187,15 +187,22 @@
     $solutionTitleContent:  map-get($Config_PartType_Example, solutionTitleContent);
     $resetAt: map-get($Config_PartType_Example, resetAt);
 
-    @if $resetAt == moduleReset {
-      @include count_countSectionExamples(exampleInSection);
-    } @else {
-      @include count_countExamples(example);
-    }
-    @if release-flag-enabled(appendix-has-numbered-examples) {
```
helenemccarron: It feels to me that it would be simpler to nest the conditional inside the release flag call:
```
    @if release-flag-enabled(appendix-has-numbered-examples) {
       @if map-has-key($Config_PartType_Example, apTitleContent) {
        $apTitleContent: map-get($Config_PartType_Example, apTitleContent);
        @include number_numberExamples($titleContent, $solutionTitleContent, $apTitleContent);
      } @else {
        @include _numberExamples($titleContent, $solutionTitleContent);
      }
    } @else {
      @include _numberExamples($titleContent, $solutionTitleContent);
    }
```
This would make it easier to remove the flag later.

recipes/mixins/_count.scss
```diff
@@ -111,20 +111,23 @@
...
-    '[data-type="chapter"], .appendix'
-    );
-  } @else {
-    @include _count(
-    $name,
-    'div[data-type="chapter"] [data-type="example"]',
-    '[data-type="chapter"], .appendix'
-    );
-  }
+@mixin count_countExamples($name) {
```
helenemccarron: I am not sure splitting the counter is absolutely necessary. With the approach I took above, I think I was able to reach the same results without changing any other files. I don't really like the usage we did before of ` @include _numberExamples($titleContent, $solutionTitleContent)` in `_generator`, though. Maybe we could rename ` _numberExamples`, which is supposed to be a mixin internal to `_number.scss` to something like `number_numberContentExamples` but that would be nitpicking :D

wilkus27: I think that changing the mixin name it's a good idea :relaxed: I did it in 
a4d636e. I modyfied also name of mixin `number_numberExamples` to `number_numberSpecifiedExamples` because it would be confusing what is the difference between `number_numberContentExamples` and `number_numberExamples`. Please let me know what do you think about it. Maybe you will come up with better name for this mixin :wink:

PR #1522, opened 2019-10-25
module-titled-notes in biol, chem and anatomy

wilkus27:
Enable module-titled-notes flag in Biology, Chemistry and Anatomy


wilkus27:
**NOTE** enabling this flag don't cause any changes in outputs after running `build-recipes` script

Issue #1528, opened 2019-10-28
Style unnumbered regular tables in AP Biology


ajluk:
Related issue: https://github.com/openstax/cnx-recipes/issues/1241

ajluk:
These tables have a "unnumbered no-title" class and look like this now: 
![image (1)](https://user-images.githubusercontent.com/55203598/67682269-3ec13600-f98f-11e9-98b0-2d3ca3862e8b.png)

ajluk:
These changes will be applied within this PR: #1566.

PR #1529, opened 2019-10-28
Style unnumbered default table apbiology

ajluk:
Issue: #1528 
The tables look like this now: 
Screenshot from 2019-10-28 15-35-25


styles/books/apbiology/book.scss
```diff
@@ -19,4 +19,5 @@
 
 @include use('APSciencePractices','carnival:::IconNoteShape');
 @include use('TeacherMaterial', 'common:::Hidden');
-@include use('APEverydayConnectionNote','carnival:::BoxedNoteShape');
\ No newline at end of file
+@include use('APEverydayConnectionNote','carnival:::BoxedNoteShape');
+@include use('BasicTable', 'carnival:::BasicTableShape');
```
wilkus27: Is it a reason that this use is included in ap-biology `book.scss` file and not in such file for biology? As I know biology `book.scss` file is imported to ap-bio so maybe this could be included there? Of course if basic tables occur in biology.

styles/output/apbiology-pdf.css
```diff
@@ -617,6 +617,34 @@ nav#toc > .os-toc-title {
...
+.os-table > table:not(.unstyled) > tbody > tr > td:first-of-type {
+  border-left: none;
+  border-right: 0.025rem solid black; }
+
+.os-table .os-caption-container > .os-title-label {
+  font-weight: bold; }
+
+.os-table .os-caption-container > .os-number {
+  font-weight: bold; }
+
```
wilkus27: In d3ea5fd you included use with Basic Tables to Biology, but in this case it should be removed from ap-bio `book.scss` file, because now there are two the same parts of code in ap-bio output files.

styles/books/apbiology/book.scss
ajluk: Yes, the basic tables occur in Biology which I wasn't sure at the moment. I have included the BasicTable in the biology/book.scss file as requested. Thanks!

ajluk:
These changes will be applied within this PR: https://github.com/openstax/cnx-recipes/pull/1566.

Issue #1530, opened 2019-10-28
Style Intoduction Page in Biology & AP Biology


ajluk:
The Chapter Outline should be styled in order to specs from this issue: https://github.com/openstax/cnx-recipes/issues/1076
The Splash image should be styled related to this issue: https://github.com/openstax/cnx-recipes/issues/1075

Stackblocks:
# Splash Images
The captions beneath splash images in Biology are unstyled and Figures are left-aligned instead of centered.

### Biology
![2019-11-07_13-08-10.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/ef4b7a98-6a70-4dec-8c47-7433d94efabd)

### AP Biology
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/d011c939-4d0d-4e31-993d-d8bc0a6d7a03)

# Image Alignment
![2019-11-07_13-08-22.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/45b7784b-7b2f-4b94-a5a8-a5a06d16ec46)

Stackblocks:
Displaying OK in Biology, but not in AP Biolgy

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/2c6a5b64-b3d9-4fb6-90e0-eea4f8600e73)o

helenemccarron:
For AP Biology, the feature flag needs to be enabled for this to work. This will be fixed when we enable all feature flags for v.v.

helenemccarron:
Please regression test the biology part then send back to v.v and backlog

MartaSztuka:
Displaying good i AP Biology
![Screenshot from 2019-11-26 17-07-50](https://user-images.githubusercontent.com/47241114/69650775-64fdf280-106f-11ea-9d7d-719d504097a3.png)

PR #1540, opened 2019-10-29
Style introduction page in Biology & AP Biology

ajluk:
Issue: #1530 

I have added the styles to Biology, because the color palette is the same for those books. I have also removed the Unit Titles according to this issue: https://github.com/openstax/cnx-recipes/issues/1427.


ajluk:
Within this PR I also added Chapter intro book specific color settings, because they weren't applied and the color palette for Biology and APBiology is the same.

Issue #1543, opened 2019-10-30
apbiology & biology: add style to links


ajluk:
Style links in Biology and AP Biology with specs from #1360

Stackblocks:
The links are Blue #0000FF instead of #027eb5. I've changed the "view content" color in the screenshot to #027eb5 for comparison.

![2019-11-07_13-21-54.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/24055077-ee42-4b18-a240-814b24fd5fc2)

Stackblocks:
Link color is #017DB4 instead of #027eb5, but when I change the hex value for part of the word "figure" there is little to no visible difference so it may be an artifact of conversion.

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/ea601000-e2b8-416c-8c5e-5dc890756b84)

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/8966a756-72ec-4555-943a-842dc946e0c3)

PR #1544, opened 2019-10-30
Add style to links in Apbiology & Biology

ajluk:
Issue: #1543


styles/output/apbiology-pdf.css
```diff
@@ -641,6 +641,10 @@ nav#toc > .os-toc-title {
 .os-index-container > .group-by > .os-index-item > .os-term-section-link {
   content: target-counter(attr(href, url), page); }
 
+<<<<<<< HEAD
```
brittany-johnson: You may need to recompile so that `HEAD` is removed.

ajluk: Thanks for catching it up. I have added requested changes.

PR #1545, opened 2019-10-30
Add with-math or no-math to the build-pdf filename

karenc:
Instead of having just `collection.pdf`, we now have either
  `collection.with-math.pdf` or `collection.no-math.pdf`.  If
  `build-pdf` used the baked html, it will output a
  `collection.no-math.pdf` and if it used the mathified html, it
  will output a `collection.with-math.pdf`.  This should make things
  clearer.

### DIFF
--- a/script/build-pdf
+++ b/script/build-pdf
@@ -40,16 +40,17 @@ fi
 for book_config in "${BOOK_CONFIGS[@]}"; do
   read -r book_name _ _ _ _ _ _ <<< "${book_config}"
 
-  book_dir="./data/${book_name}"
+  book_dir="data/${book_name}"
   mathified_file="${book_dir}/collection.mathified.xhtml"
   baked_file="${book_dir}/collection.baked.xhtml"
   style_file="${STYLE:-./styles/output/${book_name}-pdf.css}"
-  output_file="${OUTPUT_DIR:-.}/${book_dir}/collection.pdf"
 
   if [ -f "${mathified_file}" ]; then
     input_file="${OUTPUT_DIR:-.}/${mathified_file}"
+    output_file="${OUTPUT_DIR:-.}/${book_dir}/collection.with-math.pdf"
   else
     input_file="${OUTPUT_DIR:-.}/${baked_file}"
+    output_file="${OUTPUT_DIR:-.}/${book_dir}/collection.no-math.pdf"
   fi
 
   style_flag=()


karenc:
@helenemccarron I am bad with naming, so... I don't really have an opinion on this :laughing:  If you're just talking about the filename in this PR, I can certainly change it right now, something like `collection.with-math.pdf` and `collection.without-math.pdf`?

helenemccarron:
I totally get it, I enlisted people smarter than me to come up with my suggestions :D I don't know if the renaming would be limited to those files or include some of the commands etc. but yes, let's at least get started with the file names. Let me know if I should open an issue to change other places that might be using "mathify".

karenc:
@helenemccarron I changed the filenames :)

script/build-pdf
```diff
@@ -40,21 +40,24 @@ fi
...
+  rm -f "${output_file}"
 
   if [ -f "${mathified_file}" ]; then
-    input_file="${mathified_file}"
+    input_file="${OUTPUT_DIR:-.}/${mathified_file}"
+    output_file="${OUTPUT_DIR:-.}/${output_file/.pdf/.with-math.pdf}"
   else
-    input_file="${baked_file}"
+    input_file="${OUTPUT_DIR:-.}/${baked_file}"
+    output_file="${OUTPUT_DIR:-.}/${output_file/.pdf/.without-math.pdf}"
```
tomjw64: In both branches the `.pdf` suffix is removed. Would it make sense to not put it there in the first place?

tomjw64: In both branches the `.pdf` suffix is removed. Would it make sense to not put it there in the first place?

karenc:
@philschatz @tomjw64 @helenemccarron I separated out the bug fixing part to #1564 so that can be merged first while we talk about this new feature of changing filename or adding a command line option.

pumazi:
associated with openstax/cnx#749 (zenhub won't make the connection)

script/build-pdf
```diff
@@ -40,21 +40,22 @@ fi
 for book_config in "${BOOK_CONFIGS[@]}"; do
   read -r book_name _ _ _ _ _ _ <<< "${book_config}"
 
-  book_dir="${OUTPUT_DIR:-./}/data/${book_name}"
+  book_dir="data/${book_name}"
```
pumazi: Why is the data not considered output? I'm just curious why other things transitioned to being output while the data transitioned to being a fixed location.

script/build-pdf
```diff
@@ -66,7 +67,7 @@ for book_config in "${BOOK_CONFIGS[@]}"; do
   do_progress "Generating pdf using princexml" \
     docker-compose run --rm -u root pdf prince -v --output="${output_file}" "${style_flag[@]}" "${input_file}"
 
-  docker-compose run --rm -u root pdf chown -R "$(stat -c '%u:%g' ./data)" "${book_dir}"
+  docker-compose run --rm -u root pdf chown -R "$(stat -c '%u:%g' ./data)" "${OUTPUT_DIR:-.}/${book_dir}"
```
pumazi: Would it be good to set the default in one location? Then you need only reference `${OUTPUT_DIR}` in all of these statements. That would mean you'd be able to change the default in one place, no?

pumazi: Karen mentioned possibly putting this kind of thing in a separate PR.

PR #1561, opened 2019-11-04
add style to lists in preface biology

ajluk:
Issue: #1562


styles/designs/carnival/parts/_preface-components.scss
```diff
@@ -62,7 +62,7 @@ $PrefaceH5: (
 );
 $PrefaceUnorderedList: (
     _name: "PrefaceUnorderedList",
-    _subselector: " > section > section > ul",
+    _subselector: " section > ul",
```
ajluk: I have changed the subselector in this component to make it cover styles for all the ul in Preface.

Issue #1565, opened 2019-11-05
college physics: style Introduction page


wilkus27:
Add style to Introduction page in College Physics with specs from Carnival Theme.

wilkus27:
**Theme Design Specs**

1. New shape: `Chapter Intro With Abstract`
Differences to `Chapter Intro` shape:
- new component `OutlineObjectiveLearning` in `IntroBody` with property: `margin-left: 1.5rem,`
2. Physics specific specs
- `Outline` - `background-color: B5`
- `SplashFigureBg` - `background-color: B5`
3. New color in Color Map
`physicsIntroColor: (_ref: "scheme:::B5"),`

wilkus27:
NOTE for tester: https://github.com/openstax/cnx-recipes/pull/1587#issuecomment-556054254

MartaSztuka:
Introduction page styled.
![Screenshot from 2019-11-22 16-36-16](https://user-images.githubusercontent.com/47241114/69439040-75982b00-0d46-11ea-9abd-65eb6d63ef2f.png)

PR #1587, opened 2019-11-13
Style Introduction page in Physics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1565


wilkus27:
image

This is how it looks now. After discussion, decision is to leave this as-is and it will be created separate issue about cutted Chapter Outline

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -80,6 +80,14 @@ $ChapterOutlineObjectiveText: (
     )
 );
 
+$ChapterOutlineObjectiveLearning: (
+    _name: "OutlineObjectiveLearning",
+    _subselector: ' > .learning-objective',
+    _properties: (
+        margin-left: 1.5rem,
```
ajluk: Will it be a better solution to use  h-grid($value) function to set the margin-left property value in the component above?

wilkus27: I check and all margins and paddings in this case (Introduction page) are set without using v-grid and h-grid functions:

image

Because Introduction page is going to be changed maybe we can leave this as-is now and when there will be decision how to style it, then we can try to apply those functions to each component. What do you think?

ajluk: Ok. So, I think it can be left as it is for now and we can consider this solution when we got the decision about the Introduction Page design.

Issue #1620, opened 2019-11-20
Biology for AP Courses: add styles to exercises in EoC


ajluk:
The exercises in the EOC:

- [x] - Review Questions
- [x] - Critical thinking questions
- [x] - Test Prep for AP® Courses
- [x] - Science Practice Challenge Questions

 should be styled according to: https://docs.google.com/spreadsheets/d/1piXDN2me_xyD2CtqODOk9irW69z4JflMwpOf0KV-Lak/edit#gid=2076412410

Changes in Theme Design Specs:
New shapes in the **EocExerciseWithoutSubtitle** shape component:

- **Review**
- **Critical thinking**
- **TestPrep**
- **SciencePractice**

They all have have a common components and properties like so:
Container: (
            column-count: 2,
            column-gap: 2.4rem,
            column-width: auto,
        ),
EocHeader:(
            font-family: (_ref: 'typography:::sansFont'),
            font-weight: bold,
            font-size: font-scale(4), 
            line-height: line-height(4),
            text-transform: uppercase,
            column-span: all,
             // color: book specific, defined in book.scss file
        ),
ProblemNumber: (
            color: (_ref: 'colorMap:::fontBodyColor'),
            font-weight: bold,
            text-decoration: none,
        ),
ProblemDivider: (
            margin-right: h-grid(1)
        ),
    ), 

All EoC components  have a **book specific color** for APBiology set to **B7**. 

Info for the testers:

The issue regarding images form the TEST PREP FOR AP® COURSES exercices will be created and the changes will be added within another PR (#1634): 
![image](https://user-images.githubusercontent.com/55203598/69257841-01318080-0bbc-11ea-80e4-780655f3fc4c.png)

MartaSztuka:
Space between 1, 2, 3, and abc answers is too big.
1.
![Screenshot from 2019-11-27 11-20-26](https://user-images.githubusercontent.com/47241114/69858495-6a875280-1292-11ea-948e-60ff7a7c546b.png)
2. ![Screenshot from 2019-11-27 10-50-34](https://user-images.githubusercontent.com/47241114/69858436-49befd00-1292-11ea-8154-b65688a28a7f.png)
3. ![Screenshot from 2019-11-27 11-13-14](https://user-images.githubusercontent.com/47241114/69858441-4b88c080-1292-11ea-8c0b-94faedf4a592.png)
[...]

PR #1621, opened 2019-11-20
Style eoc exercises apbiology in Biology for AP Courses

ajluk:
Issue: #1620


styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -234,4 +234,79 @@
...
+            column-width: auto,
+        ),
+        EocHeader:(
+            font-family: (_ref: 'typography:::sansFont'),
+            font-weight: bold,
+            font-size: font-scale(4), 
+            line-height: line-height(4),
+            text-transform: uppercase,
+            column-span: all,
+            color: (_ref: 'colorMap:::chemEOCHeaderColor'),
```
brittany-johnson: I see that the `EocHeader` color is being defined here are in the `book.scss` file. I think the color definitions can be removed in this file for each `EocHeader`. What do you think?

ajluk: I have left it, because I relied on the components created before and they all have it preserved. Do you think I should removed them from every component entirely or replace them with REQUIRED value?

wilkus27: For `Module` I did something like that:

image

What do you think to apply this also here? Then developer who is working on other book has an information that color has to be defined in a book.scss file.

ajluk: Thank Natalia, the solution seems great.

ajluk: Changes added and code is ready for review.

Issue #1579, opened 2019-11-08
Update books.txt file


wilkus27:
Some of books have second edition, but informations in `books.txt` are outdated. 

We need to go through this file and verify book name, UUID and collection number and change them if it is needed.

![image](https://user-images.githubusercontent.com/40228854/68488740-04d11900-0246-11ea-8f60-3620b1cba0bc.png)

brittweinstein:
@openstaxalina to pull the data and analyze for accuracy, suggest changes and then a Dev will update with accurate data.

helenemccarron:
current data stored in books.txt:
```
  "calculus-vol1              calculus                           8b89d172-2927-466f-8661-01abc7ccdba4   col11964   katalyst01.cnx.org  mixins/"
  "calculus-vol2              calculus                           1d39a348-071f-4537-85b6-c98912458c3c   col11965   katalyst01.cnx.org  mixins/"
  "calculus-vol3              calculus                           a31cd793-2162-4e9e-acb5-6e6bbd76a5fa   col11966   katalyst01.cnx.org  mixins/"
  "precalculus                precalculus                        fd53eae1-fa23-47c7-bb1b-972349835c3c   col11667   easyvm5.cnx.org    mixins/"
  "college-algebra            precalculus                        9b08c294-057f-4201-9f48-5d6ad992740d   col11759   easyvm5.cnx.org    mixins/"
  "algebra-trigonometry       precalculus                        13ac107a-f15f-49d2-97e8-60ab2e3b519c   col11758   easyvm5.cnx.org    mixins/"
  "microbiology               microbiology                       e42bd376-624b-4c0f-972f-e0c57998e765   col12087   easyvm5.cnx.org    mixins/  template1"
  "anatomy                    anatomy                            14fb4ad7-39a1-4eee-ab6e-3ef2482e3e22   col11496   easyvm5.cnx.org    mixins/  template1"
  "history                    history                            a7ba2fb8-8925-4987-b182-5f4429d48daa   col11740   easyvm5.cnx.org    mixins/"
  "aphistory                  ap-history                         0a621f27-abe1-4c17-8f1c-d80d07958977   col10034   content05.cnx.org  mixins/"
  "psychology                 psychology                         4abf04bf-93a0-45c3-9cbc-2cefd46e68cc   col11629   easyvm5.cnx.org    mixins/"
[...]

PR #1637, opened 2019-11-22
update scripts, update books.txt, rename ap-bio

brittany-johnson:
For issue: 
https://github.com/openstax/cnx-recipes/issues/1579
(name of branch doesn't match issue name, ignore : / ) 

- book and recipes names, uuids and collections have been updated 
- using recicpe_name instead of book_name to generate style files because books that share the same recipe will share the same style so they will have the same name, but the book_name will not always match the name of the style


script/build-styles
```diff
@@ -20,12 +20,11 @@ if [[ ${#TEMPLATE_NAMES[@]} == 0 ]]; then
...
 for style_config in "${TEMPLATE_NAMES[@]}"; do
-  read -r _ book_name <<< "${style_config}"
-
-  sass_file="${styles_root}/${book_name}/book.scss"
+  read -r _ _ recipe_name <<< "${style_config}"
 
+  sass_file="${styles_root}/${recipe_name}/book.scss"
   for platform in "${PLATFORMS[@]}"; do
-    css_file="${output_dir}/${book_name}-${platform}.css"
+    css_file="${output_dir}/${recipe_name}-${platform}.css"
```
tomjw64: What's the benefit in using recipe name for the style file? Is the plan to have a one-to-one relationship between recipes and styles? That is, if I build a style for statistics and a style for business-statistics, I would want them to be named the same (or be the same?), since it seems they use the same recipe?

brittany-johnson: Yes! I confirmed with @lwtchu that the plan is that books that share recipes will also share styles. So the recipe and style name should be the same

tomjw64: Neato. Might be worth using a different name than recipe for this, since it might get a bit confusing. I believe the term used in legacy to describe a recipe/style (kinda) is 'print-style'. Would we want to stick with that, or stick with using 'recipe', or come up with a different term?

c0ldcalamari:
- [ ] tested with pdf-output producer.

How:
1. Denoted specific branch for `cnx-recipes` resource
```
  - name: cnx-recipes
    type: git
    source:
      uri: https://github.com/openstax/cnx-recipes.git
      branch: issue-1624
```
2. Updated only reference in pipeline tasks using cnx-recipes, `t-book-lookup.yml`
```
...
      cp workflow-events-queued/version         book-title/version
      cp workflow-events-queued/content_server  book-title/server
      wget -q https://raw.githubusercontent.com/openstax/cnx-recipes/issue-1624/books.txt
      set +x
      . books.txt
      set -x
...
```

c0ldcalamari:
👍 LGTM

Issue #1588, opened 2019-11-14
college physics: style notes


wilkus27:
All notes in College Physics should be styled with specs from Carnival Theme.

wilkus27:
Ready for testing. Specs are added to Theme Design Specs table.

MartaSztuka:
![Screenshot from 2020-01-13 13-25-19](https://user-images.githubusercontent.com/47241114/72256100-b440b800-3608-11ea-9367-549f54c61be0.png)
![Screenshot from 2020-01-13 13-24-24](https://user-images.githubusercontent.com/47241114/72256102-b4d94e80-3608-11ea-87d3-9ed40dc7fbb2.png)

PR #1643, opened 2019-11-26
Physics notes

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1588


recipes/mixins/_modify.scss
```diff
@@ -428,14 +428,28 @@
   @if $depth > 0 {
     $minorLevel: 'depth#{$depth}';
   }
-  @if $replaceHeader {
```
wilkus27: This condition is too general. In case when in `Config_Notes` we are using key `useHeader` we don't need `replaceHeader`, but if there is no `replaceHeader` then existing header is moved to note body to tag with class `os-subtitle`. What is causing problems when we want to use existing header as the main title, not subtitle.

Here is the screen from University Physics:
image

https://katalyst01.cnx.org/contents/1Q9uMg_a@13.51:mJu6YvwH@17/3-5-Free-Fall

For now, I just change conditions to when `repalceHeader` is set to true or to false. So `useHeader` is working properly now, but I haven't tested this flag for books which don't use `useHeader` and don't use `replaceHeader`. I want to create separate issue for this and modify this release flag to fit to all cases. Let me know if it's ok for you to merge it for now in this form and work on it more in other PR.

styles/books/physics/book.scss
```diff
@@ -54,6 +96,11 @@
 @include use('FigureContainer','carnival:::Figure');
 @include use('Link', "carnival:::Link");
 @include use('Term', 'carnival:::Term');
+@include use('DefaultNoClassNote', 'carnival:::BoxedNoteShape');
+@include use('Example', 'carnival:::IconNoteShape');
+@include use('CheckUnderstandingExerciseNote', 'carnival:::CheckUnderstandingShape');
+@include use('PhetExplorationNote', 'carnival:::IconLogoNoteShape');
+@include use('IFrame', 'common:::Hidden');
```
wilkus27: @helenemccarron we have had a discussion about iframes. We want to hide them in PDF but not on webview. I've found that there is a shape `Hidden` and it's created in pdf folder:
image

Is it ok for you that I use this and don't create separate shape for iframe? Let me know what do you think.

styles/designs/carnival/parts/notes/_notes-checkUnderstanding-shapes.scss
```diff
@@ -0,0 +1,31 @@
...
+        map-merge($IconCondensedTitleComponent, (
+          _components: (
+            $IconCondensedTitleTextComponent,
+            $IconCondensedTitleIconComponent,
+          ),
+        ))
+      )
+    ))
+  ),
+));
```
wilkus27: I created separate shape for Check Your Understanding from Physics. As I explain in comment (lines 3-6), there are two types of this element. 

This shape is based on Icon Condensed shape, so I used some Icon Condensed components. But maybe it will be better to create separate components with the same properties but different names started by `CheckUnderstanding`. Let me know what do you think.

My other concern is, maybe if this shape is based on  Icon Condensed shape it should be put to `_iconCondensed-shapes.scss` file. As you can see, I created separate file for this shape. 

I'm curious what do you think about it.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -23,6 +23,23 @@ $NoteUnorderedList: (
...
+    width: 100%,
+    display: inline-block,
+    padding-right: enum('ValueSet:::OPTIONAL'),
+    position: relative,
+    margin-bottom: enum('ValueSet:::OPTIONAL'),
+    bottom: enum('ValueSet:::OPTIONAL'),
+    content: "' '",
+  )
+);
+
```
wilkus27: I created general note logo component. That's why I removed `$PartnerNoteIconComponent` (line 247 in this file). Partner Note is a Boxed Note with logo but PhET Exploration note which I'm creating in this PR is a Icon Note with logo. 

For Partner Note there is created new shape in separate file: (https://github.com/openstax/cnx-recipes/blob/master/styles/designs/carnival/parts/notes/_notes-partner-shapes.scss). 

For PhET Note I also created separate shape, but inside `icon-shapes.scss` file and I named this shape by `IconLogoNoteShape`. What do you think about such convention? If we want to go with that solution I can do the same with Partner Note, so move the shape to `boxed-shapes.scss` file and name it by `BoxedLogoNoteShape`. 
If you think it's better to keep shapes with logos in separate files I will create that for PhET Note.

Let me know your thoughts about it.

wilkus27:
@brittany-johnson and @helenemccarron I added you both to code review because I have a lots of questions about my solutions and I'm curious of your opinion :relaxed:

styles/books/physics/book.scss
brittany-johnson: I think this is fine!

styles/designs/carnival/parts/notes/_notes-components.scss
brittany-johnson: I think this is a really good idea. I didn't realize before that there would be different types of logo notes. 
If you can do the same with Partner note that would be great!

recipes/mixins/_modify.scss
brittany-johnson: I believe it would be fine to merge it for now and work on it more in another PR but @helenemccarron can you confirm?

styles/designs/carnival/parts/notes/_notes-checkUnderstanding-shapes.scss
helenemccarron: I really would like to avoid using feature specific shapes: that makes them difficult to reuse. For example, what if you want to use the exact same shape in another book but with a different note name? It will get confusing very quickly. 

More importantly, it appears that this Check Your Understanding is not really a note but an exercise styled to look like a note, is that correct? If that is the case, I think we need to create something like `boxedNoteExercise`. 

The next question I have is do we even want to keep it in notes? The answer is probably yes because it looks like one but I would make sure it is isolated (i.e. at the end of the file or something) and as you did, thoroughly commented. 
We should also make sure to document those in the specs, because this is actually adding another "note" type but to an exercise. @Stackblocks and @brittany-johnson we might want to discuss how we handle those in the framework in general.

styles/designs/carnival/parts/notes/_notes-components.scss
helenemccarron: In my mind, if I am trying to select a note, I would consider the following types: with a logo or without, then with an icon or without. Would it make sense to create a new note type and build the different shapes (with and without icons) using the existing components for that type? In this case, I would rename the partner component and shape to something more generic like `LogoNote` and group the shapes with and without icons. Does that make sense? We can discuss if needed.

wilkus27:
@helenemccarron I added all requested changes and removed, for now, changes in mixin. I will create a separate issue for that. PR is ready for review.

PR #1648, opened 2019-11-26
REVIEW: Add conversion function from image url to data uri

tomjw64:
Main thing here is that I added a function `toDataUri(type, path)` which is injected with node-sass for styles only and migrated all `url` calls  (with the exception of ones for fonts) to use the function. Function can be viewed in `styles/build/build.js`. Paths given to `toDataUri` are relative to the styles directory.


philschatz:
Is there a unit test that could run to verify that the `url(...)` in `.css` files never contain relative links since that is the new assumption?

tomjw64:
@philschatz Very good idea

brittweinstein:
@tomjw64 please do not put any more work into this. The TBD team will review this as it stands.

PR #1652, opened 2019-11-29
add styles to ToC u-physics

ajluk:
Issue: #1653


recipes/mixins/_toc.scss
```diff
@@ -109,6 +109,67 @@
   }
 }
 
+@mixin EocCompositeChapterTOC($hasCompositeChapter) {
```
wilkus27: I created new mixin, it's almost the same as mixin `compositeChapterTOC` (line 55). I change selector at the beginning, add classes to `li`, and `ol`, and I correct one of the string (there was a typo). Although those mixins are similar I create separate one (not change the old one) because it will be easier to delete release flag in the future (then we will just remove whole old mixin).

ajluk:
@helenemccarron and @brittany-johnson I have assigned you both to code review, because Natalia have made the main changes here and we would like to know what do you think about this solution. (Natalia's comment on these changes: https://github.com/openstax/cnx-recipes/pull/1652#discussion_r354251637 )

recipes/mixins/_toc.scss
brittany-johnson: Since the mixin above on line 55 `compositeChapterTOC` is still being implemented with a release flag, can we make the changes to the original mixin instead of including a new one? I locally tried this and it seemed to work? But let me know!

ajluk: @brittany-johnson I have modified the original mixin `compositeChapterTOC` as you suggested with some help from Natalia and it works properly now. The code is ready for your review.

recipes/mixins/_toc.scss
```diff
@@ -53,57 +53,118 @@
...
-        }
-        &::after {
-          content: pending(CpageLink);
-          attr-cnx-archive-uri: attr(id);
-          attr-cnx-archive-shortid: string(page-shortid);
-          container: li;
-          move-to: Cpage-list;
+  @if release-flag-enabled(toc-eoc-composite-chapter){
+    @if $hasCompositeChapter {
+      > [data-type="chapter"] {
```
helenemccarron: Not all composite chapters are within chapters (for example the answer key), are you sure this works for all books?

wilkus27: All settings for Answer Key are added in the next mixin (line 173) https://github.com/openstax/cnx-recipes/blob/dcbc9d18d3640d1b756f29d30328f96faf0227ad/recipes/mixins/_toc.scss 

There is a condition in which we add `[data-type="composite-chapter"]` selector:

image

And in lines 230 and 351 there are settings for this selector, so in output.css file (https://github.com/openstax/cnx-recipes/blob/dcbc9d18d3640d1b756f29d30328f96faf0227ad/recipes/output/u-physics.css) we can see that it covers `[data-type="chapter"]` and `[data-type="composite-chapter"]` (which is direct child of body like Answer Key)

image

ajluk: Yes and the situation takes place in University Physics. This book contains Answer Key with the selector `data-type="composite-chapter"` as a direct child of the body and the mixin is covering this as well.

helenemccarron: I see, thank you for clarifying!

PR #1654, opened 2019-11-29
Images in all EoC exercises need to be styled in Biology for Ap Courses

ajluk:
Issue: #1655


styles/output/apbiology-pdf.css
```diff
@@ -1683,6 +1683,9 @@ span[data-type="term"]:not(.no-emphasis) {
 [data-type='chapter'] > .os-review-container [data-type='problem'] .os-divider {
   margin-right: 8px; }
 
+[data-type='chapter'] > .os-review-container [data-type='problem'] .os-problem-container img {
+  max-width: 100%; }
+
```
wilkus27: Are conflicts resolved properly? `apbiology` directory was renamed to `ap-biology`. At this moment you have changes in  two files: `apbiology-pdf.css` and in `ap-biology-pdf.css`. The first one shouldn't be generated.

ajluk: After the conversation with Helene on the call I guess I should leave the files and they will be deleted in another PR.

PR #1697, opened 2019-12-16
apbiology: add space margin-bottom nested lists exercises 

ajluk:
Issue: #1620 https://github.com/openstax/cnx-recipes/issues/1620#issuecomment-559722360

This PR also contains applying margins to ordered and unordered lists in the following books: 

- Anatomy,
- Chemistry,
- Biology 2e,
- Biology for AP Courses,
- College Physics,
- High School Physics,

partly includes:

- University Physics,
- Microbiology

Elements that have added list styles:
- Preface,
- Module,
- EoC (Summary, Exercises),
- Answer Key,
- Appendices

Within this PR I have also added styles to images and tables in EoC Exercises for Biology 2e, Biology for AP Courses and College Physics.  

Documentation for testing purposes is available in this document: https://docs.google.com/document/d/1cry7GPpIZ00sCudMtYExkw3Y6vEH8NjfiVGUT8DgLCk/edit


styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -216,7 +216,7 @@
             width: 100%,
         ),
         ExerciseNestedList: (
-            transform: translateX(-20px),
+            transform: translateX(-2rem),
```
ajluk: I have changed px to rem in this property because of the Travis error, where it's been suggested to do such change.

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -277,7 +277,7 @@
...
```
ajluk: Same as above.

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -313,7 +313,7 @@
...
```
ajluk: Same as above.

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -349,7 +349,7 @@
...
```
ajluk: Same as above.

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -215,6 +215,10 @@
             table-layout: fixed,
             width: 100%,
         ),
+        ExerciseNestedList: (
+            transform: translateX(-2rem),
```
helenemccarron: Is this used to override a margin or a padding left? If so, it's better to set that element back to 0 rather than using `transform`. Another consideration, if the original setting is too broad would be to redesign it to apply separately to each component if possible rather than overriding it.

ajluk: I have try to set the padding to 0 (it's the padding I think generated by the browser, seen on the screen below) but the then the nesting was broken. I was trying to decrease the padding as well but then the space was getting even bigger in the pdf file, so I have used translate as a last resort. Here is how it looks before:
image

helenemccarron: Have you tried adding `padding: 0` in `_reset`? It seemed to work for me when I tested it there. I understand it will make a really broad change that we need to test (`ol` will have to be checked in all books)

ajluk: No I didn't. Thank you for the suggestion. I will add this solution within this PR.

ajluk: I have reset the paddings and margins for the ol lists in the _reset.scss file and set them in the EoC exercises (in the_eoc.settings.scss file), where they occur. @helenemccarron do you think it would be better to set the rest of the padding and margins for ol lists in the other parts of this book within another PR?

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -215,6 +215,18 @@
             table-layout: fixed,
             width: 100%,
         ),
+        ExerciseContainerList: (
+            padding-left: h-grid(4),
+            margin-bottom: v-grid(1),
+        ),
```
ajluk: I have created one more lists component (ExerciseContainerList), because I have found some lists nested in divs like in the screen below:
image

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -272,6 +284,18 @@
...
```
ajluk: Same as above

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -304,6 +328,18 @@
...
```
ajluk: Same as above

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -336,5 +372,18 @@
...
             width: 100%,
         ),
+        ExerciseList: (
+            padding-left: h-grid(4),
+            margin-bottom: v-grid(1),
+        ),
+        ExerciseContainerList: (
+            padding-left: h-grid(4),
+            margin-bottom: v-grid(1),
+        ),
```
ajluk: Same as above.

styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -134,4 +134,28 @@ $Exercise_Table: (
...
+    )
+);
+$Exercise_Container_List: (
+    _name: "ExerciseContainerList",
+    _subselector: " .os-problem-container > div > ol",
+    _properties:(
+        padding-left: enum('ValueSet:::OPTIONAL'),
+        margin-bottom: enum('ValueSet:::OPTIONAL'),
+    )
+);
```
ajluk: I have created one more lists component (ExerciseContainerList), because I have found some lists nested in divs like in the screen below:
image

styles/templates/template1/parts/_eoc-settings.scss
helenemccarron: Is the translate still necessary after resetting the padding in `_reset.scss`? When I tested this, I was able to use padding after making sure to rest it first.

styles/templates/template1/parts/_eoc-settings.scss
helenemccarron: No, I think in this PR is preferable so we can keep track of all the changes related to this.

Issue #1712, opened 2019-12-19
Add styles to figures in University Physics vol1


ajluk:
Figures should be styled according to Carnival Specs. Related issue: #1081.
No changes in Theme Design Specs are needed.

MartaSztuka:
Styles added.
![Screenshot from 2020-01-08 17-09-40](https://user-images.githubusercontent.com/47241114/71995631-2e54f380-323b-11ea-82f5-718138a55179.png)

PR #1713, opened 2019-12-19
u-physics-vol1: add styles to figures 

ajluk:
Issue: #1712


brittweinstein:
@ajluk we can go ahead and merge this without @Stackblocks review since he is out of town.

Issue #1726, opened 2019-12-20
University Physics vol1: Style equations and equation call out note. 


ajluk:
Equations in University Physics vol1 need to be styled according to Carnival Specs. 
Related issues: https://github.com/openstax/cnx-recipes/issues/1482, https://github.com/openstax/cnx-recipes/issues/1479

Changes in Theme Design Specs:
New shape needs to be added **BoxCondensedEquationShape**
containing components:
 
- **CondensedContainer**
    background: book specific 
    padding-top: v-grid(.5),
    padding-bottom:v-grid(.5),
    margin-bottom: v-grid(1),

- **CondensedEquationContainer**
   position: relative,
   display: flex,
   flex-direction: row, 
   align-items: center,
    justify-content: center,
    margin-bottom: 0,

- **BoxedCondensedNoteEquationNumber** (I quess I must change it later to CondensedEquationNumber )
    position: absolute,
    right: 0,
    top: 50%,
    transform: translate(0, -50%),
    border-width: 0.025rem,
    border-style: solid,
    line-height: normal,
    padding: 0.5rem,
    margin-right: v-grid(.5),
This shape will be used by **EquationCallOut** note. The only book specific property need to be added is background in **CondensedContainer** for University Physics:
 background: (_ref: "scheme:::Grey1"),

lwtchu:
Updated thank you :)

MartaSztuka:
Tested
![Screenshot from 2020-01-22 10-44-18](https://user-images.githubusercontent.com/47241114/72883207-32d1df80-3d04-11ea-80e0-e4f96b277b3d.png)
![Screenshot from 2020-01-22 10-47-06](https://user-images.githubusercontent.com/47241114/72883439-95c37680-3d04-11ea-816e-47cb08189478.png)

![Screenshot from 2020-01-22 10-43-47](https://user-images.githubusercontent.com/47241114/72883209-32d1df80-3d04-11ea-9f94-510788ccf365.png)
![Screenshot from 2020-01-22 10-46-03](https://user-images.githubusercontent.com/47241114/72883347-6f054000-3d04-11ea-8c1d-1e4d8f6c78f4.png)

PR #1727, opened 2019-12-20
Equations u physics

ajluk:
Issue: #1726


styles/output/u-physics-pdf.css
```diff
@@ -613,4 +613,50 @@ a {
...
+  border-style: solid;
+  line-height: normal;
+  padding: 0.5rem; }
+
+[data-type="equation"] {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin-bottom: 1.5rem; }
+
```
wilkus27: I think that some changes in equations settings are needed. When I have been creating shapes for equations I didn't predict that there could be two types of equations in one book (I mean numbered and unnumbered) and now there are two different settings for one selector `[data-type="equation"]`.

wilkus27: I mean that those selectors have to be more specific:

image

styles/designs/carnival/parts/notes/_notes-boxedCondensed-shapes.scss
```diff
@@ -0,0 +1,17 @@
...
+            _components: (
+            map-merge($EquationContainer, (
+                _components: (
+                  $EquationNumber,
+                    ),
+                ))
+            ),
+        ))
+    ),
+));
```
wilkus27: I'm not sure about this solution. Because `$EquationContainer` subselector is set to `""`, so it takes selector from the parent in this case from `$BoxCondensedEquationContainer`. But because this component has its own properties in output occurs the same selector with different properties:

image

I think that would be better to set in settings selector to `.equation-callout > [data-type="equation"]` and then removed from the shape `$BoxCondensedEquationContainer`. And all of the properties add to `$EquationContainer` but of course set them to OPTIONAL in `equation-components.scss` file.

It's just my proposition, @brittany-johnson I'm curious what do you think about such solution.

styles/output/u-physics-pdf.css
ajluk: I have added additional class selectors to numbered (:not(.unumbered) and unnumbered (.unnumbered) equations to distinguish them.

styles/designs/carnival/parts/notes/_notes-boxedCondensed-shapes.scss
ajluk: @wilkus27 I have added suggested changes and the code is ready for review. @brittany-johnson  please let me know what do you think about that solution as well.

styles/designs/carnival/parts/notes/_notes-boxedCondensed-shapes.scss
```diff
@@ -0,0 +1,13 @@
...
+//Boxed Condensed
+@include create_shape('carnival:::BoxCondensedEquationShape', (
+    _components: (
+        map-merge($EquationContainer, (
+            _components: (
+                $EquationNumber,
+            ),
+        )),
+    ),
+));
```
wilkus27: Ania I'm wondering now how can we use this solution for other Boxed Condensed notes. Because now background property is set in equation component (as I suggested). Is it possible to set background color to `[data-type="note"]` selector? If yes, we can create component directly related with Boxed Condensed and use it in the future for other notes. Sorry for the confusion, I just came up with this idea.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -40,6 +40,14 @@ $BoxedNoteContainerComponent: (
...
 );
 
+$BoxedCondensedNoteContainerComponent: (
+  _name: "CondensedContainer",
+  _subselector: '',
+  _properties: (
+    background: enum('ValueSet:::OPTIONAL'),
+  ),
+ );
+
```
ajluk: I have created additional container component for Boxed Condensed Notes with background property set to OPTIONAL.

styles/designs/carnival/parts/notes/_notes-boxedCondensed-shapes.scss
ajluk: Natalia I have added the solution the discussed (create container for Boxed Condensed Notes with empty selector instead of using [data-type="note"]). The code is ready for review.

wilkus27: Hmm I meant to use `[data-type="note"]`. So maybe to do somethink like that. To put here:
image

selector `[data-type="note"]..equation-callout`. Then this component:

image

will be related with selector from above.

styles/designs/carnival/parts/_equation-components.scss
```diff
@@ -19,8 +19,8 @@ $EquationNumber: (
         right: 0,
         top: 50%,
         transform: translate(0, -50%),
-        border-width: 0.025rem,
-        border-style: solid,
+        border-width:  enum('ValueSet:::OPTIONAL'),
+        border-style:  enum('ValueSet:::OPTIONAL'),
```
wilkus27: Because after discussion we decided to not use those equation components I think we can remove this change and leave explicit values.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -58,6 +58,45 @@ $BoxedNoteTitleComponent: (
...
 
+//boxed condensed
+$BoxedCondensedNoteContainerComponent: (
+  _name: "CondensedContainer",
+  _subselector: '',
+  _properties: (
+    background: enum('ValueSet:::REQUIRED'),
+    padding-top: enum('ValueSet:::REQUIRED'),
+    padding-bottom: enum('ValueSet:::REQUIRED'),
+    margin-bottom: enum('ValueSet:::REQUIRED'),
```
wilkus27: Can those properties have explicit values, the same for all Boxed Condensed Notes? Or do you think other notes of this type will have different paddings?

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -58,6 +58,45 @@ $BoxedNoteTitleComponent: (
...
+  ),
+ );
+ $BoxedCondensedNoteEquationContainer: (
+    _name: "BoxedCondensedNoteEquationContainer",
```
wilkus27: This is just a proposition :) I'm wondering if the name in `_name` could be shorter? Like it is done for component from above? I mean where name `$BoxedCondensedNoteContainerComponent` is detailed, but `_name: "CondensedContainer"` is not so precise.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -58,6 +58,45 @@ $BoxedNoteTitleComponent: (
...
+        display: flex,
+        flex-direction: enum('ValueSet:::OPTIONAL'),
+        align-items: center,
+        justify-content: center,
+        margin-bottom: 0,
+    )
+);
+
+$BoxedCondensedNoteEquationNumber: (
+    _name: "BoxedCondensedNoteEquationNumber",
```
wilkus27: Same suggestion here :)

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -58,6 +58,45 @@ $BoxedNoteTitleComponent: (
...
+    margin-bottom: enum('ValueSet:::REQUIRED'),
+  ),
+ );
+ $BoxedCondensedNoteEquationContainer: (
+    _name: "BoxedCondensedNoteEquationContainer",
+    _subselector: " > [data-type="equation"]",
+    _properties: (
+        position: enum('ValueSet:::OPTIONAL'),
+        display: flex,
+        flex-direction: enum('ValueSet:::OPTIONAL'),
```
wilkus27: Is it a reason why `position` and `flex-direction` are OPTIONAL here? In files related with equations they are set in this way, because there we have two types of equations and one of them use those properties, but another not. Here I think there is one BoxedCondensed note which have equation, so maybe we can set all of the properties (related with equations) in explicit way? What do you think?

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -58,6 +58,45 @@ $BoxedNoteTitleComponent: (
...
+$BoxedCondensedNoteEquationNumber: (
+    _name: "BoxedCondensedNoteEquationNumber",
+    _subselector: " > .os-equation-number",
+    _properties: (
+        position: absolute,
+        right: 0,
+        top: 50%,
+        transform: translate(0, -50%),
+        border-width:  enum('ValueSet:::OPTIONAL'),
+        border-style:  enum('ValueSet:::OPTIONAL'),
```
wilkus27: Same suggestion here.

styles/templates/template1/parts/_equation-settings.scss
```diff
@@ -1,12 +1,16 @@
...
-        _selectors: ('[data-type="equation"]'),
+        _selectors: ('[data-type="equation"]:not(.unnumbered)'),
         EquationContainer: (
             position: relative,
             flex-direction: row,
         ),
+        EquationNumber: (
+            border-width: 0.025rem,
+            border-style: solid,
+        ),
```
wilkus27: This can be moved back to the equation-components.scss file. See my comment in this file.

styles/designs/carnival/parts/_equation-components.scss
```diff
@@ -19,9 +19,10 @@ $EquationNumber: (
         right: 0,
         top: 50%,
         transform: translate(0, -50%),
-        border-width:  enum('ValueSet:::OPTIONAL'),
-        border-style:  enum('ValueSet:::OPTIONAL'),
+        border-width: 0.025rem,
+        border-style: solid,
         line-height: normal,
         padding: 0.5rem,
+        margin-right: v-grid(.5),
```
ajluk: I have added margin-right to the equation number for consistency, because adding padding right to the container doesn't work in this case (positioning). This is how it looks now: 
image

brittweinstein:
@ajluk Let's go ahead and merge this one too :)

PR #1739, opened 2019-12-30
style eoc u-physics

ajluk:
Issue: #1747


styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -142,3 +149,65 @@ $Exercise_Table_Row: (
...
+    )
+);
+$ChallengeProblems:(
+    _name: "ChallengeProblems",
+    _subselector: " > .os-review-challenge-container",
+    _properties:(
+        column-count: 2,
+        column-gap: 2.4rem,
+        column-width: auto,
+        margin-bottom: v-grid(4),
```
brittany-johnson: I noticed that you had to write this block of styles quite a few times and we've written it a few times for other EOC/EOB components that are in columns. 

I think if this can be rewritten in a more DRY way that would be great!

ajluk: I have rewritten the code modeling on other function created in this file. @brittany-johnson let me know if the solution is accurate in this case.

Issue #1771, opened 2020-01-20
hs physics: hide Teacher Support


wilkus27:
Element **Teacher Support** shouldn't be visible

![image](https://user-images.githubusercontent.com/40228854/72716277-39c8e880-3b72-11ea-9d61-cb2aac3801b4.png)

MartaSztuka:
no Teacher Support
![Screenshot from 2020-01-23 17-01-32](https://user-images.githubusercontent.com/47241114/73001002-08664c00-3e02-11ea-9f93-dfc1fadc1a79.png)

PR #1773, opened 2020-01-20
Hide Teacher Support

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1771


brittweinstein:
@wilkus27 I think we can merge this now! :)

Issue #1774, opened 2020-01-20
hs physics: style titles


wilkus27:
Chapter titles and titles from regular modules should be styled with specs from Carnival Theme.

![image](https://user-images.githubusercontent.com/40228854/72729930-a05cff00-3b90-11ea-8deb-fd463c764e8f.png)

MartaSztuka:
![Screenshot from 2020-01-23 17-03-07](https://user-images.githubusercontent.com/47241114/73001144-47949d00-3e02-11ea-8626-c4520f0c5205.png)
![Screenshot from 2020-01-23 17-02-53](https://user-images.githubusercontent.com/47241114/73001145-47949d00-3e02-11ea-8095-72ec7e9887dd.png)

PR #1775, opened 2020-01-20
Style titles in hs physics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1774


brittweinstein:
@wilkus27 we can merge now! 🎉

Issue #1818, opened 2020-01-27
Biology for AP Courses: Scientific method connection note 


ajluk:
Scientific method connection note in Biology for AP Courses has a different class than in Biology 2e and need to be styled according to Carnival Theme Specs. Related issue: https://github.com/openstax/cnx-recipes/issues/1190

Changes in TDS:
**ScientificMethodNote** book specific styles need to be added for **AP Biology** (column H):
**IconNote Container**:
group-border-color: B6
**IconNote Title**:
font-family': Muli,
color: B6,
**IconNote TitleIcon**:
background:  carnival-scientific.svg no-repeat top left,

Screenshot example from pdf file:
![image](https://user-images.githubusercontent.com/55203598/73555046-30326100-444d-11ea-8f6a-9f6c06f7d78c.png)

lwtchu:
Added to the specs

MartaSztuka:
styled
![Screenshot from 2020-02-04 11-24-20](https://user-images.githubusercontent.com/47241114/73736342-e98c8180-4740-11ea-9158-7200ec6f2f30.png)

PR #1819, opened 2020-01-27
apbiology: scientific method connection note 

ajluk:
Issue: #1818


brittweinstein:
@ajluk I think we're good to merge!

Issue #1833, opened 2020-01-29
hs physics: style eoc and eob


wilkus27:
End of Chapter and End of Book should be styled with specs from Carnival Theme.

wilkus27:
## Changes to provide to Carnival Theme

NOTE: there are created two new shapes `EocKeyEquWithSectionLink` and `EocSummaryWithSectionLink`. They are almost the same as shapes `EocKeyEqu` and `EocSummary`. The only difference is additional component `EocSectionTitleLink` with property: `font-family: Muli,`.

![image](https://user-images.githubusercontent.com/40228854/73732590-bba43e80-473a-11ea-8294-8e850bf2744d.png)

This is a temporary solution. New shapes were created to avoid changes in other books then HS Physics. During cleaning up Carnival Theme it's going to be created one shape. So my proposition is to not put those new shapes to Theme Design Specs for now and wait for future changes.

Here are the new shapes which should be put into the Theme Design Specs:

**Shape:** `EocCompositeChapter`
_Component:_ `Container`
_Properties:_
```
        margin-bottom: v-grid(4),
```

_Component:_ `EocHeader`
_Properties:_
```
        color: #00819A,
        font-weight: bold,
        font-size: font-scale(4),
        line-height: line-height(4),
        text-transform: uppercase,
        font-family: IBM Plex Sans, sans-serif;
        margin-bottom: v-grid(1)
```
**Elements using this shape**:
- `ChapterReviewCompositeChapter`
- `TestPrepCompositeChapter`

![image](https://user-images.githubusercontent.com/40228854/73733640-9ca6ac00-473c-11ea-81a6-5adffb4e80e8.png)

![image](https://user-images.githubusercontent.com/40228854/73733987-35d5c280-473d-11ea-8945-7f14889f4135.png)

wilkus27:
**Shape:** `EocCompChapExercises`

_Component:_ `Container`
_Properties:_
```
        margin-bottom: v-grid(4),
        column-count: enum('ValueSet:::OPTIONAL'),
        column-gap: enum('ValueSet:::OPTIONAL'),
        column-width: enum('ValueSet:::OPTIONAL'),
```

![image](https://user-images.githubusercontent.com/40228854/73743691-750d0f00-474f-11ea-8326-b00f2a5d2f46.png)

_Component:_ `CompChapPageHeader`
_Properties:_
```
        color: enum('ValueSet:::REQUIRED'),
[...]

PR #1834, opened 2020-01-29
Hs physics eoc eob

wilkus27:
Issue: #1833


styles/designs/carnival/parts/_eoc-shapes.scss
```diff
@@ -53,6 +53,61 @@
     ))
 );
 
+// This is temporary solution to avoid changes in other books.
+// During cleaning up Carnival Theme will be created one shape for Key Equation
```
wilkus27: The only one difference between this shape and shape `EocKeyEq` is section links component. HS Physics is the first book which contains section links in this part from EOC. So for now I created new shape to not provide changes for other books in which we can find Key Equation, but I think it's better to have one shape with section links component even some of the book don't have those links.

PR #1835, opened 2020-01-30
Style notes in hs physics

wilkus27:
Issue: #1839


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -66,36 +68,51 @@ $BoxedCondensedNoteContainerComponent: (
...
+    transform: translate(0, -50%),
+    border-width: 0.025rem,
+    border-style: solid,
+    line-height: normal,
+    padding: 0.5rem,
+    margin-right: v-grid(.5),
+  )
+);
+
+$BoxedCondensedNoteTitle: (
```
helenemccarron: Note for later: we should move this with the other `BoxedCondensedNoteContainerComponent`, for a second I thought we replaced the regular boxed condensed notes with the equation one

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -209,6 +226,20 @@ $IconNoteParagraphTitleComponent: (
   )
 );
 
+$IconNoteDivTitleComponent: (
```
helenemccarron: Same with this one

Issue #1813, opened 2020-01-24
chapter folio header continuing in appendices


oscryan:
# Scenario
The last chapter page title headers are continuing into the end-of-book content

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [x] Yes  [ ] No  (_attach below_)
<img width="589" alt="Screen Shot 2020-01-24 at 3 24 05 PM" src="https://user-images.githubusercontent.com/1895432/73105184-f0ec9900-3ebd-11ea-8d88-e2b586601afb.png">
<img width="593" alt="Screen Shot 2020-01-24 at 3 23 58 PM" src="https://user-images.githubusercontent.com/1895432/73105185-f0ec9900-3ebd-11ea-8cf4-fc406679bd6b.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

helenemccarron:
Addressed in https://github.com/openstax/cnx-recipes/pull/1892

MartaSztuka:
tested
![Screenshot from 2020-02-17 13-18-29](https://user-images.githubusercontent.com/47241114/74653225-0638a880-5188-11ea-9138-207dc24a15a0.png)

PR #1849, opened 2020-02-03
Remove last chapter page title headers from appendices eob

ajluk:
Issue:  #1813


helenemccarron:
addressed in https://github.com/openstax/cnx-recipes/pull/1892

Issue #1848, opened 2020-02-03
hs physics: style section exercises


wilkus27:
Text of the exercises and number should be inline:

![image](https://user-images.githubusercontent.com/40228854/73657455-9b20a980-4692-11ea-8000-a01cb2b5e837.png)

brittweinstein:
@Stackblocks could you please review Natalia's PR here? It is related to the table work you are doing.

wilkus27:
Some of the exercises have solutions:

![image](https://user-images.githubusercontent.com/40228854/73764773-d6df7000-4773-11ea-8a53-dc8dbba6462b.png)

Here are the specs for styling header:
```
IBM Plex Sans
Bold
font-scale(1)
Uppercase: off
```

lwtchu:
@wilkus27 what are the specs for the solution header? 

also is it possible to have the question stem go with the number

wilkus27:
@lwtchu sorry for the delay but I merged my PR at the end of my work so I had no time to prepare description for Theme Design Specs.

wilkus27:
## Changes to provide to Theme Design Spec

**New shape**: `ExerciseShape`

_Component:_ `Container`
_Properties:_ `empty`

![image](https://user-images.githubusercontent.com/40228854/73922645-cd681c00-48c9-11ea-8b82-de17c0872a62.png)

_Component:_ `ProblemWrapper`
_Properties:_ 
```
        display: flex,
        flex-direction: row,
```
![image](https://user-images.githubusercontent.com/40228854/73922706-e40e7300-48c9-11ea-938e-daf73c837723.png)

_Component:_ `ProblemNumber`
_Properties:_ 
```
        font-weight: bold,
        text-decoration: none,
```
![image](https://user-images.githubusercontent.com/40228854/73922778-fdafba80-48c9-11ea-8552-642b7b7c6b51.png)

_Component:_ `ProblemDivider`
_Properties:_ 
```
        margin-right: enum('ValueSet:::REQUIRED'),
```
![image](https://user-images.githubusercontent.com/40228854/73922838-14eea800-48ca-11ea-864a-4c063b877e5c.png)

_Component:_ `ProblemContainer`
_Properties:_ `empty`

![image](https://user-images.githubusercontent.com/40228854/73922874-2932a500-48ca-11ea-96aa-5af2750818ee.png)

_Component:_ `SolutionWrapper`
_Properties:_ `empty`

[...]

PR #1851, opened 2020-02-03
Section exercises in hs physics

wilkus27:
Issue: #1848


styles/designs/carnival/parts/_exercises-components.scss
```diff
@@ -0,0 +1,57 @@
...
+);
+
+$ExerciseProblemWrapper:(
+    _name: "ProblemWrapper",
+    _subselector: " [data-type='problem']",
+    _properties:(
+        display: flex,
+        flex-direction: row,
+    )
+);
```
wilkus27: @helenemccarron as I said on the call I didn't use `display: table` property because then when exercise starts with figure its bottom was displayed at the same level as number. With those properties it looks like that:

image

Now top of the figure is at the same level as number. Do you think it is acceptable? Figures will be centered, so it's not gonna be so close to the number.

helenemccarron: I think this is ok for now, let's create an issue to clean up the exercise component so we can reuse it in the various shapes.

Issue #1789, opened 2020-01-21
carnival theme: chapter summary EocHeader


lwtchu:
![image](https://user-images.githubusercontent.com/11299125/72840391-b71b5880-3c59-11ea-84f7-0c14926d517b.png)

Change "1.1 The Science of Biology" font to font-scale(2).

I THINK this is the EocSectionSummary EocHeader? But I'm not sure...let me know and I can change it on the spec sheet.

brittweinstein:
@lwtchu and @andymuddimer could you please confirm the header and fontscale do/don't match?

lwtchu:
@brittweinstein current it is at 14pt which isn't on the font scale. so we're thinking of having it be font-scale(2) therefore resulting in 12.96pt.

helenemccarron:
The specs need to be updated with the following:
`SectionTitleLink` and `EocSectionTitleLink` should include `font-size: font-scale(3)` after the font-family.

helenemccarron:
Note: this is an h3, so I applied the h3  font-scale(3)
![2020-02-04_11-49-33](https://user-images.githubusercontent.com/8514591/73771691-6cfba200-4744-11ea-840c-c3279af5db5b.png) cc @lwtchu

MartaSztuka:
tested
![Screenshot from 2020-02-11 10-57-54](https://user-images.githubusercontent.com/47241114/74226733-624d8980-4cbd-11ea-85a4-19be5ae51046.png)

PR #1857, opened 2020-02-04
Add font size to link h3 

helenemccarron:
Issue #1789


styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -91,6 +106,7 @@ $ReviewQuestionsTitle_Container: eocSectionTitleH3(".os-review-questions-contain
         _subselector: " #{$section_selector} > a > h3, #{$section_selector} >[data-type="composite-page"] > a > h3",
         _properties:(
             font-family: enum('ValueSet:::REQUIRED'),
+            font-size: font-scale(3)
```
wilkus27: @helenemccarron Ania asked me to look into this PR. The current code is little bit confusing. We definitely need to clean up this file. The problem is that previously we forgot to enable release flag related with section links, thats why in some of the shapes there are styled regular headers. And we use for that function from `83` line. But after we have enabled flag I created function for links, but I based on previous function, where for each element  (like Summary, Exercises, Review Questions) there is created separate component, so also each element has separate shape.

image

But I think it's not a good solution, because we repeat a lot of code, and shapes are almost the same. So in my last PR related with EOC in HS physics I created another function:

image

which is more general, it doesn't depend on specific class. 

And this component `$Eoc_SectionTitleLink` is used in shape `EocSummaryWithSectionLink`. So, change which you provide is changing font only in College Physics, where those components are applied:

image

and not in Biology. Font size property should also be added to the new function.

And as I'm writing here:

image

my proposition is to not create new shapes if the only one difference is presence or absence of section links in module. Please let me know what do you think about it.

helenemccarron: Understood, this is definitively not ideal. I'll add it to your function. If that's the case, I can remove the @include use('SummaryWithSectionLinks','carnival:::EocSummaryWithSectionLink'); from biology and it should work everywhere, correct?

helenemccarron: Nevermind, after trying it out, it looks like it's needed.

styles/books/biology/book.scss
```diff
@@ -125,7 +125,7 @@
 @include use('UnitToc', 'carnival:::UnitToc');
 @include use('Preface','carnival:::Preface');
 @include use('KeyTerm','carnival:::EocKeyTerm');
-@include use('Summary','carnival:::EocSummary');
+@include use('SummaryWithSectionLinks','carnival:::EocSummaryWithSectionLink');
```
wilkus27: @helenemccarron there should be added also those settings in this file:
```
SummaryWithSectionLinks:(
    _selectors: ("[data-type = 'chapter'] > .os-summary-container"),
    'EocHeader:::color': (_ref: 'colorMap:::physicsHeaderColor'),
  ),
```
instead of those settings:

image

ajluk: @helenemccarron I guess that it has been missed. There is name fix needed in the line 19 in the above file (from Summary to SummaryWithSectionLinks).

Issue #1635, opened 2019-11-22
Chapter Outline Cropped- Physics


MartaSztuka:
All locations where Chapter Outline is cropped:
1.
![Screenshot from 2019-11-22 17-12-15](https://user-images.githubusercontent.com/47241114/69441589-50f28200-0d4b-11ea-9810-0edb25c85dcc.png)
2.
![Screenshot from 2019-11-22 17-13-09](https://user-images.githubusercontent.com/47241114/69441652-62d42500-0d4b-11ea-8d3a-bc402469b1a3.png)
3.
![Screenshot from 2019-11-22 17-13-35](https://user-images.githubusercontent.com/47241114/69441690-71bad780-0d4b-11ea-9fec-1e446b3cdb6b.png)
4.
![Screenshot from 2019-11-22 17-13-58](https://user-images.githubusercontent.com/47241114/69441720-7f705d00-0d4b-11ea-8264-88ea64a609ae.png)
5.
![Screenshot from 2019-11-22 17-14-25](https://user-images.githubusercontent.com/47241114/69441755-8f883c80-0d4b-11ea-8ddc-99efd0e92b06.png)
6.
![Screenshot from 2019-11-22 17-14-49](https://user-images.githubusercontent.com/47241114/69441788-9d3dc200-0d4b-11ea-8379-a806f94fc3f6.png)
7.
![Screenshot from 2019-11-22 17-15-17](https://user-images.githubusercontent.com/47241114/69441826-ad55a180-0d4b-11ea-91bc-85337c4949ce.png)
8.
![Screenshot from 2019-11-22 17-16-05](https://user-images.githubusercontent.com/47241114/69441893-ce1df700-0d4b-11ea-8653-124e3b7f8c77.png)
9.
![Screenshot from 2019-11-22 17-16-36](https://user-images.githubusercontent.com/47241114/69441922-dd9d4000-0d4b-11ea-9dcc-98e4cc83dd37.png)
10.
![Screenshot from 2019-11-22 17-17-01](https://user-images.githubusercontent.com/47241114/69441940-ebeb5c00-0d4b-11ea-9e93-e97c997380e1.png)
11.
![Screenshot from 2019-11-22 17-17-30](https://user-images.githubusercontent.com/47241114/69441965-fc9bd200-0d4b-11ea-9cb2-d39fa8b32b71.png)
12.
[...]

PR #1867, opened 2020-02-05
chapter outline new design

ajluk:
Issue: #1635 
It looks like this now: 
image


styles/templates/template1/parts/_chapterintro-settings.scss
```diff
@@ -130,5 +138,84 @@
         SplashCaptionBg: (
             background-color: (_ref: "colorMap:::chapterSplashCaptionColor")
         ),
+    ),
+    ChapterIntroWithAbstract: (
+        _selectors: (".introduction"),
+        Outline: (
+            width: 6.4in,
```
ajluk: I have calculated the width value of the Outline Container to be equal width of the top border line with Chapter Outline title width with 8px margin included.

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -262,3 +281,19 @@ $ChapterSplashFigureWrapper: (
         $SplashCaption,
     )
 );
+
+$ChapterOutlineTitleBottomBorder: (
+    _name: "OutlineTitleBottomBorder",
+    _subselector: '::after',
+    _properties: (
+        content: "''",
+        width: 5in,
```
ajluk: The width has been calculated to be equal to width of the bottom-border line of the Outline Container (when the margin left which contains exact width of the Chpater Outile tile + 8px space is added).

ajluk:
@helenemccarron and @brittany-johnson I have assigned you to code review to this PR as well, so you can decide if the solution is acceptable.

styles/books/physics/book.scss
```diff
@@ -37,10 +37,14 @@
 ));
 
 @include add_settings((
-  ChapterIntro: (
+  ChapterIntroWithAbstract: (
```
helenemccarron: I think this should be renamed, historically, an abstract and a learning objective was the same but different mark up. Maybe something like fullWidthChapterOutline?

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -65,26 +73,38 @@ $ChapterOutlineObjectiveLink: (
...
 );
 
 $ChapterOutlineObjectiveLearning: (
     _name: "OutlineObjectiveLearning",
     _subselector: ' > .learning-objective',
+    _properties: (
+        margin-left: enum('ValueSet:::OPTIONAL'),
+    )
+);
+$ChapterOutlineAbstract: (
```
helenemccarron: See comment about naming

styles/designs/carnival/parts/_chapterintro-components.scss
helenemccarron: maybe add this as a comment in the code, we might forget why we have a "magic" number.

styles/designs/carnival/parts/_chapterintro-shapes.scss
```diff
@@ -64,19 +64,27 @@
 @include create_shape('carnival:::ChapterIntroWithAbstractShape', (
     _components: (
         $ChapterSplashFigureWrapper,
-        map-merge($ChapterIntroBody, (
+        map-merge($ChapterNoPropIntroBody, (
```
helenemccarron: What is NoProp? Can this be a bit more descriptive?

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -22,41 +22,49 @@ $ChapterOutline: (
     _name: "Outline",
     _subselector: ' > .os-chapter-outline',
     _properties: (
-        position: absolute,
-        width: 1.8in,
-        right: -2in,
+        position: enum('ValueSet:::OPTIONAL'),
```
helenemccarron: Why move all those to the settings? Are we expecting those to change often? It feels like that this is so specific, we shouldn't want to modify this often.

styles/templates/template1/parts/_chapterintro-settings.scss
```diff
@@ -14,9 +17,10 @@
             font-family: (_ref: "typography:::sansFont"),
             font-size: font-scale(1),
             line-height: line-height(1),
-            font-weight: 500, //medium
+            font-weight: 500,
```
helenemccarron: good catch 👍

styles/templates/template1/parts/_chapterintro-settings.scss
helenemccarron: See comment above about making a comment in the code.

styles/designs/carnival/parts/_chapterintro-components.scss
ajluk: @helenemccarron I needed to left some of these properties as OPTIONALS, because these components are used by two shapes and some of the properties are only used in one of them, rest I have left as REQUIRED, because these properties values differs depending on the shape. I can create new components for this shape if that will be a better solution, let me know what do you think?

styles/designs/carnival/parts/_chapterintro-components.scss
ajluk: Comments added! Changes are ready for review.

styles/designs/carnival/parts/_chapterintro-components.scss
helenemccarron: That works, thanks!

PR #1868, opened 2020-02-06
applied Natalia's styling from #1463 to Bio, Phys, and hsPhys

### LINKED ISSUE #1716: Carnival Theme: tables centering
Tables in Carnival Theme should display at the center. Width should be set to the widest element, and number/caption of the table should display on the left side of the table.

Examples:

image

image

image

Stackblocks:
Applied Natalia's work from #1463 to Bio, Phys, and HS Phys to center the tables, etc. and resolve #1716. We'll need to do this again for the other titles once these are out for publication.


recipes/books/_release.scss
```diff
@@ -252,6 +252,9 @@ $RELEASE_FLAGS: (
     "chemistry",
     "anatomy",
     "hs-physics",
+    "biology",
```
brittany-johnson: I think you may want to run the release flag script to regenerate the chart

Stackblocks: d'oh

PR #1877, opened 2020-02-06
Disable web styles

helenemccarron:
This disables web styles and the webpackable json file. This will speed up styles generation and merging.

The next step will be to see if we can do away with the build script but for now this halves the time it takes to generate the files and should help while I work on the next phase.

**Note: we might want to comment out books we don't use frequently, such as the extra volumes of Univeristy physics etc., this would speed up generating as well.


wilkus27:
@helenemccarron I tested it and I have found one thing :( 
I changed random property. First I set line-hight from `3` to `4` 

image

and after running `build-styles` any output file changed. 
I have tried to change other property and then everything was ok. 

Here is screen from output:

image

as you can see bold disappeared, but line-height is the same (as for value `3`).

helenemccarron:
@wilkus27 I think the behavior you are experiencing is expected. The line-height is not 1 to 1 (ie line-height(1) > line-height: 1rem. I got the same results as you did with line-height(4) but if go with something outrageous like line-height(20), I get the following:


I added a bunch of properties and got consistent results:

helenemccarron:
Since this is the week end for you and @brittany-johnson tested it successfully and we can all use ti to reduce pain generating styles, I'll merge this today. If you find anything else on Monday, I'll fix it right away, @wilkus27 .

Issue #1729, opened 2019-12-20
Invisible-bullet lists showing bullets


oscryan:
# Scenario
Markup: ```<list id="fs-idp137115008" list-type="bulleted" bullet-style="none">```
Showing bullets

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [ ] Yes  [ ] No  (_attach below_)
<img width="532" alt="Screen Shot 2019-12-20 at 11 15 08 AM" src="https://user-images.githubusercontent.com/1895432/71277550-de110280-231a-11ea-8800-9f7c3864b87d.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

ajluk:
Changes in TDS:
New Component needs to be added in BasicTableShape under the TableDataList (l.194):
TableDataUnstyledList
properties:
list-style-type: none,

MartaSztuka:
tested
![Screenshot from 2020-02-12 17-03-50](https://user-images.githubusercontent.com/47241114/74353063-ae7df400-4db9-11ea-83f9-bc0578ccb29f.png)

PR #1881, opened 2020-02-07
remove bullets from data-bullet-style:none lists

ajluk:
Issue: #1729


Stackblocks:
When I regenerate the book using this branch I still get the bulleted list in Appendix C, so it may need some additional work.

image

ajluk:
@Stackblocks it is awkward but I have checked it again today (bake and build pdf on this branch) and it looks fine. Could you check again?:
image

Stackblocks:
@ajluk You're right! it looks like it was an issue on my end. Thanks for double checking, i'm glad nothing else has to be done. 🎉 

image

ajluk:
Changes in TDS: 
New Component needs to be added in **BasicTableShape** under the **TableDataList** (l.194):
**TableDataUnstyledList**
properties:
list-style-type: none,

Issue #1331, opened 2019-08-27
Implement  full-width and full bleed in PDF 


helenemccarron:
Implement functions for the full-width and full bleed in PDF common, so they can be used throughout the designs

This needs to be implemented in Carnival and then it will be available for all themes going forward.

brittweinstein:
@brittweinstein put in SS Sprint.

brittweinstein:
This will be for every title--in Carnival theme and beyond.

openstaxalina:
We will implement the bleeds the same way this was done in the business series - example screenshot below.

This will scale the images up, **_proportionally_** (images will be expanded both vertically and horizontally).  

We know that this may necessitate some images to be replaced.  Therefore, when this issue is complete, we will do a targeted test of all splash images for all Carnival Theme books, and we will catalog images that are not suitable for the proportional expansion described above.  We will then source replacement for such images.  

(We also know that, even before this change, Katalyst has already identified some splash images for replacement.  If Katalyst prefers to discontinue testing/opening issues related to splash images until this bleeds issue is resolved, that is acceptable.) 

![Screen Shot 2019-11-25 at 11 34 07 AM](https://user-images.githubusercontent.com/7863733/69563871-8c719400-0f77-11ea-806f-6cfa6021e28e.png)

lwtchu:
Adding additional notes here: Any color bar that hits the edge of the page should be expanded to be 0.025in off the page as well for bleeds. Chapter outline, the beige splash image caption, EOC ribbons, etc will need to be a little larger for the bleeds.

I think the images don't necessarily have to be expanded but placed 0.025in off the edge of the page until we settle on a standard splash image size.

helenemccarron:
@wilkus27 Here is a screenshot inspecting the PDF elements and stretching them crudely to give you an idea of what we are trying to do. I hope this helps.
![2020-02-11_15-58-48](https://user-images.githubusercontent.com/8514591/74283704-f7fe0e80-4ce7-11ea-9464-d555929b3a8b.png)

lwtchu:
Might I add that, we can refrain from stretching the image, by simply _placing it_ off the page a little.

wilkus27:
FINAL DECISION: use streched image

wilkus27:
### Changes to provide to Theme Design Specs

#### Shape: `ChapterIntroShape`

_Component:_ `Outline`
_Properties:_ 
```
 margin: 0 -1.2rem 0 0,
```
[...]

PR #1886, opened 2020-02-10
Full bleeds

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1331


helenemccarron:
@wilkus27 I think the requirements for full-bleeds were unclear. We do not need the full bleed marks on the pdf itself, those will be enbled by the printer. What needs to be done, however, is to make sure that any image (the splash, the bands) are "bleeding" outside of the current page. I believe Thomas had that computed already, the work should be to double check that all areas are working as intended and fix what isn't. @lwtchu, I know this is a PR but would it be possible to add specific requirements and testing methods to the issue to make sure we are adhering to them?

wilkus27:
@helenemccarron I removed marks. It should be ok now. In this PR image is moved off the edge. If we are sure we don't want this solution this PR can be closed. Here https://github.com/openstax/cnx-recipes/pull/1913  are changes with streched image.

wilkus27:
Decision is to apply solution with streched images. This PR could be closed.

Issue #1783, opened 2020-01-21
Carnival theme: TOC Chapter Link


lwtchu:
TocChapterLink	
“font-family: “”IBM Plex Sans”“, sans-serif,
**font-size: font-scale(4),**
**line-height: line-height(4)**,
font-weight: bolder,
color: black,
text-decoration: none”

Change from font-scale(4) to font-scale(3)
line-height: line-height(3)

brittweinstein:
We will treat this as we will the header work.

helenemccarron:
![2020-02-10_14-59-32](https://user-images.githubusercontent.com/8514591/74189431-fc59f700-4c15-11ea-9b0b-8ae38792f3a8.png)

MartaSztuka:
tested
![Screenshot from 2020-02-14 10-41-36](https://user-images.githubusercontent.com/47241114/74519751-9bcd0180-4f16-11ea-94bb-af35d53fa141.png)
![Screenshot from 2020-02-14 11-09-27](https://user-images.githubusercontent.com/47241114/74521951-907bd500-4f1a-11ea-9760-1b3a487d029b.png)
![Screenshot from 2020-02-14 11-23-46](https://user-images.githubusercontent.com/47241114/74522952-7d6a0480-4f1c-11ea-8e14-b32b9a6e7952.png)

PR #1894, opened 2020-02-10
Change font-scale

helenemccarron:
issue #1783


brittweinstein:
@helenemccarron  I think we can go ahead and merge this :)

PR #1899, opened 2020-02-11
Generate the TOC during the disassembly step

m1yag1:
Generate the TOC during the disassembly step

### DIFF
--- a/script/disassemble.py
+++ b/script/disassemble.py
@@ -1,6 +1,6 @@
 import sys
 from pathlib import Path
-from shutil import rmtree
+
 from cnxepub.collation import reconstitute
 from cnxepub.formatters import DocumentContentFormatter
 from cnxepub.models import flatten_to, Document
--- a/script/disassemble.py
+++ b/script/disassemble.py
@@ -7,7 +7,6 @@ from cnxepub.models import flatten_to, Document
 
 in_dir = Path(sys.argv[1]).resolve(strict=True)
 out_dir = (in_dir / "disassembled").resolve(strict=True)
-# raw_dir = (in_dir / "raw").resolve(strict=True)
 baked_file = (in_dir / "collection.baked.xhtml").resolve(strict=True)
 
 with open(baked_file, "rb") as file:
--- a/script/disassemble.py
+++ b/script/disassemble.py
@@ -1,7 +1,11 @@
 import sys
 from pathlib import Path
 
+from lxml import etree
+from lxml.builder import ElementMaker, E
+
 from cnxepub.collation import reconstitute
+from cnxepub.html_parsers import HTML_DOCUMENT_NAMESPACES
 from cnxepub.formatters import DocumentContentFormatter
 from cnxepub.models import flatten_to, Document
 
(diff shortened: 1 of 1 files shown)


script/disassemble.py
```diff
@@ -1,18 +1,32 @@
...
 with open(baked_file, "rb") as file:
+    html_root = etree.parse(file)
     binder = reconstitute(file)
 
+nav = html_root.xpath("//xhtml:nav", namespaces=HTML_DOCUMENT_NAMESPACES)
+
+toc_maker = ElementMaker(namespace=None,
+                         nspam={None: "http://www.w3.org/1999/xhtml"})
+toc = toc_maker.html(E.head(E.title("Table of Contents")),
+                     E.body(nav))
```
rnathuji: @m1yag1 - Passing the nav data this way seems to be problematic (?) when I try to run as I get a `TypeError: bad argument type: list([<Element {http://www.w3.org/1999/xhtml}nav at 0x7f5b525c4508>])` .

rnathuji:
@m1yag1 - Just FYI, as part of #850 we'll be adding some basic testing for these various steps / scripts. You'll be merging this to master first anyways, so if it makes sense to you we can incorporate a basic test case for this as part of #850 since we'll have to merge with your changes anyways (e.g. to be sure we don't break you =) ).

script/disassemble.py
```diff
@@ -1,18 +1,32 @@
...
 baked_file = (in_dir / "collection.baked.xhtml").resolve(strict=True)
 
 with open(baked_file, "rb") as file:
+    html_root = etree.parse(file)
     binder = reconstitute(file)
 
+nav = html_root.xpath("//xhtml:nav", namespaces=HTML_DOCUMENT_NAMESPACES)
+
+toc_maker = ElementMaker(namespace=None,
+                         nspam={None: "http://www.w3.org/1999/xhtml"})
```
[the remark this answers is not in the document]
m1yag1: yep it should be. lol

script/disassemble.py
m1yag1: interesting... it ran fine for me.

m1yag1: now that i can execute tasks in concourse based on the code in my local machine i'll try it out and see.

script/disassemble.py
```diff
@@ -1,18 +1,32 @@
...
 
+nav = html_root.xpath("//xhtml:nav", namespaces=HTML_DOCUMENT_NAMESPACES)
+
+toc_maker = ElementMaker(namespace=None,
+                         nsmap={None: "http://www.w3.org/1999/xhtml"})
+toc = toc_maker.html(E.head(E.title("Table of Contents")),
+                     E.body(nav))
+
+with open(f"{out_dir}/collection.toc.xhtml", "wb") as out:
+    out.write(etree.tostring(toc, encoding="utf8", pretty_print=True))
```
tomjw64: pretty_printing typically comes with size/bandwidth drawbacks. What is the benefit of pretty_printing in this case?

m1yag1: I saw it used within the code in many places. That's the only reason.

m1yag1: @tomjw64 For example, in the same function you used to output the disassembled xhtml file it uses `pretty_print`. https://github.com/openstax/cnx-epub/blob/master/cnxepub/formatters.py#L64

m1yag1: and here: https://github.com/openstax/cnx-epub/blob/master/cnxepub/formatters.py#L307-L310

m1yag1: sooo.. is that a good reason? "That's the way it's always been done". It's kinda precedent at this point so I stuck with it. I can remove it if you think that would be best. I don't have a strong preference either way.

tomjw64: Good enough for me!

Issue #1802, opened 2020-01-23
Embedded content should generate link text for print


oscryan:
# Scenario
The Concept Trailers embedded in various chapters' intros in College Physics just say "Click to view content" with no url anywhere. The same is true of any linked content, including the Phets. That'll be awkward in print. We need a url for print so people can type it in if they want to visit from print version.

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [x] Yes  [ ] No  (_attach below_)
<img width="563" alt="Screen Shot 2020-01-23 at 12 11 11 PM" src="https://user-images.githubusercontent.com/1895432/73011485-ee187800-3dd9-11ea-8144-3087f522fa46.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

brittweinstein:
The link should display in parentheses after.

lwtchu:
![Screen Shot 2020-02-03 at 3 53 27 PM](https://user-images.githubusercontent.com/11299125/73694577-f2784700-469d-11ea-88d4-becf4dbb58f7.png)

![Screen Shot 2020-02-03 at 3 59 52 PM](https://user-images.githubusercontent.com/11299125/73694708-3cf9c380-469e-11ea-9b2e-958aa5170079.png)

here's a screenshot from the old pdf's if it helps.

MartaSztuka:
tetsted![Screenshot from 2020-02-20 15-48-55.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/802d5ce3-c216-4389-8408-245159d243fb)

PR #1903, opened 2020-02-12
display link adress in links

ajluk:
Issue: #1802


recipes/mixins/_modify.scss
```diff
@@ -155,6 +155,13 @@
     attr-target: "_blank";
   }
 }
+@mixin modify_interactiveNoteLinks() {
```
ajluk: @helenemccarron I have added you and @brittany-johnson to this review, because I wasn't sure if I created and placed this mixin correctly (It has been created for the links that are found in Link to learning notes). Let me know if it is the right solution.

ajluk:
How it look now:
image

ajluk:
How the links look now: 
image

styles/books/anatomy/book.scss
```diff
@@ -255,3 +255,9 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
 @page chapter:first {
   margin-right: 2in;
 }
+
+@media print {
```
helenemccarron: Should we put this in the common folder for PDF? This would avoid repetition.

ajluk: @helenemccarron I have put the rule in the _folio.scss file. Is it accepetable? I wanted to avoid creating new file just for few lines of code, but if it will be better please let me know.

styles/designs/common/pdf/_folio.scss
```diff
@@ -130,4 +130,10 @@ div[data-type="page"].preface {
         font-family: "Muli";
         font-weight: 700;
     }
-}
\ No newline at end of file
+}
+
+@media print {
```
helenemccarron: do we need the print selector? This should apply only to PDF (this is in the common PDF folder)

styles/books/anatomy/book.scss
brittany-johnson: @ajluk I agree, I think it would be too much to create a new file just for this, but also putting it in the folio file is a little weird too. There really isn't a great place for this. 

If we assume that there will eventually be more styles like this that are hard to place, maybe we should consider creating a miscellaneous styles file or common styles (as there is a common components file) 
OR we could rename common components to common styles instead of creating a new file. 

I think either option would be acceptable but I'm leaning toward the first. 

What do you think? @helenemccarron

ajluk: @helenemccarron I have created separate file as @brittany-johnson suggested and removed `@media print rule`.

styles/books/biology/book.scss
```diff
@@ -304,3 +304,10 @@ $bandBorder: #{$bandSize + $fullBleedWidth} solid $bandColor;
 @page chapter:first {
   margin-right: 2in;
 }
+
+@media print {
```
brittany-johnson: Sorry @ajluk,  I have a question just so that I can make sure I understand. 
Why do we need to have this implemented in both the book file and common styles file? Would it work as intended if it's just in the common styles file?

helenemccarron: I agree with @brittany-johnson, I think having it in the common file for PDF, without the `@media print` should be enough, unless we are missing something?

ajluk: @helenemccarron , @brittany-johnson sorry about the left code in the book.scss file, I missed it somehow. Now it has been removed and everything works fine.: 
image

brittweinstein:
@ajluk we can merge!

PR #1913, opened 2020-02-13
Bleed with streched image

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1331


wilkus27:
@helenemccarron in this PR is almost the same work as in https://github.com/openstax/cnx-recipes/pull/1886. But here image is streched.

styles/designs/common/pdf/_bleed.scss
```diff
@@ -0,0 +1,34 @@
+$BLEED_REM: 1.2;
+$BLEED_IN: 0.15;
+
+@function fullWidth($value, $unit) {
```
helenemccarron: I like the function to use inches or rems! However, I find those function names a bit confusing. When I read `fullWidth`, I think of a set width number (maybe the PDF's page width) but it looks like `fullWidth` is just twice the value of `half-fullWidth` then you are defining the element's width? 
If I am understanding what you are trying to achieve here is the normal width of an element + a predefined bleed, is that correct? Would something like `fullBleedsSize`be more descriptive and concise?

styles/designs/common/pdf/_bleed.scss
```diff
@@ -0,0 +1,34 @@
...
+
+@function half-fullWidth($value, $unit) {
+    @if $unit == 'rem' {
+        @return ($value + $BLEED_REM) * 1rem;
+    } @else if $unit == 'in' {
+        @return ($value + $BLEED_IN) * 1in;
+    }
+}
+
+@function fullMargin($value, $unit) {
```
helenemccarron: I think this would be clearer if this was renamed to be `fullBleedsPadding` and `fullBleedsMargin` or something like that

styles/designs/common/pdf/_bleed.scss
wilkus27: That is correct. Those functions are used to increase normal width of an element by bleed value. The difference between functions is that if element should "bleeding" on one side only then `half-fullWidth` is used and if we want to expend element to the right and to the left, the `fullWidth` should be used. For example:

image

Image container have bleeds just on the right side, so `half-fullWidth` is used but caption container is "bleeding" on both sides, so `fullWidth` is used.

helenemccarron: Ahhh I see! How about naming them fullWidthBleed and sideBleed? The fullwidth keyword is a bit loaded (the entire width of a page)

Issue #861, opened 2019-02-18
Overview and decisions: Implement Page Breaks Library-wide


helenemccarron:
Design and implement a comprehensive approach to page breaks for Baked PDF

List of items that should *not* break:
- [ ] Table headers
- [ ] Titles
- [ ] Exercises titles

openstaxalina:
@brittany-johnson will park questions at dev roundtable or sometime soon.

brittany-johnson:
**ToDo**
- [ ] Create a list of page break issues from present to 8/17 and group them into categories 
- [ ] Gather all of the page break styles from OER 
- [ ] Create a function or mixin to handle predicted page breaks 
- [ ] Test by comparing before and after mixin on baked pdf styles in different books

Per book/theme imports for page breaks (ex: Book doesn't have an answer key so don't import it)

**Test in:** 
- [ ] Economics 

**Question:** 
- Should page break mixins handle padding?
- What elements do we know that should never break? 
- Are there any elements that we have to explicitly say to break? 

**Page break related issues:** 

**Notes**
Title in note: 
https://github.com/openstax/oer.exports/issues/3786

Subtitle in Note 
https://github.com/openstax/oer.exports/issues/3878

Exampe (Note): 
https://github.com/openstax/oer.exports/issues/2628

**EOC**
EOC Titles: 
https://github.com/openstax/oer.exports/issues/3837
https://github.com/openstax/oer.exports/issues/3426
https://github.com/openstax/oer.exports/issues/3150

**Equations**
https://github.com/openstax/oer.exports/issues/3821

**Footnotes**
https://github.com/openstax/oer.exports/issues/3537

**Tables**
https://github.com/openstax/oer.exports/issues/3302

**Captions (Figures, tables, ect.)**
https://github.com/openstax/oer.exports/issues/3437

**Chapter Outline**
https://github.com/openstax/oer.exports/issues/3057

brittweinstein:
@brittweinstein create 5/7 release and epic and g.g. milestone and put this in there.

brittweinstein:
Do not put any page breaks anywhere for now.

openstaxalina:
We started opening issues with specific page break problems: https://github.com/openstax/cnx-recipes/issues/1462

brittweinstein:
For Devs: Make note of **any** page breaks/extra rules we should consider in this epic.

brittany-johnson:
[...]

PR #1914, opened 2020-02-13
Page breaks

brittany-johnson:
Link to spec docs: https://docs.google.com/spreadsheets/d/1vW0EHvn8fZgvTxZIm9xQXsqsnIDcolqwt5rWKtx4JEY/edit#gid=1804222758

more details: 
https://github.com/openstax/cnx-recipes/issues/861


styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -0,0 +1,78 @@
+//notes should be able to break but take the first element with them
```
helenemccarron: I really like the comment describing what the rules are, maybe we could have more of those so developers can refer to them later for ease and convenience?

styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -0,0 +1,78 @@
...
+nav#toc > ol > .os-toc-unit > ol > .os-toc-chapter > a > .os-number,
+nav#toc > ol > .os-toc-chapter > a > .os-number {
+  page-break-after: avoid;
+}
+
+div.preface {
+  page-break-before: right;
+  page-break-after: left;
+}
+
```
helenemccarron: I am not sure but I think we might want to add the rule to make sure the chapters like the index or appendices are starting on the left(?) side?

styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -0,0 +1,78 @@
...
+  break-inside: avoid;
+}
+
+.learning-objectives {
+  page-break-inside: avoid;
+  page-break-before: avoid;
+}
+
+[data-type="note"], [data-type="example"] {
+  page-break-inside: auto;
```
helenemccarron: Is this necessary? This should be default, unless something else is setting it to a different value?

styles/designs/common/pdf/_page-breaks.scss
brittany-johnson: I'm glad you pointed this out, this was actually a note for myself to implement this 😅

styles/designs/common/pdf/_page-breaks.scss
brittany-johnson: Ahhh okay. This is one that I brought over from oer. It was probably in place to overwrite something else lol. I'll remove it!

styles/designs/common/pdf/_page-breaks.scss
helenemccarron: Can you please open an issue for it, so we don't forget?

Issue #1910, opened 2020-02-12
hs physics: reduce space between eoc sections


lwtchu:
![image.png](https://images.zenhubusercontent.com/5b57310fb26dd832dfe45b43/c5e99620-f79c-4072-97a2-574603ce41cf)

Can we do like h-grid(2)? This seems excessive and only in the EOC between each sub section. And only in hs physics.

helenemccarron:
@lwtchu do you mean it seems to be happening only in hs physics or reduce the padding only in hs physics?  Currently all EOC containers have a  `margin-bottom: v-grid(4),`

lwtchu:
@helenemccarron i've only seen it in hs physics. but if we can do an over all change to `margin-bottom: v-grid(2)`? 

i dont' mind if it impacts the other books

helenemccarron:
You got it.

helenemccarron:
![2020-02-18_10-59-51](https://user-images.githubusercontent.com/8514591/74759047-cf24ce80-523d-11ea-9a67-d5c4617e4d69.png)

MartaSztuka:
tested![Screenshot from 2020-02-19 13-08-01.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6e1064b7-ea1a-46b5-a954-d273af69fed9)

PR #1917, opened 2020-02-13
reduce margins after eoc cntainer

helenemccarron:
issue #1910


brittweinstein:
@helenemccarron we can go ahead and merge!

Issue #1722, opened 2019-12-20
Post-section header (h2) padding too small


oscryan:
# Scenario
It's like a pixel

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [ ] Yes  [ ] No  (_attach below_)
<img width="549" alt="Screen Shot 2019-12-20 at 10 41 29 AM" src="https://user-images.githubusercontent.com/1895432/71269608-5b397900-2315-11ea-94f2-eb544eee0f26.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

brittweinstein:
Should be v-grid(1).

wilkus27:
## Changes to provide to Theme Design Specs

### Shape: `ModuleShape`

_New component:_ `TitleOverFigure`
_Properties:_
```
padding-top: v-grid(1),
```
![image](https://user-images.githubusercontent.com/40228854/74743444-ae21a500-5260-11ea-8f55-2a5627b5d8ee.png)

(Note: it is set for h4 headers over the image)

MartaSztuka:
tested
![Screenshot from 2020-02-19 17-08-46](https://user-images.githubusercontent.com/47241114/74850713-93bbfa00-533a-11ea-98f8-66e4fff9cef2.png)

PR #1918, opened 2020-02-14
Add top padding to figures placed under titles

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1722


brittweinstein:
@wilkus27 we can go ahead and merge!

Issue #1827, opened 2020-01-28
One column tables


wilkus27:
One column tables are not styled correctly:

![image](https://user-images.githubusercontent.com/40228854/73280828-5e1a6a00-41ef-11ea-9223-5370400b00fe.png)

Example from HS Physics. Styled with `BasicTableShape`.

In regular tables (with several columns) there are no borders on the left and right side. We need to decide what should be done in this case.

Stackblocks:
Resolved in PR #1921 with a temporary fix, but further work will need to be done. Created #1922 to track that work.

MartaSztuka:
tested
![Screenshot from 2020-02-20 17-31-21](https://user-images.githubusercontent.com/47241114/74956757-e19e3400-5406-11ea-892c-2e8949ee5581.png)

PR #1921, opened 2020-02-14
Issue 1827 - Single Column Tables

Stackblocks:
Interim fix for #1827 

# Before

# After


brittany-johnson:
The table looks good! Can we confirm if this had any negative visual effects on other tables in the same book and other books?

Stackblocks:
Tables appear to be fine in this title and others impacted by the change, but this will need to be legitimately tested to make sure.

## HS Physics



## Bio

PR #1923, opened 2020-02-14
Incorporate JSON metadata across pipeline steps

rnathuji:


script/bake-book.py
```diff
@@ -0,0 +1,14 @@
...
+
+json_data = {}
+
+# Passthrough but can be used to extract metadata from bake-book step
+
+with open(raw_meta, "r") as in_meta:
+    json_data = json.load(in_meta)
+
+with open(baked_meta, "w") as out_meta:
+    json.dump(json_data, out_meta)
```
tomjw64: This script seems like it's just doing `cp` (blocking approval)

script/disassemble-book.py
```diff
@@ -0,0 +1,109 @@
...
+    # NOTE: The returned tree has 'id' values which are based upon ident_hash
+    # fields in the provided model
+    tree = model_to_tree(binder)
+    amend_tree_with_slugs(tree)
+    slugs = {}
+    extract_slugs_from_tree(tree, slugs)
+    return slugs
+
+def main():
+    """Main function"""
```
tomjw64: Ya don't say... :stuck_out_tongue: (not blocking approval)

script/disassemble-book.py
```diff
@@ -0,0 +1,109 @@
...
+    computed slugs using ident_hash values"""
+
+    # Publishing calls these and these are needed to generate json,
+    # but they are harder to reach (cnx-publishing -> cnx-common)
+    #
+    # https://github.com/openstax/cnx-publishing/blob/master/cnxpublishing/bake.py#L92
+    # tree = model_to_tree(binder)
+    # amend_tree_with_slugs(tree)
+    #
+    # Using a similar approach here
```
tomjw64: Seems like this comment block can safely be removed now that you've made note above the `amend_tree_with_slugs` function (not blocking approval)

script/bake-book.py
rnathuji: Yes, as it turns out, that's what it amounted to since we only piped through `abstract` as one piece of metadata end-to-end and it didn't need anything else here. The reason we added / left this in there was to leave piping to decouple`collection.assembled-metadata.json` and `collection.baked-metadata.json` although the contents are the same for now. That way, as we identify more required metadata the script can be modified to inject additional metadata during this pipeline stage. It's not clear to me if this is over engineering or whether that will end up being useful (e.g for cnx #871). The alternative could be to remove this script altogether for now and have `disassemble-book` directly consume `collection.assembled-metadata.json` for the time being (basically bypassing `bake-book` for metadata).

script/disassemble-book.py
rnathuji: Sigh...yeah, I know. But I'm a slave to`pylint` and not particularly creative :)

script/bake-book.py
tomjw64: I was thinking more along the lines of actually using cp if you want to keep it separate, but using the raw meta as input I feel would work fine too. I don't see the baking step contributing metadata, at least in its current form, which we can't pull out during disassemble.

rnathuji: Ah, got it. I like the `cp` approach for a simple placeholder. I'll make the change.

Issue #1700, opened 2019-12-17
Notes mixins don't work properly


wilkus27:
From discussion in PR https://github.com/openstax/cnx-recipes/pull/1643

![image](https://user-images.githubusercontent.com/40228854/70984231-63e24300-20ba-11ea-9f9a-fed84d9b4759.png)

https://katalyst01.cnx.org/contents/1Q9uMg_a@13.51:mJu6YvwH@17/3-5-Free-Fall

lwtchu:
```
<note id="fs-id1168327878077" class="problem-solving">
<title>Problem-Solving Strategy: Free Fall</title>
<list id="fs-id1168328364214" list-type="enumerated" number-style="arabic">
<item>Decide on the sign of the acceleration of gravity. In <link target-id="fs-id1168328328571"/> through <link target-id="fs-id1168328282779"/>, acceleration <emphasis effect="italics">g</emphasis> is negative, which says the positive direction is upward and the negative direction is downward. In some problems, it may be useful to have acceleration <emphasis effect="italics">g</emphasis> as positive, indicating the positive direction is downward.</item>
<item>Draw a sketch of the problem. This helps visualize the physics involved.</item>
<item>Record the knowns and unknowns from the problem description. This helps devise a strategy for selecting the appropriate equations to solve the problem.</item>
<item>Decide which of <link target-id="fs-id1168328328571"/> through <link target-id="fs-id1168328282779"/> are to be used to solve for the unknowns.</item>
</list>
</note>
```

We'll need to remove "Problem Solving Strategy" from the subtitle. So it's JUST `Free Fall`. 

The class `problem-solving` should pull in the title Problem Solving Strategy.  And Free Fall should be just the subtitle.

You should be unblocked - it will just be displaying double until we can take the time to remove it in the content.

Let me know if you have any questions

brittweinstein:
@ajluk is this issue still a blocker? Are we able to start this now?

ajluk:
@brittweinstein I added the title label via recipe and created a content issue to remove the duplicated title label text in the subtitle: https://github.com/openstax/content-managers/issues/1451. There is an issue regarding University Physics blocked by it as well, but I guess we will work on this book later.

ajluk:
@MartaSztuka this issue need to be tested after fetching new content for College Physics.

MartaSztuka:
tested
[...]

PR #1924, opened 2020-02-17
add labelText phet note college physics

ajluk:
Issue:  #1700


ajluk:
Now it looks like this. The additional title text in subtitle needs to be removed from the content: 
image

brittweinstein:
@ajluk lets go ahead and merge this one :)

Issue #1730, opened 2019-12-20
Biology 2e, Concepts of Biology, Anatomy & Physiology baked pdfs missing answer key


oscryan:
# Scenario
The current pdfs have answer keys and the baked pdfs should too.
We suppressed it for recipe work but it is time to unsuppress.
# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [ ] Yes  [ ] No  (_attach below_)

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

MartaSztuka:
tested
![Screenshot from 2020-02-13 09-53-34](https://user-images.githubusercontent.com/47241114/74417287-bf287b80-4e46-11ea-813c-7f0575ca0837.png)
![Screenshot from 2020-02-13 10-10-49](https://user-images.githubusercontent.com/47241114/74418757-2e06d400-4e49-11ea-9476-1a51ccd8de6a.png)
 In Anatomy didn't work.
![Screenshot from 2020-02-13 10-17-47](https://user-images.githubusercontent.com/47241114/74419826-c3569800-4e4a-11ea-9a43-58f1f9d8cbe8.png)

openstaxalina:
I'm sorry, I missed this card coming through.  We purposely suppressed the Answer Keys in the recipes because we could not achieve the same set of answers in the old PDFs and the recipes.  

@ajluk could you please **undo this change** and close this card?  I apologize again for missing this.  cc @MartaSztuka

MartaSztuka:
Answer Key is no more.
![Screenshot from 2020-02-25 11-55-22](https://user-images.githubusercontent.com/47241114/75241097-c4ca7d80-57c5-11ea-9d1f-536b40904bea.png)
[...]

PR #1925, opened 2020-02-18
Suppress answerkey in Biology Anatomy

ajluk:
Due to the comment in Issue: #1730: https://github.com/openstax/cnx-recipes/issues/1730#issuecomment-585879873 Answer Key in Biology 2e and Anatomy has been suppressed again.


recipes/books/anatomy/_config.scss
```diff
@@ -60,7 +60,6 @@ $Config_ChapterCompositePages: (
 $Config_PartType_Exercise: (moveTo: $AREA_EOC, resetAt: $RESET_CHAPTER, numberAt: $NUMBER_BEFORE_MOVE, titleContent: $Config_exerciseTitleContent, );
 
 $Config_BookCompositePages: (
-  (className: "solution",    clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_EOB, name: "Answer Key"),
```
wilkus27: As I remember correctly, previously there was added also this line `$Config_hasCompositeChapter: true;` in this file. I this it should be removed.

recipes/books/biology/_config.scss
```diff
@@ -68,7 +68,6 @@ $Config_ChapterCompositePages: (
 );
 
 $Config_BookCompositePages: (
-  (className: "solution",    clusterBy: $CLUSTER_CHAPTER, compoundComposite: true, moveSolutionsTo: $AREA_EOB, name: "Answer Key"),
```
wilkus27: The same in this file.

recipes/books/anatomy/_config.scss
ajluk: Thanks for catching it up @wilkus27!

PR #1927, opened 2020-02-18
Issue 1900 -- Update Entrepreneurship details in books.txt

Stackblocks:
Changed books.txt to point to the new eship data.


brittweinstein:
@Stackblocks we can go ahead and merge since it has 2 approvers

Stackblocks:
@brittweinstein This is waiting on the completion of a content refresh on staging. If we merge this now HS-Stats will break.

Issue #1847, opened 2020-01-31
library wide: attribution link on the bottom every other page


lwtchu:
![Screen Shot 2020-01-31 at 1 14 05 PM](https://user-images.githubusercontent.com/11299125/73567242-95755a80-442b-11ea-913e-954c1b0df087.png)

This is missing in the baked PDFs. Need to discuss with @openstaxalina to define what this needs to be.

openstaxalina:
I'm checking this proposal with Anthony this week:

`Access for free at openstax.org.`   

This is generic for all books and is also consistent with the language REX uses (example at the bottom of this page: https://openstax.org/books/college-physics/pages/1-introduction-to-science-and-the-realm-of-physics-physical-quantities-and-units).

openstaxalina:
Approved by Anthony - getting final approval from David/Beth/Daniel on Monday at Ops.

brittweinstein:
@openstaxalina any update from the OPS parking lot?

openstaxalina:
Yes, this is finalized now.

We will use the attribution:

`Access for free at openstax.org` 

...for the footer language on every other page in all books in the library.

@lwtchu this does mean that, when we release a PDF in the baked styles, we need to change the language on the copyright page to use that phrase (currently it uses a long URL that is specific to the book on openstax.org).

ajluk:
Changes in TDS:
This property needs to be added in every other page type (page left ) in **Folio** except **Chapter Intro** (maybe addtiional row can be added - page: left to every page type).
bottom-left
font-size: font-scale(-1),
font-family: "Muli";
color: #A5A5A5;
font-weight: 700;
content: "Access for free at openstax.org.";

MartaSztuka:
tested
![Screenshot from 2020-02-25 12-44-17](https://user-images.githubusercontent.com/47241114/75244723-9bf9b680-57cc-11ea-8f27-8ac449100800.png)
![Screenshot from 2020-02-25 12-43-44](https://user-images.githubusercontent.com/47241114/75244736-9ef4a700-57cc-11ea-8475-6e89e6d1fddd.png)
![Screenshot from 2020-02-25 12-43-07](https://user-images.githubusercontent.com/47241114/75244738-9ef4a700-57cc-11ea-918f-8c191261f2bb.png)

PR #1930, opened 2020-02-19
add attribution link on every other page

ajluk:
Issue: #1847


ajluk:
@helenemccarron and @brittany-johnson I have added you to this review so you could verify if it is done correctly and acordingly to the requirements.

styles/designs/common/pdf/_folio.scss
```diff
@@ -120,6 +119,10 @@ div[data-type="composite-page"].os-eob {
       content: counter(page);
       margin-right: h-grid(5);
     }
+    @bottom-left {
+        @include folioStyles();
+        content: $attributionlink;
```
helenemccarron: You need to make a couple of changes before it will work. First, you created a variable but did not pass it (in the top of the mixin, l. 111), so it will not work. 
That said, in this case, the text should not change per location so I believe using the hard coded string is acceptable: `content: "Access for free at openstax.org.";`

helenemccarron: Also, do not forget to add the same ` @bottom-left ` block to both mixins so it will be displayed in the Appendices, EOC and EOB. (in `folioWithBand`).

ajluk: @helenemccarron I have made changes you have suggested and the code is ready for your review.

styles/designs/common/pdf/_folio.scss
```diff
@@ -158,10 +165,10 @@ $prefaceContent: "Preface";
 $eocContent: string(chapter-number) " • " string(eoc-title);
 $appendixContent: string(appendix-number) " • " string(appendix-title);
 $eobContent: string(eob-title);
-
-@include folio($defaultLeftContent, $defaultRightContent);
-@include folio($prefaceContent, $prefaceContent, preface );
-@include folioWithBand($eocContent, $eocContent, eoc);
-@include folioWithBand($appendixContent, $appendixContent, appendix);
-@include folioWithBand($eobContent, $eobContent, eob);
+$attributionlink: "Access for free at openstax.org.";
```
helenemccarron: As I mentioned in my previous comment, I think it would be simpler to avoid creating a variable for all the  mixins and hard code the string l.149 and  124 :D

ajluk: @helenemccarron I have removed the variable and the changes are ready for review.

ajluk:
@lwtchu Changes in TDS:
This property needs to be added in every other page (page left ) type in **Folio** except **Chapter Intro** (maybe addtiional row can be added - **page: left**) 
bottom-left
    font-size:  font-scale(-1),
    font-family: "Muli";
    color: #A5A5A5;
    font-weight: 700;
    content: "Access for free at openstax.org.";

ajluk:
Changes in TDS:
This property needs to be added in every other page type (page left ) in Folio except Chapter Intro (maybe addtiional row can be added - page: left to every page type).
bottom-left
font-size: font-scale(-1),
font-family: "Muli";
color: #A5A5A5;
font-weight: 700;
content: "Access for free at openstax.org.";

Issue #1932, opened 2020-02-19
High school physics: some composite pages are still going to page 1 or has no page number


lwtchu:
# What happened?
![image](https://user-images.githubusercontent.com/11299125/74853216-3c4e6780-5303-11ea-95ac-ee4f7fba8e51.png)
As you can see here Chapter Review and Test Prep are still going to page 1 in the ToC. 

# Reproducibility Details

This is in HS Physics.

lwtchu:
@brittweinstein i put the top three priority book labels on this first. but i tihnk it's impacting all books?

openstaxalina:
@lwtchu @brittweinstein I think that this is isolated to high school _Physics_.  Here are screenshots of the _College Physics_ and _Biology 2e_ PDFs generated Monday.  Are you observing this in PDFs generated more recently?

_College Physics_:
<img width="853" alt="Screen Shot 2020-02-19 at 10 37 52 AM" src="https://user-images.githubusercontent.com/7863733/74853761-fe057800-5303-11ea-9a67-bc81fc54e6c0.png">

_Biology 2e_:
<img width="862" alt="Screen Shot 2020-02-19 at 10 38 02 AM" src="https://user-images.githubusercontent.com/7863733/74853789-052c8600-5304-11ea-8589-0bcc6196890d.png">

lwtchu:
@openstaxalina bah you are correct i'm looking at old pdfs.

openstaxalina:
@lwtchu thank you -- I am going to delete some comments above just to clean up the issue to reduce confusion.

MartaSztuka:
tested
![Screenshot from 2020-02-25 12-33-57](https://user-images.githubusercontent.com/47241114/75243954-26411b00-57cb-11ea-9de5-412280a6ac8e.png)

PR #1937, opened 2020-02-20
Eoc composite chapters numbering in Toc

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1932

### DIFF
--- a/recipes/mixins/_modify.scss
+++ b/recipes/mixins/_modify.scss
@@ -973,7 +973,7 @@
 }
 
 @mixin modify_addCompositeIDsForToc() {
-  [data-type="composite-chapter"] > h1[data-type="document-title"]{
+  [data-type="composite-chapter"] > [data-type="document-title"]{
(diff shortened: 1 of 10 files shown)


recipes/mixins/_modify.scss
```diff
@@ -973,7 +973,7 @@
 }
 
 @mixin modify_addCompositeIDsForToc() {
-  [data-type="composite-chapter"] > h1[data-type="document-title"]{
+  [data-type="composite-chapter"] > [data-type="document-title"]{
```
wilkus27: There are two types of composite chapters. Answer Key in EOB with header in `h1` tag, but also EOC composite chapters which have header in `h2` tag. That's why it doesn't work for them.

brittweinstein:
@wilkus27 we can merge :)

wilkus27:
Note: Waiting for solving problems with checks in milestone branch.

Issue #1911, opened 2020-02-12
hs physics: key equations left align


lwtchu:
![image.png](https://images.zenhubusercontent.com/5b57310fb26dd832dfe45b43/868ef12f-b04d-44d3-9cb7-5ed1a5132eca)

Can we have the key equations be left aligned so they're not floating in the middle? ONLY the key equation table.

ajluk:
Changes in TDS: 
**TableContainer** need to be replaced in the **EocKeyEquWithSectionLink** shape to **KeyEquationsTableContainer** with following properties:
        margin-bottom: table-v-grid(2),
        margin-left: 0,
        margin-right: 0,
 **TableData** properties need to be changed to:
padding: table-v-grid(1) table-h-grid(1)
**TableHeadCell** properties need to be changed as well to:
 padding: table-v-grid(1) table-h-grid(1),
I couldn't find it in the table so it is possible the whole shape (**EocKeyEquWithSectionLink**) needs to be added there (But if it is somewhere please ignore the rest). These are it's components:
**EocHeader**
            font-family: B5
            font-weight: bold,
            font-size: font-scale(4),
            line-height: line-height(4),
            text-transform: uppercase,
          margin-bottom: v-grid(1)
**EocSectionTitleLink**
        font-family: Muli,
        font-size: font-scale(3),
**KeyEquations**
       padding: 0
**UnnumberedUnstyledTable**
        //empty properties
**TableHead**
    //empty properties
**TableRow**
    //empty properties
**TableHeadCell**
  padding: table-v-grid(1) table-h-grid(1),
**TableBody**
   //empty properties
**TableData**
  padding: table-v-grid(1) table-h-grid(1),

MartaSztuka:
tested![Screenshot from 2020-02-27 11-49-28.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/e43e3d63-aed7-4a2f-999c-3de78a37f4fc)

PR #1940, opened 2020-02-20
key equations tables alignement hs physics

ajluk:
Issue:  #1911


styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -528,11 +528,13 @@
...
         ),
-        TableContainer: (
+        KeyEquationsTableContainer: (
             margin-bottom: v-grid(2),
+            margin-left: 0,
+            margin-right: 0,
         ),
         TableData: (
-            padding: v-grid(1) h-grid(1), 
+            padding: v-grid(.5) h-grid(.5),
```
Stackblocks: Should this use `table-v-grid(1)` and `table-h-grid(1)` so it is aligned with the other table spacing constructs?

styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -286,3 +286,12 @@ $Exercise_Table_Row: (
         padding: enum('ValueSet:::OPTIONAL'),
     )
 );
+$Key_Equations_Table_Container: (
+    _name: "KeyEquationsTableContainer",
+    _subselector: " > .os-table",
+    _properties:(
+        margin-bottom: enum('ValueSet:::REQUIRED'),
+        margin-left: enum('ValueSet:::OPTIONAL'),
+        margin-right: enum('ValueSet:::OPTIONAL'),
```
wilkus27: Is there a reason why those properties are not hard coded? I think that Key Equation always has to have such margins, but maybe I miss something.

ajluk: @wilkus27 Yes, you're right. I have replaced them as hard coded. The code is ready for review.

styles/templates/template1/parts/_eoc-settings.scss
ajluk: @Stackblocks you're right I replaced the functions with table-grid ones.

styles/output/hs-physics-pdf.css
```diff
@@ -1559,14 +1559,16 @@ iframe {
...
-  margin-bottom: 3rem; }
+.os-eoc.os-key-equations-container > .key-equations > .os-table {
+  margin-bottom: 1.5rem;
+  margin-left: 0;
+  margin-right: 0; }
 
-.os-eoc.os-key-equations-container > .key-equations > .unnumbered.unstyled > thead > tr > th {
-  padding: 1.5rem 8px; }
+.os-eoc.os-key-equations-container > .key-equations > .os-table > .unnumbered.unstyled > thead > tr > th {
+  padding: 0.75rem 0.75rem; }
```
wilkus27: @Stackblocks I wasn't sure why after Ania applied `table-v-grid` and `table-h-grid` function units changed from px to rem. I checked in the code and both functions use `v-grid`:

image

Is it intentional?

brittweinstein:
@ajluk we can merge this one

styles/output/hs-physics-pdf.css
Stackblocks: image

@wilkus27 the `table-v/h-grids` both use the older `half-v-grid` and are simply a vertical and horizontal pair of aliases, so this is expected behavior.

wilkus27: Ok :+1: Thank you @Stackblocks for confirmation :relaxed:

PR #1962, opened 2020-02-25
Page break update

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/861


styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -1,23 +1,40 @@
...
 nav#toc > ol > .os-toc-unit > ol > .os-toc-chapter > a > .os-number,
-nav#toc > ol > .os-toc-chapter > a > .os-number {
+nav#toc > ol > .os-toc-unit > ol > .os-toc-chapter > a > .os-text,
+nav#toc > ol > .os-toc-chapter > a > .os-number,
+nav#toc > ol > .os-toc-chapter > a > .os-text,
+ {
   page-break-after: avoid;
 }
 
+//The list of module titles should not break before so that it
```
helenemccarron: so that it....? :D

styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -1,23 +1,40 @@
...
 [data-type="document-title"],
 [data-type="title"] {
   break-after: avoid;
   break-inside: avoid;
 }
 
-.learning-objectives {
+//Unfortunately the singular and pluaral form of these classes exists
+//LO's should be orphaned from their title 
+.learning-objectives,
```
helenemccarron: Should we open an issue to solve this?
Also, should this read "LOs **shouldn't** be orphaned"

brittany-johnson: Yes! I've made a note to open an issue

brittany-johnson:
@helenemccarron I've updated the comments. Thanks for catching that 😅

Issue #1747, opened 2020-01-08
University Physics: style EoC sections


ajluk:
EoC sections need to be styled according to Carnival Theme Specs. Related issues: #1486, #1479.

brittweinstein:
@ajluk I'm going to move this to the backlog since Phys isn't high prio rn

ajluk:
Changes in TDS: 
New exercises needs to be added under the EocCompChapExercises (the all using it):

**ChallengeProblems**

          Container
          //empty properties

          CompChapPageHeader:(
             font-family: ("IBM Plex Sans", sans-serif),
             column-span: all,
             // color: book specific, defined in book.scss file
             font-weight: bold,
             font-size: font-scale(3),
             line-height: line-height(3),
             text-transform: uppercase,
            margin-bottom: v-grid(1)
        
        EocSectionTitleLink:
            font-family: Muli
            font-size: font-scale(3),

        ProblemWrapper
          display: table,
       
       ProblemContainer
           display: table-cell,

       ProblemParagraph
          margin-bottom: 0,

        ProblemNumber: 
             color: black,
            font-weight: bold,
            text-decoration: none
            display: table-cell,
        
        ProblemDivider: (
            margin-right: h-grid(1),
        
        ExerciseOrderedList: 
            margin-left: h-grid(4),
            margin-bottom: v-grid(1),
        
        ExerciseFigure: 
            max-width: 100%,
        
**ProblemsCompChapPageWithTwoColumns**

        Container:
            column-count: 2,
            column-gap: 2.4rem,
            column-width: auto,
        
        CompChapPageHeader
             font-family: ("IBM Plex Sans", sans-serif),
             column-span: all,
             // color: book specific, defined in book.scss file
             font-weight: bold,
             font-size: font-scale(3),
             line-height: line-height(3),
             text-transform: uppercase,
            margin-bottom: v-grid(1)
        
        EocSectionTitleLink:(
              font-family: Muli
            font-size: font-scale(3),

        ProblemWrapper
          display: table,

        ProblemContainer
        display: table-cell,

          ProblemParagraph
          margin-bottom: 0,

          ProblemNumber: 
              color: black,
            font-weight: bold,
            text-decoration: none
            display: table-cell,

        ProblemDivider:
            margin-right: h-grid(1)

        ExerciseOrderedList:
            margin-left: h-grid(4),
            margin-bottom: v-grid(1),

        ExerciseNestedOrderedList: 
[...]

Issue #1942, opened 2020-02-21
CLEAN UP: refactor eoc components/shapes to be more reusable


helenemccarron:
There are too many uniquely named components, could we consolidate those into fewer shapes and components.

wilkus27:
There are four types of EOC modules:
- Key Terms
- Key Equations
- Summary
- Exercises

For each of this type we need separate shape.

There are two types of differences which appear in different books for each of these modules. 

1. Module has section links or not.
2. Module is composite page and has h2 header or module is compound composite page and has h3 header. (h2 and h3 are styled differently)

So what is the best solution in this case?

a) to have one shape, e.g for Exercises, in which components for h2, h3 and section links are added and all their properties are hard coded, even those elements don't appear in all Exercises modules?

b) to create one shape with components for h2, h3 and section links, but their properties are OPTIONAL and defined in settings.scss file if element exists?

c) to create separate shape for each case (e.g Exercises with links and h2, Exercises with links and h3, Exercises without links and with h2 etc.)?

ajluk:
The problem with column gap getting tighter described in this issue: https://github.com/openstax/cnx-recipes/issues/1978 is related to EoC Exercise cleaning issue included here. The thing that causing the problem is display: table and display: table-caption properties set for Problem Wrapper and Problem Container. There is a separate issue related to this, created by Natalia (#1870) with proposal changes to exercises.

helenemccarron:
A lot of the initial work was done as part of #2074. Some more clean up is necessary for all the books.

brittweinstein:
We're keeping this in the Icebox to double-check after all shapes/components work is done that it is resolved.

wilkus27:
Will be resolved after merge work from https://github.com/openstax/cnx-recipes/issues/2188.

wilkus27:
Tested here https://github.com/openstax/cnx-recipes/issues/2188

Issue #1767, opened 2020-01-16
CARNIVAL CLEAN UP - review shapes for EOC and reduce clutter if possible


helenemccarron:
There are many shapes that seem to be repeating the same structure for the different EOC parts, do we need all of them or can we consolidate them to avoid future bugs.

wilkus27:
Duplicate of https://github.com/openstax/cnx-recipes/issues/1942. I'm closing this issue.

PR #1964, opened 2020-02-26
add style to EoC uphysics

ajluk:
Issue: #1747


styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -143,15 +143,29 @@ $SummaryOrderedList:(
...
 // During cleaning up Carnival Theme we can remove functions from above 
 // and aply this general link component to other shapes
 $Eoc_SectionTitleLink:(
     _name: "EocSectionTitleLink",
     _subselector: " a:first-of-type > h3",
     _properties:(
-        font-family: enum('ValueSet:::REQUIRED'),
-        font-size: font-scale(3),
+        font-family: enum('ValueSet:::OPTIONAL'),
+        font-size: enum('ValueSet:::OPTIONAL'),
```
wilkus27: I think it will be better to leave this hard coded for now, because it's causing changes in some books (look at my comment in ap-biology-pdf.scss file). To be sure we don't change anything, all books should be checked whether they have links or not in some modules. So maybe better is to define too much for now rather then not enough. We can work on that in https://github.com/openstax/cnx-recipes/issues/1767

ajluk: @wilkus27 I have revert back the section links ($Eoc_SectionTitleLink) font property's values to hardcoded/REQUIRED and added them to the left exercises to avoid changes in the other components in the other books.

ajluk:
The cleaning of the EoC files will be done later within this issue: #1942

brittweinstein:
@ajluk i think we're good to merge!

Issue #1965, opened 2020-02-26
Enable all feature flags for University Physics


ajluk:
All feature flags need to be enabled for University Physics.

ajluk:
Enabled flags:

- chapter-page-class
- appendix-has-numbered-examples
- increment-section-counter-for-lo
- trash-abstract-in-preface
- titles-in-examples
- module-titled-notes
- use-title-nodes
- goodbye-warnings
- table-move-fake-title
- several-composite-chapters
- subtitled-examples

MartaSztuka:
- [ ] chapter-page-class
- [ ] appendix-has-numbered-examples
- [ ] increment-section-counter-for-lo
- [ ] trash-abstract-in-preface
- [ ] titles-in-examples
- [ ] module-titled-notes
- [ ] use-title-nodes
- [ ] goodbye-warnings
- [ ] table-move-fake-title
- [ ] several-composite-chapters
- [ ] subtitled-examples

ajluk:
Now the Example subtitle looks like this: 
![image](https://user-images.githubusercontent.com/55203598/75567989-c197f180-5a52-11ea-82f8-0a6e176d2135.png)

MartaSztuka:
Examples subtitles looks ok.
![Screenshot from 2020-03-03 13-13-02](https://user-images.githubusercontent.com/47241114/75774741-da551f80-5d50-11ea-8038-82837912c24c.png)

PR #1966, opened 2020-02-26
enable all feature flags uphysics

ajluk:
Issue:  #1965


recipes/books/_release.scss
```diff
@@ -255,18 +263,21 @@ $RELEASE_FLAGS: (
...
+    "u-physics",
   ),
   subtitled-examples: enable-for(
     "physics",
     "ap-physics",
     "hs-physics",
     "biology",
     "ap-biology",
-    "-biology",
+    "biology",
```
wilkus27: Good catch! :relaxed: But I think we can just removed `-biology`, because two lines above there is `biology`.

brittweinstein:
@wilkus27 lets go ahead and merge this :)

Issue #1955, opened 2020-02-24
Symmetry Feedback: Reduce Page Count: Link to Learning Feature


openstaxalina:
# Scenario

I would like to make a change here to reduce page count. Please provide specifications.  This may require content and/or code changes.
![Screen Shot 2020-02-24 at 1 36 54 PM](https://user-images.githubusercontent.com/7863733/75184935-d733b100-570a-11ea-994b-bbcfd278b350.png)

# Acceptance Criteria (required)

# Test Plan

ajluk:
It is caused by bottom margin set to 1.5 rem for all p elements in the books: 
![image](https://user-images.githubusercontent.com/55203598/75248853-4249ba00-57d5-11ea-9a9c-17e02152560a.png)
@lwtchu Should it be completely removed from paragraph in this type of notes?

lwtchu:
We can do two things:

- Reduce the size to `1 rem`
- Move the links into the previous paragraph so that it's not a separate sentence

lwtchu:
Corresponding content issue: https://app.zenhub.com/workspaces/content-management-5c86d8c40b66e07993f45cfa/issues/openstax/content-managers/1474

ajluk:
@lwtchu I have set it to `margin-bottom: half-v-grid(1)` which return 0.75rem, because we don't have a function that returns 1rem for vertical spaceing and now it looks like this. Is it ok?: 
![image](https://user-images.githubusercontent.com/55203598/75440384-fa08d400-595b-11ea-98fa-8ac3257cc706.png)

ajluk:
I tried set the bottom margin to 0 as well. Maybe it will be a better solution. @lwtchu what do you think?  Here is how it looks: 
![image](https://user-images.githubusercontent.com/55203598/75441146-6b955200-595d-11ea-9e72-f59d4c01494d.png)

lwtchu:
Can we do a half v grid - and will this apply to all p elements?

ajluk:
Ok. I will add this to all p elelements in this note and check how it looks.

lwtchu:
@ajluk let's do this for all p elements in condensed notes

ajluk:
Changes in TDS:
**Paragraph** component need to be added to  **IconCondensedShape** and **IconCondensedExerciseShape** as a replacement of existing **EndText** component in these shapes l.334 and l.347) with properties:
 margin-bottom: half-v-grid(1),

MartaSztuka:
[...]

PR #1970, opened 2020-02-27
reduce margin-bottom link to learning biology

ajluk:
Issue: #1955


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -339,9 +339,9 @@ $IconCondensedBodyComponent: (
   )
 );
 
-$IconCondensedEndTextComponent: (
-  _name: "EndText",
-  _subselector: " > p:last-child",
+$IconCondensedParagraphComponent: (
+  _name: "Pargraph",
```
Stackblocks: Would this need to be `"Paragraph",` ?

Issue #1971, opened 2020-02-27
folio: composite chapters in EOC


wilkus27:
Settings for folio didin't work for composite chapters in EOC:

![image](https://user-images.githubusercontent.com/40228854/75450932-f67f4800-596f-11ea-842b-62f5a646af3c.png)

![image](https://user-images.githubusercontent.com/40228854/75450989-0e56cc00-5970-11ea-900e-9ad54b54bff4.png)

As you can see on the screenshots, even Chapter Review starts on the next page there is still label **Key Equations**.

wilkus27:
@lwtchu There should be label of composite chapter: **Chapter Review** or of its composite page: **Concept Items**?

lwtchu:
Chapter Review

MartaSztuka:
tested
HS Physics
![Screenshot from 2020-03-02 12-09-17](https://user-images.githubusercontent.com/47241114/75671293-b5907780-5c7e-11ea-8bef-2cfe829aaebc.png)
University Physics - folio doesn't work for EoC, also there is blank page with header only at each chapter's end.
![Screenshot from 2020-03-02 12-13-03](https://user-images.githubusercontent.com/47241114/75671711-88909480-5c7f-11ea-9204-4bca05f30151.png)
![Screenshot from 2020-03-02 12-12-52](https://user-images.githubusercontent.com/47241114/75671713-89292b00-5c7f-11ea-878c-2c9216a4f550.png)

MartaSztuka:
University Physics
![Screenshot from 2020-03-02 16-59-05](https://user-images.githubusercontent.com/47241114/75693425-35cad300-5ca7-11ea-95c1-d2e857d55b88.png)
![Screenshot from 2020-03-02 17-00-18](https://user-images.githubusercontent.com/47241114/75693524-51ce7480-5ca7-11ea-8dc7-72080bcfe2f1.png)
![Screenshot from 2020-03-02 17-01-08](https://user-images.githubusercontent.com/47241114/75693600-74f92400-5ca7-11ea-9d03-18ee80930603.png)

PR #1975, opened 2020-02-28
Folio eoc

wilkus27:
Issue: #1971


brittweinstein:
@wilkus27 we can go ahead and merge with 2 reviewers

PR #1976, opened 2020-02-28
u-physics: subtitled-examples flag

ajluk:
Issue: https://github.com/openstax/cnx-recipes/issues/1965
After enabling all feature flags for University Physics during the tests it appeared that the last flag `subtitled-examples` doesn't work correctly in this book (the example subtitle isn't moved to h4 tag as it should be): 
image


ajluk:
Now it looks like this: 
image

Issue #1462, opened 2019-10-16
Learning Objectives should not be allowed to break


MartaSztuka:
There are single lines and tables/figure captions on tops and bottoms of the pages. 
![Screenshot from 2019-10-16 15-16-35](https://user-images.githubusercontent.com/47241114/66922943-4c3dfe00-f028-11e9-95fe-088ab52945f2.png)

MartaSztuka:
tested
![Screenshot from 2020-03-05 14-29-52](https://user-images.githubusercontent.com/47241114/75986172-de6a7400-5eed-11ea-8e3f-bc32ab3b03f9.png)
![Screenshot from 2020-03-05 14-27-41](https://user-images.githubusercontent.com/47241114/75986176-df030a80-5eed-11ea-83e3-0c339532bcc1.png)

PR #1987, opened 2020-03-02
add page brak to learning objectives in Chemistry

ajluk:
Issue: #1462


ajluk:
In Chemistry and Chemistry AF Learning Objectives have different markup than other books. Additional selector in page breaks for this element is needed: 
image

styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -27,7 +27,8 @@ h1, h2, h3, h4, h5, h6,
 //Unfortunately the singular and pluaral form of these classes exists
 //LO's shouldn't be orphaned from their title 
 .learning-objectives,
-.learning-objective {
+.learning-objective,
+[data-type="page"]:not(.instroduction) > [data-type="abstract"] {
```
brittany-johnson: It looks like this class may be misspelled? Is this selector working: `.instroduction`

brittany-johnson: Ah sorry! I Just saw the screenshot. We should have this updated by content

brittany-johnson: Is this misspelled class `instroduction` throughout the book? @ajluk

ajluk: Yes it was working anyway, that's why I haven't noticed the typo. Thanks for catching it up!

Issue #1467, opened 2019-10-17
Carnival Theme: Embiggened Footnotes


Stackblocks:
# What happened?

Something has happened to the footnotes to cause them to be `font-scale(1)` (10.8pt) instead of `font-scale(0)` (9pt) or less. 

# Reproducibility Details

- Occurred in Chemistry and Microbiology on `traitorous.treacletart` across multiple bakes.

# Environment:

- **Book(s)**: Chemistry, Microbiology, probably others.
- **URL/Server**: easyvm5.cnx.org
- **Location (page number, chapter/section)**: 
    - Microbiology:  p. 114, Appendix E.
    - Chemistry:  pp. 1291-2

# Screenshots
## Microbiology
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/59d06bdb-aef2-4f4d-b3b5-ab5742a3c1a8)
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/fd095895-31ff-4016-9e99-84733093c9c4)
---
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/069accb7-397d-41d0-998d-b20d548ab32a)
## Chemistry
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/cbfdc544-b39d-4b43-91c4-f7a5a6c38af6)
# Additional notes

 (_For example, steps to reproduce, related issues, more context, etc._)

helenemccarron:
Should the font-scale be -1? @lwtchu

lwtchu:
@helenemccarron are the footnotes in the main text also -1?

lwtchu:
@helenemccarron i misread this. yes let's do -1

ajluk:
Changes in TDS:
in Footnotes shapes elements in below properties only the values has been changed:
**Footnotes** shape
**Container**  l.204
line-height: line-height(1), 
font-size: font-scale(-1),
**FootnoteCall** shape (this name of the shape should be added as a separate row under l.211 in the way how it was done for Footnotes shape in l.203, I guess that was missed earlier :))
**DefaultFootnoteCall** l.211
 line-height: line-height(0),
font-size: font-scale(-1),

ajluk:
![image](https://user-images.githubusercontent.com/55203598/76217036-f4d83e80-6211-11ea-954f-43416e2aedde.png)

[...]

PR #1988, opened 2020-03-02
decrease font size footnotes

ajluk:
Issue:  #1467


styles/templates/template1/parts/_footnotes-settings.scss
```diff
@@ -8,7 +8,7 @@
             color: (_ref: 'colorMap:::fontBodyColor'),
             line-height: line-height(1),
             font-family: (_ref: 'typography:::baseFont'),
-            font-size: font-scale(0),
+            font-size: font-scale(-1),
```
wilkus27: @ajluk what about `line-height` in this case? It's `1.5rem` at this moment. Shouldn't be smaller too? @helenemccarron what do you think?

helenemccarron: yep, agreed.

ajluk: @wilkus27 I have decreased line-height as well consistently. I havn't done it earlier, because it wasn't mentioned in the issue. Now it looks like this: 
image

styles/output/ap-biology-pdf.css
```diff
@@ -1318,7 +1318,7 @@ a {
   color: #000000;
   line-height: 1.5rem;
   font-family: Alegreya;
-  font-size: 1rem; }
+  font-size: 0.83333rem; }
```
wilkus27: Changing `line-height` to `-1` didn't change the value, it is still `1.5rem`.

ajluk:
Not the footnotes look like this: 
image

styles/output/ap-biology-pdf.css
ajluk: @wilkus27 after discussion with Larissa I have set back the line-height to line-height(0) (the output css value is the same as it was before (1.5rem)), because the line-height() function doesn't support negative values.

Issue #1521, opened 2019-10-25
Double numbering


MartaSztuka:
Step 1, Step 2 etc. list has double numbering.
![Screenshot from 2019-10-25 13-26-13](https://user-images.githubusercontent.com/47241114/67568062-70d25e00-f72b-11e9-9da9-0ebf7ddfab81.png)

MartaSztuka:
Chemistry AF
![Screenshot from 2019-10-30 11-12-29](https://user-images.githubusercontent.com/47241114/67849121-42c39400-fb06-11e9-8e61-cc78f66cdf72.png)

brittweinstein:
@MartaSztuka can you confirm if this is still an issue.

MartaSztuka:
@brittweinstein yes it is

brittweinstein:
@MartaSztuka could you please provide a link to this? @lwtchu will investigate the class.

MartaSztuka:
![Screenshot from 2020-01-08 10-20-59](https://user-images.githubusercontent.com/47241114/71965894-d8635a00-3200-11ea-8822-f1805dc8a83e.png)

ajluk:
Changes in TDS:
@lwtchu I'm not sure where in the TDS table it should be place, but these styles are common for all books and doesn't have shapes or components. They are placed in the _common-styles.scss file. So maybe something like that can be added to the table as a name.

lwtchu:
@ajluk if it's alright with you i won't put it in the spec doc since it's common to all books.

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-03-10 13-46-10](https://user-images.githubusercontent.com/47241114/76313722-c754c900-62d5-11ea-8584-3fbaafff3100.png)
Chemistry af
![Screenshot from 2020-03-10 13-49-37](https://user-images.githubusercontent.com/47241114/76313833-0551ed00-62d6-11ea-80f0-45447f6f69cb.png)

PR #1990, opened 2020-03-02
unstyle stepwise lists chemistry

ajluk:
Issue: #1521


ajluk:
@brittany-johnson I have added you as a reviewer to make sure hiding the lists style is the right solution in this case. 
now these Examples look like this: 
image

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -284,6 +284,14 @@ $IconNoteAnswerTitleComponent: (
...
   )
 );
+$IconNoteStepwiseOrderedList: (
+  _name: "StepwiseOrderedList",
+  _subselector: " > ol.os-stepwise",
+  _properties: (
+    list-style-type: none,
+  ),
+);
+
```
wilkus27: I'm not sure whether step lists occurs only in Icon Notes. Maybe safer solution is to add this settings to common-styles.scss file? @brittany-johnson what do you think?

brittany-johnson: I just asked Larissa if these occur in other locations. I'll update you once I know.

brittany-johnson: Yes, there will be other locations where stepwise steps will eventually appear, so they will not always be in notes. 
I think that idea will work. @wilkus27

ajluk: @wilkus27 @brittany-johnson I have moved the styles for stepwise lists to common-styles.scss file and it works fine in this case:
image

Issue #1885, opened 2020-02-07
HS Physics collating element with no content still shows header


oscryan:
# Scenario
When a chapter has no content tagged to a collating element, for example the Problems section, the header for it still shows up in the collated material for that chapter.

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [x] Yes  [ ] No  (_attach below_)
<img width="256" alt="Screen Shot 2020-02-07 at 11 27 21 AM" src="https://user-images.githubusercontent.com/1895432/74051096-f6a5ac80-499c-11ea-867c-2271f2dffd35.png">
<img width="261" alt="Screen Shot 2020-02-07 at 11 27 15 AM" src="https://user-images.githubusercontent.com/1895432/74051099-f7d6d980-499c-11ea-848d-fae990cc8e55.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

lwtchu:
Content fix - chunks should have `problems`

lwtchu:
correlating: https://app.zenhub.com/workspaces/content-management-5c86d8c40b66e07993f45cfa/issues/openstax/content-managers/1463

wilkus27:
@openstaxalina @brittweinstein @helenemccarron Marta didn't find wrong classes in Chapter 22 and 23, so problem with empty header still will present. We need a recipe fix as well. 

@helenemccarron I think it will be a huge change. Similar problem occured in Astronomy: https://github.com/openstax/cnx-recipes/issues/658. Then you moved includes of mixins to different passes (https://github.com/openstax/cnx-recipes/pull/719/files). I guess that this problem has to be resolved in similar way. What probably will be very tricky and a lot of testing will be needed.

openstaxalina:
[...]

PR #1997, opened 2020-03-02
move section separated to its own pass

helenemccarron:
issue #1885 

This affects HS physics and University Physics but all books are affected (I changed the pass order).


helenemccarron:
Good catch! Thank you, Natalia, I'll see how I can fix this.

wilkus27:
Since this PR was created long time ago I'm going to close it, because it might be hard to update this branch with all of new changes. I opened new one: https://github.com/openstax/cnx-recipes/pull/2268 and there fixed problem with section links.

Issue #1972, opened 2020-02-27
HS Physics Introduction


MartaSztuka:
Introduction in chapter 23 is cropped.
![Screenshot from 2020-02-27 15-06-48](https://user-images.githubusercontent.com/47241114/75452724-0e0c0000-5973-11ea-9870-4effa9815416.png)

wilkus27:
I think that Content Managers are working on that. @lwtchu can you confirm?

lwtchu:
@wilkus27 in your professional opinion, do you know if changing the image size will fix it?

wilkus27:
@lwtchu I was trying to check that and I think that also coding work is needed. Let me try to fix that by styling first. If also changing image size is needed I will go back to you.

wilkus27:
## Changes to provide to Theme Design Specs

### Shape: `Figure`

_Component:_ `FigureContainer`
_Removed properties:_
```
        display: flex,
        justify-content: center,
        align-items: center,
        flex-direction: column,
```
_Component:_ `CaptionWrapper`
_New properties:_
```
        display: table,
        margin-left: auto,
        margin-right: auto,
```
_Component:_ `PicCaption`
_New properties:_
```
        display: table,
        margin-left: auto,
        margin-right: auto,
```

wilkus27:
Note: I have to change a way of centering figures to fix problem from this issue. @MartaSztuka please test figures generally in different places from the book.

MartaSztuka:
Introduction looks ok.
![Screenshot from 2020-03-09 10-45-38](https://user-images.githubusercontent.com/47241114/76201226-39081680-61f3-11ea-9df3-543e8a331cd9.png)
Figures - centerd, tested in random locations.
![Screenshot from 2020-03-09 10-52-24](https://user-images.githubusercontent.com/47241114/76201880-440f7680-61f4-11ea-8db5-40029055b0a3.png)

PR #2010, opened 2020-03-04
Figures centering

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1972


wilkus27:
Problem from the issue was caused by `display: flex` property. So I center figures using `display: table` property. I have to apply it separately for `figure` tag and for `.os-caption-container`, because width of the caption should be independed of figure width. Sometimes figure is small and the caption is long, and in this cases we don't want such efect:

image

This is how it looks with my solution:

image

This is the same effect which was obtained previously with `display: flex`.

Issue #2025, opened 2020-03-05
Symmetry Feedback: Ch. 10, 16, 23 Outline Boxes Too Long


openstaxalina:
# Scenario

![Screen Shot 2020-03-05 at 11 40 41 AM](https://user-images.githubusercontent.com/7863733/76009065-385f3f80-5ed6-11ea-9cfb-3ba5c126a4d4.png)

![Screen Shot 2020-03-05 at 1 20 23 PM](https://user-images.githubusercontent.com/7863733/76017373-18cf1380-5ee4-11ea-9271-4a2260554724.png)

![Screen Shot 2020-03-05 at 3 39 06 PM](https://user-images.githubusercontent.com/7863733/76028116-82a4e880-5ef7-11ea-9b5c-514e649532d8.png)

# Acceptance Criteria (required)

# Test Plan

lwtchu:
I think this is because of the intro paragraphs. This book might be a good candidate to use option 2 for the chapter outline

lwtchu:
Before we go with that option @helenemccarron there's definitely no way to shrink it right? cause it's basically a table?

wilkus27:
@helenemccarron for explanation, it displays in this way because of my changes from https://github.com/openstax/cnx-recipes/pull/1931/files. 

Previously we had such settings for Chapter Outline: 

![image](https://user-images.githubusercontent.com/40228854/76197879-43271680-61ed-11ea-918d-6ddbd15a0aaf.png)

But then, when intro-text was short and Chapter Outline long, the whole intro-body didn't jump to the second page, but left on the first page and cut Chapter Outline box on the bottom.

What I did is, I moved whole intro-body (which contains intro-text and Chapter Outline):

![image](https://user-images.githubusercontent.com/40228854/76198187-e7a95880-61ed-11ea-8056-0d59c5ae2880.png)

So now it jumps to the second page either if intro-text is long or if Chapter Outline is long. But side effect of this solution is issue described by Alina.

brittweinstein:
Decision: we will go with the full-width chapter outline.

ajluk:
Changes in TDS:
These **FullWidthChapterIntro** shape components properties need to be added as a book specific specs colors to HS Physics:
    **SplashFigureBg** M 126
       background-color: B5
    **OutlineTitle** M.137
      color: B5
    **Outline** M 138
       border-bottom-color': B5
    **OutlineTitleBottomBorder** M139
       border-bottom-color: B5
[...]

PR #2031, opened 2020-03-09
fullwidth chapter outline hsphysics

ajluk:
Issue:  #2025


styles/books/hs-physics/book.scss
```diff
@@ -210,7 +212,7 @@
 ));
 
 @include use('BookRoot', "common:::BookRoot");
-@include use('ChapterIntro', 'carnival:::ChapterIntroShape');
+@include use('FullWidthChapterIntro', 'carnival:::FullWidthChapterIntroShape');
```
wilkus27: @ajluk when shape is changed from ChapterIntro this part should be also removed from this file:

image

ajluk: Thanks Natalia, I just have forgotten about this part!

PR #2033, opened 2020-03-09
Add slice everywhere! 🔪

helenemccarron:


styles/designs/carnival/parts/notes/_notes-boxed-shapes.scss
```diff
@@ -5,6 +5,7 @@
   _groups: (
     group-border-style: solid,
     group-border-width: 0.2rem,
+    group-box-decoration-break: slice,
```
ajluk: @helenemccarron I have a question about this property set in the shape as group property. I didn't see it was used anywhere. Is it set as just a reminder for us to always set it in the components of this shape where we have a border?

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -78,6 +78,7 @@ $BoxedNoteTitleComponent: (
     margin-bottom: v-grid(1),
     border-bottom-style: dotted,
     border-bottom-width: 0.2rem,
+    box-decoration-break: slice,
```
ajluk: See the comment above.

styles/designs/carnival/parts/notes/_notes-boxed-shapes.scss
helenemccarron: No, those are group properties. They are used at the shape level and apply to all the settings that use that shape. Just like in a component, since this one is hardcoded, it doesn't need to be invoked.

PR #2036, opened 2020-03-10
extract baked book/toc meta json

c0ldcalamari:
book_json:
    {
        "031da8d3-b525-429c-80cf-6c8ed997733a" : {
            "abstract": '\n'
             '        <div xmlns="http://www.w3.org/1999/xhtml" '
             'xmlns:c="http://cnx.rice.edu/cnxml" '
             'xmlns:md="http://cnx.rice.edu/mdml" '
             'xmlns:qml="http://cnx.rice.edu/qml/1.0" '
             'xmlns:mod="http://cnx.rice.edu/#moduleIds" '
             'xmlns:bib="http://bibtexml.sf.net/" '
             'xmlns:data="http://dev.w3.org/html5/spec/#custom" '
             'xmlns:epub="http://www.idpf.org/2007/ops" '
             'xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" '
             'xmlns:dc="http://purl.org/dc/elements/1.1/" '
             'xmlns:lrmi="http://lrmi.net/the-specification"/>\n'
             '      \n'
             '      ',
            "legacy_id": "col11406",
            "license": {'code": 'by',
                        "name": 'Creative Commons Attribution License',
                        "url": 'http://creativecommons.org/licenses/by/4.0/',
                        "version": '4.0'},
            "title": 'College Physics'
            "tree": .....
        }
    }

missing abstract - issue CNX#
missing tree - issue CNX#
    "tree": {
        "title": metadata.title,
        "id": "#uuid@version",
        "contents": [],
        "slug": 
    }
    generate slug for book (root of tree)
    title_seq = title_seq + [tree['title']]
    tree['slug'] = generate_slug(*title_seq)
    legacy_id = get_legacy_id_from_uri(metadata.cnx_archive_uri)


script/bake-book
```diff
@@ -79,8 +78,10 @@ for book_config in "${BOOK_CONFIGS[@]}"; do
   do_progress_quiet "Baking ${raw_book} using ${recipe_name} (~1min)" \
     cnx-easybake --coverage-file "${lcov_file}" ${debug_flag1} ${debug_flag2} ${debug_flag3} ${debug_flag4} "${css_file}" "${raw_book}" "${baked_book}"
 
-  do_progress_quiet "Copying metadata from ${raw_meta} to ${baked_meta}" \
-    cp "${raw_meta}" "${baked_meta}"
+  # Generate JSON metadata using venv python assumed to be available via Docker
+  # setup and previous bootstrap
+  do_progress_quiet "Extracting metadata from ${baked_book}" \
```
rnathuji: It looks like there's a `\t` here after the `\` in this line which causes an error when the script runs.

script/bake-book
```diff
@@ -79,8 +78,10 @@ for book_config in "${BOOK_CONFIGS[@]}"; do
   do_progress_quiet "Baking ${raw_book} using ${recipe_name} (~1min)" \
     cnx-easybake --coverage-file "${lcov_file}" ${debug_flag1} ${debug_flag2} ${debug_flag3} ${debug_flag4} "${css_file}" "${raw_book}" "${baked_book}"
 
-  do_progress_quiet "Copying metadata from ${raw_meta} to ${baked_meta}" \
-    cp "${raw_meta}" "${baked_meta}"
+  # Generate JSON metadata using venv python assumed to be available via Docker
+  # setup and previous bootstrap
+  do_progress_quiet "Extracting metadata from ${baked_book}" \	  
+    python ./script/bake-book.py "${raw_meta}" "${baked_meta}" "${baked_book}" "${book_uuid}"
```
rnathuji: Are these arguments passed in the correct order? (based on the code / tests I think it should be `"${raw_meta}" "${baked_book}" "${baked_meta}" "${book_uuid}"`)?

script/bake-book.py
```diff
@@ -0,0 +1,47 @@
...
+        baked_metadata = json.load(raw_json)
+    
+    with open(baked_xhtml_file, "r") as baked_xhtml:
+        html = etree.parse(baked_xhtml)
+        metadata = DocumentMetadataParser(html)
+
+        required_metadata = ('title', 'license_url')
+        for required_data in required_metadata:
+            if getattr(metadata, required_data) is None:
+                raise ValueError("A value for '{}' could not be found.")
```
rnathuji: Do we know how often this might occur, and whether we should treat it as being catastrophic? I believe if we hit this condition and throw an exception it will rollup into a pipeline failure (probably worth confirming) within Concourse, so do we want missing metadata to be "noisy" and fail like that or simply log and insert an empty value (?).

script/bake-book.py
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        baked_metadata[book_uuid] = baked_book_json
+
+    with open(baked_metdata_file, "w") as json_out:
+        json.dump(
+            baked_metadata,
+            json_out
+        )
+        json_out.close()
```
rnathuji: I think you can remove this explicit `close()` due to the context manager

script/disassemble-book.py
```diff
@@ -73,6 +74,11 @@ def main():
 
     with open(baked_metdata_file, "r") as baked_json:
         baked_metadata = json.load(baked_json)
+        book_toc_metadata = baked_metadata.get(book_uuid)
+
+    with open(f"{out_dir}/collection.toc-metadata.json", "w") as toc_json:
+        json.dump(book_toc_metadata, toc_json)
+        toc_json.close()
```
rnathuji: Same comment as in `bake-book.py` re: not needing the explicit `close()`

c0ldcalamari:
this will be overcome by branch against openstax/output-producer-service, because we have separated bakery (script) concerns

branch: https://github.com/openstax/output-producer-service/tree/bn/cnx%23871/extract-baked-book-metadata

Issue #2015, opened 2020-03-05
EOC Issues


brittweinstein:
Main card (#1942) - might want to delete it and replace with this.

- [x]  Remove all the list components, move them to a lists components file, add to shapes. 
**Note**: suggestion on new naming scheme. 
`block`
`block__element`
`block--modifier` 
for example:
`list`
`list__margin`
`list_margin--left`
See https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/ ~~(we can discuss this before implementing)~~ Decision: we will experiment with the naming scheme and re-evaluate in a couple sprints.

- [x]  Remove all titles functions, move them to one component (done, I think), add to shapes. #2039
- [x]  Add a title + image component to handle #1983 and #1978
note: do we need a title components file that can be reused in all shapes? yes, see #2043

- [x]  Remove the named table components and add existing table components to shapes in EOC (this includes exercises etc.) #2040

- [x]   Remove the EOC specific figure components, use the existing figure components, add to shapes #2041

- [x]  Move exercises out of EOC to their own components file, include EOC specific exercises to EOC and everywhere else, make sure all exercises are handled in the appropriate shapes for other parts, use optional where applicable to avoid repetition. - #1870 

- [x]  remove `Multiple_Choice` component and replace with generic component or unique settings if necessary  #2042

- [x]  **Rework exercises columns to display better (`display: table` seems to be wrecking havok): #1956 (high priority)**

- Handle un-clustered EOC sections (probably with two shapes: one `section-separated`, one `unclustered-section` or something) - #2013

wilkus27:
@helenemccarron I like this idea:

![image](https://user-images.githubusercontent.com/40228854/76204747-46280400-61f9-11ea-8b30-afc17b8b0f82.png)

I think that then we can remove all files related with `chapter` and `module` which contain just titles. I mean those files: https://github.com/openstax/cnx-recipes/blob/master/styles/designs/carnival/parts/_chapter-components.scss and https://github.com/openstax/cnx-recipes/blob/master/styles/designs/carnival/parts/_module-components.scss and of course related with them `-shapes.scss` and `-settings.scss`.

brittany-johnson:
All issues in this epic have been completed, closing.

Issue #1862, opened 2020-02-04
CLEAN UP: create one reusable component for all lists spacing


helenemccarron:
There should be one component to unify display of all lists instead of many named components. 

This PR contains a lot of the work done on lists: https://github.com/openstax/cnx-recipes/pull/1697

ajluk:
EOC Issues https://github.com/openstax/cnx-recipes/issues/2015

ajluk:
Changes in TDS:
In the **AnswerKey** shape:
The **OrderedList** ,**UnorderedList** components need to be replaced with one component:
**AnswerKey__List--MarginBottomLeft**
with properties:
margin-left: h-grid(3);
margin-bottom: 0;

In the **Appendix** shape:
The **UnorderedList**, **OrderedList**, **SectionOrderedList** components need to be replaced with one component:
**List--MarginLeft**
with properties:
margin-left: h-grid(3);

ajluk:
@MartaSztuka here is the link to the document for testing purposes, where you can spot places that need to be checked shown on screenshots for College Physics: https://docs.google.com/document/d/1cry7GPpIZ00sCudMtYExkw3Y6vEH8NjfiVGUT8DgLCk/edit#heading=h.n1yzk266rntj

MartaSztuka:
**College Physics**
![Screenshot from 2020-03-26 11-05-48](https://user-images.githubusercontent.com/47241114/77634773-c5d00580-6f51-11ea-89ab-302e7a633bb0.png)
![Screenshot from 2020-03-26 11-12-42](https://user-images.githubusercontent.com/47241114/77635362-be5d2c00-6f52-11ea-8820-6e123a12f5a5.png)
![Screenshot from 2020-03-26 11-13-43](https://user-images.githubusercontent.com/47241114/77635447-e2207200-6f52-11ea-86e3-81f64a850c7c.png)
![Screenshot from 2020-03-26 11-14-43](https://user-images.githubusercontent.com/47241114/77635555-0d0ac600-6f53-11ea-96be-b6516015f482.png)
![Screenshot from 2020-03-26 11-16-31](https://user-images.githubusercontent.com/47241114/77635709-45120900-6f53-11ea-922d-58b333457974.png)
![Screenshot from 2020-03-26 11-17-09](https://user-images.githubusercontent.com/47241114/77635773-5e1aba00-6f53-11ea-9e84-c7cb2c110917.png)
[...]

PR #2044, opened 2020-03-11
clean lists 

ajluk:
Issue: #1862
Related epic issue: #2015


styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -0,0 +1,130 @@
...
+        margin-bottom: enum('ValueSet:::OPTIONAL'),
+    )
+);
+$EoC_Exercise__Section__List--Ordered: (
+    _name: "EoCExerciseSectionListOrdered",
+    _subselector: " section > ol",
+    _properties:(
+        margin-left: enum('ValueSet:::OPTIONAL'),
+        margin-bottom: enum('ValueSet:::OPTIONAL'),
+    )
```
ajluk: @helenemccarron I have used more general selector for these lists in EoC exercises present in HSPhysics, because I thought the nesting for them is too deep and otherwise I will need to create 3 components with empty properties just for this nesting (it occurs only in one place in the book) or use long nested selector, but let me know if it is the better way to do it: 
image

styles/templates/template1/parts/_eoc-settings.scss
```diff
@@ -682,32 +682,32 @@
...
             margin-bottom: v-grid(1),
         ),
-        ExerciseSectionUnorderedList: (
+        EoCExerciseSectionListUnordered: (
             margin-left: h-grid(4),
             margin-bottom: v-grid(1),
         ),
-        ExerciseSectionOrderedList: (
+        EoCExerciseSectionListOrdered: (
             margin-left: h-grid(4),
```
ajluk: After dicussion with Larissa I have set the left margins for lists in the EoC exercises the same value as the rest of the book, except this two types of lists, where they are nested in sections and I needed to increase the margin for them to look the same as other lists in exercises: 
image

styles/designs/carnival/parts/_lists-components.scss
helenemccarron: @ajluk It still looks like there is a lot of repetition here. Looking at al the components, I can identify two patterns:  `margin-left: h-grid(2),` and `  margin-left: enum('ValueSet:::OPTIONAL'), margin-bottom: enum('ValueSet:::OPTIONAL'),`

To me, it would make more sense to create only those two components and pass the classes in the parent or create a function and pass the class as parameter.

ajluk: @helenemccarron I have chosen the second option you proposed and used one function that covers different variations of the lists selectors and properties present in the books parts, because I thought it will be more efficient. But let me know what do you think about that solution. Thank you in advance!

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -0,0 +1,86 @@
...
+    $margin_bottom_value: null;
+
+    @if $element_type == main {
+        $name: "ListMarginLeft";
+        $margin_left_value: h-grid(3);
+    } @else if $element_type == nested {
+        $name: "ListNestedMarginLeft";
+        $margin_left_value: h-grid(2);
+    } @else if $element_type == exercise {
+        $name: "ExerciseListMarginLeftBottom";
```
helenemccarron: `ExerciseListMarginLeftBottom`, 
`ExerciseNestedListMarginLeftBottom`, 
`ExerciseNestedListMarginLeftBottom`, 
`exercise-with-increased-margin` are exactly the same and optional, is it really necessary to create condition for those?

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -0,0 +1,86 @@
...
+    @return (
+        _name: "#{$name}",
+        _subselector: "#{$sub_selector}",
+        _properties:(
+            margin-left: $margin_left_value,
+            margin-bottom: $margin_bottom_value,
+        )
+    );
+}
+
```
helenemccarron: After reading those, I wonder if making two mixins wouldn't make more sense:
`list--marginLeftBottom` which would contain both margins and `list--noMarginBottom` would be more concise. Then pass the margins as a parameter per shape rather than having to maintain so many unique cases.

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -0,0 +1,86 @@
...
+// General Prefece Modules Summary
+
+$List_Component--Unordered: styleList(main, " > ul");
+$List_Component--Ordered: styleList(main, " > ol");
+$Section__List_Component--Unordered: styleList(main, " > section > ul");
```
helenemccarron: I think for the naming convention, I would use CamelCase for inner parts of the name: 
`Section__ListComponent--Unordered`

styles/output/anatomy-pdf.css
```diff
@@ -634,13 +634,22 @@ nav#toc > .os-toc-title {
 .preface > ul {
   margin-left: 24px; }
 
-.preface section > ul {
+.preface > section > ul {
```
helenemccarron: It looks like the output is quite different from the original, is that intended? 

In general, when refactoring code, it's a good idea not to change any functionalities, so your output should be the same. 

If it cannot be avoided,  please list all the changes and how to test them, everything in the CSS files should be accounted for before sending this to testing.

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -0,0 +1,86 @@
...
+$Sub_Sub_Section__List_Component--Unordered: styleList(main, " > section > section > section > ul");
+$Nested__List_Component--Unordered: styleList(nested, " > li > ul");
+$Nested__List_Component--Ordered: styleList(nested, " > li > ol");
+$SectionSummary__List_Component--Unordered: styleList(main, " .section-summary > ul");
+$SectionSummary__List_Component--Ordered: styleList(main, " .section-summary > ol");
+$Summary__List_Component--Unordered: styleList(main, " .summary > ul");
+$Summary__List_Component--Ordered: styleList(main, " .summary > ol");
+$Summary__Section__List--Unordered: styleList(main, " section > ul");
+
+// Exercices
```
helenemccarron: Love the comments! You might even want to add the generic rule for those (for example, "exercises in columns need more spacing at the bottom") which would help us quickly identify the rules and why we are applying them

styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -99,48 +99,6 @@ $Exercisetitle_Container: eocSectionTitleH3(".exercises");
 $ReviewQuestionsTitle_Container: eocSectionTitleH3(".os-review-questions-container");
 // $FillInTheBlankTitle_Container: eocSectionTitleH3(".os-fill-in-the-blank-container");
 
-$SectionSummaryUnorderedList:(
```
helenemccarron: YEEES! :D

styles/designs/carnival/parts/_lists-components.scss
ajluk: Yes, because the name for each of these components type should be different in this case, when they are used in the same shape in the _eoc-settings.scss file and have different values there.

styles/output/anatomy-pdf.css
ajluk: Yes, it was intended. At first I wanted to nest everything deeper to have the smallest possible number of lists components and I have added in the preface also support for sections nested in one another, where the lists are placed here to avoid the general selector: 
image

styles/designs/carnival/parts/_lists-components.scss
ajluk: @helenemccarron I have created the second function as you suggested above(I left functions because I couldn't include a mixins in components properties) and reduced the number of cases (I have set the values for margins in exercises lists as hardcoced to avoid creating separate case in the second function, so now all exercises which are using a particular shape with lists will have set the style for them in advance, what do you think about this solution I guess we talked about this on one of the calls?). I have also used the same selector to reduce number of cases for summary where it was possible. The name convention has been changed as well and as you said it generates warnings and I can be left for now (I think the styles are only build a little longer now, should it be kept?).

PR #2056, opened 2020-03-17
Carnival100

ajluk:
Milestone branch


wilkus27:
To be able to test changes from this branch we have to apply new shapes and components to all books (other then biology 2e, hs physics, college physics). At this moment there are errors and files are not generated properly.

Issue #2061, opened 2020-03-18
Additional chapter label in Answer Key


ajluk:
In the Answer Key in University Physics additional chapter label occured. This issue regards also other books that have Answer Keys: Microbiology, Chemistry 2e.
Example form University Physics vol1 xhtml file: 
![image](https://user-images.githubusercontent.com/55203598/76976547-657b0b80-6934-11ea-873b-5ddc9a4a6e11.png)

The section headers from Check Your Understanding solutions need to be removed as well.

yblock:
Duplicate header in table of contents as well: "UNIT 1 UNIT 1"

![Screen Shot 2020-03-18 at 10.44.02 AM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/cb6937c3-4c27-4a8f-a9c9-e6bd08965331)

ajluk:
@yblock thank you! Do you know when that has happened?

yblock:
I'll make a card for CMs to remove hard-coded unit numbers

ajluk:
How it looks now: 
![image](https://user-images.githubusercontent.com/55203598/76982193-d3770100-693b-11ea-8fe7-fab9fcdd2fa6.png)

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-04-06 16-32-22](https://user-images.githubusercontent.com/47241114/78569649-34e80c80-7824-11ea-863c-0801af24e41f.png)
Microbiology
![Screenshot from 2020-04-06 17-23-37](https://user-images.githubusercontent.com/47241114/78575347-82b44300-782b-11ea-88f6-790d1d93bd94.png)
University Physics
![Screenshot from 2020-04-06 17-22-44](https://user-images.githubusercontent.com/47241114/78575380-91025f00-782b-11ea-894f-ce252b5b379b.png)

PR #2062, opened 2020-03-18
remove additional chapter label section headers answerkey

ajluk:
Issue: #2061
This PR will resolve also issue #2064


ajluk:
Now it looks like this (University Physics vol1 PDF file): 
image

recipes/mixins/_compose.scss
```diff
@@ -684,19 +684,20 @@
...
             }
-            &::after {
-              content: nodes($sectionHeaderNode);
-              move-to: moduleTitleName;
-            }
+            //Left commented just in case the section headers will be needed in the future
+            // &::after {
+            //   content: nodes($sectionHeaderNode);
+            //   move-to: moduleTitleName;
+            // }
```
wilkus27: Why is this removed? It adds section titles.

ajluk: In the old pdf form opentstax.org  the sections aren't present. I have asked Larissa about it and she said it is ok.

wilkus27: Ok, but this change removed also section titles from Precalculus where we want them.

wilkus27:
@ajluk I dive into it more carefully. Here is how mixin `compose_createEOBSolutions` works.

In config `$Config_BookCompositePages` we can choose if solutions are clustered by chapter or section. In the first case (e.g applied to Chemistry) it looks like that:

image

All solutions from the whole chapter are placed next to each other.

In the second case we have two results. First:

image

There are added titles of the EOC sections. And regular section titles in answers from section notes.  And they are added by code which you removed.

I checked why they are needed in Precalculus. So in this book numbering of the notes resets by module:

image

That's why we need those headers to know from which module are they. 
So definitelly this condition from the mixin is wrong:

image

(file: `_compose.scss`)

I think we can use this information:

image

(file: config.scss for Precalculus)

I mean, that we can create condition that those titles will be added in case when key `resetAt` in `$Config_Notes` is equal to `moduleReset`.

Please let me know if I should explain something more precisely.

ajluk:
@wilkus27 I have added the suggested changes and the code is ready for review. Thanks for the idea!

styles/designs/carnival/parts/_answerkey-components.scss
```diff
@@ -70,6 +70,19 @@ $AnswerKeyExerciseTitleTextComponent: (
...
+  _properties: (
+    color: enum('ValueSet:::REQUIRED'),
+    font-family: enum('ValueSet:::REQUIRED'),
+    font-size: font-scale(3),
+    line-height: line-height(3),
+    font-weight: bold,
+  )
+);
+
+
```
wilkus27: I think we don't need this component now. 

image

This element has now `os-solution-area` class like other elements.

ajluk: Thanks @wilkus27! I have missed that! The component has been removed.

brittweinstein:
@ajluk we can merge w/o BrittJ's review since Alan has approved :)

PR #2074, opened 2020-03-19
Space saving

helenemccarron:


Stackblocks:
# Summary
From a QA standpoint I noticed two things when comparing the space-saver version to the version currently on master: 

1. There are a number of places where titles collide with or severely crowd the elements that follow them now that the spacing has been reduced, this may be resolved by the title work you were discussing this morning.

2. The font scale of captions has been reduced, but it looks like the line height is still fairly substantial. This could be trimmed down to make space for the titles to have a little more breathing room.

# Examples

## Collisions / Crowding
image
---
image
---
image
---
image

## Caption Line Spacing
image
---
image

ajluk:
@helenemccarron I have noticed that there are some book specific color settings for H2 and H3 Titles missing in the book.scss files for all of the books. The shapes that are affected are mainly form EoC (KeyTerms, Summary, Exercises) and EoB (AnswerKey, Appendices) parts. Examples:
HSPhysics:
image
image
College Physics:
image
Biology:
image
image

styles/output/biology-pdf.css
```diff
@@ -668,18 +669,18 @@ nav#toc > .os-toc-title {
...
   column-span: all;
-  margin-bottom: 1.5rem; }
+  margin-bottom: 0.7rem; }
 
 [data-type='chapter'] > .os-glossary-container > h3 {
   font-family: IBM Plex Sans, sans-serif;
   font-weight: bold;
   font-size: 1.728rem;
-  line-height: 3rem;
+  line-height: 1.4rem;
```
wilkus27: @helenemccarron I have one question, is it intentional that sometimes line-height is smaller then font-size? Like here: `line-height: 1.4rem` and `font-size: 1.728rem`.

helenemccarron:
@wilkus27 Yes, I went over some of those this morning, I will wrap them all up with your changes. They are so many unique components, it's a mess. Thanks for listing all of them, I will use it make sure I don't miss any.

helenemccarron:
@Stackblocks Yes, for the TOC, I've already fixed it (I'll double check), for the crashing text, I should have `v-spacing(1)` under all titles, I'll double check it's applied. For the captions, I know, the grid isn't allowing me to fix it (I am still working on it)

styles/output/biology-pdf.css
helenemccarron: Yes, that is to reduce the problem of very spread out titles. I can add a comment to it.

styles/output/hs-physics-pdf.css
```diff
@@ -1568,139 +1462,130 @@ iframe {
...
 
-[data-type='chapter'] > .os-summary-container a:first-of-type > h3 {
-  font-family: Muli;
+[data-type='chapter'] > .os-summary-container > section > a > h3[data-type=document-title] {
+  font-family: IBM Plex Sans, sans-serif;
   font-size: 1.44rem;
-  line-height: 1.5rem; }
+  line-height: 1.4rem;
+  margin-bottom: 0.7rem;
+  font-weight: bold; }
```
wilkus27: Previously section links were Muli and unbolded. Is it intentional change?

styles/output/hs-physics-pdf.css
```diff
@@ -1714,283 +1599,1026 @@ iframe {
...
-
-.os-problems-container {
-  margin-bottom: 3rem; }
+  line-height: 1.4rem;
+  margin-bottom: 0.7rem;
+  font-weight: bold; }
 
-.os-problems-container > h2 {
-  color: #00819A;
+.os-concept-container > section > a > h3[data-type=document-title] {
```
wilkus27: This selector doesn't work for compound composite pages. 
Here is the markup from HS Physics:

image

styles/books/biology/book.scss
```diff
@@ -138,25 +115,111 @@
...
-@include use('SummaryWithSectionLinks','carnival:::EocSummaryWithSectionLink');
-@include use('VisualExercise', 'carnival:::EocTwoColumnsUnclustered');
-@include use('MultipleChoice', 'carnival:::EocTwoColumnsUnclustered');
-@include use('CriticalThinking', 'carnival:::EocTwoColumnsUnclustered');
-@include use('RegularExercise','carnival:::ExerciseShape');
+@include use('EocSummary','carnival:::EocSummary');
+@include use('VisualExercise', 'carnival:::ExerciseShape');
+@include use('MultipleChoice', 'carnival:::ExerciseShape');
+@include use('CriticalThinking', 'carnival:::ExerciseShape');
+
```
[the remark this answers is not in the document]
helenemccarron: Will do

styles/output/hs-physics-pdf.css
helenemccarron: Nope, I'll restore it.

styles/output/hs-physics-pdf.css
helenemccarron: Thanks, I'll update it.

ajluk:
@helenemccarron I have checked how lists arew working after your changes and found few spots were they aren't covered in these 3 priority books:

**HS Physics:**

- ul lists in Performance Task exercises
image

- an empty spot under one section link and there the ul nested list isn't indented as well: 
image

- ul nested deeper unstyled
image

**College Physics**

- the icon in Check Understanding Notes became smaller, was it intentional?
image

- ul list in section-summary remain unstyled
image

wilkus27:
@ajluk about Hs Physics it could be related with https://github.com/openstax/content-managers/issues/1556. Whole content from those section should be inside `[data-type=exercise"]` and as Larissa said probably it is done on production, so we should check if this problem occurs on production content.

Stackblocks:
@helenemccarron The only issues I saw vis a vis spacing were collisions and near-misses in the following circumstances:

### Punctuation descenders collide with over-tall risers
image

Not sure there's a ton we can do for this short of changing the font or altering the title to something with a level top-line, both of which it's too late to do for this run.

### Descenders in link-to-learning URLs colliding with note-end line
image

Theoretically some more space can be added to prevent this.

### Something that irks me personally but if UX says it's fine it's fine
image

image

Spacing between the chapter number and chapter title in the text sometimes gets down to an individual pixel of spacing and that's pretty bad for accessibility. If this is OK for the current release / print run, I'd like to request that we clean it up before the next one.

ajluk:
@helenemccarron I have question about tiles in notes (Snal Lab, Virtual Physics) that using Boxed Note Shape in HS Physics. I havn't catch it earlier. Aren't they too small?: 
image
image

helenemccarron:
@ajluk that's expected, we changed it per larissa https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/1989 (see the issue about notes)

Issue #2013, opened 2020-03-05
Sections without section links in exercises have inherited margins


ajluk:
In Microbiology there are various exercises in the EoC nested in sections without section links, which looks oddly (generate unnecessary spaces in this case). I have created a separate issue about because the removing of this margins can affect other exercises with section links in other books. Now it looks like this: 
![image](https://user-images.githubusercontent.com/55203598/75963719-4ad17d00-5ec6-11ea-941d-32d1df004684.png)

MartaSztuka:
![Screenshot from 2020-04-14 13-44-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/16db3f3b-5e53-41fe-b977-02c6367be125)

PR #2078, opened 2020-03-23
remove margin bottom sections without section links exercises microbi…

ajluk:
Issue:  #2013


styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -151,6 +151,15 @@ $Eoc_SectionTitleLink:(
         font-family: enum('ValueSet:::REQUIRED'),
         font-size: font-scale(2),
         line-height: line-height(2),
+        padding-top: enum('ValueSet:::OPTIONAL'),
```
helenemccarron: I think we want to remove this.

styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -151,6 +151,15 @@ $Eoc_SectionTitleLink:(
...
+    )
+);
+
+$Eoc_SectionWithoutSectionLink:(
+    _name: "EocSectionWithoutSectionLink",
```
helenemccarron: After spending some time in the EOC, I found that those components are hard to maintain (it's almost impossible to remember what they do and which shape should or shouldn't be using them. I'd like to have more robust components that would contain those options (with or without links etc.) rather than a generic component `EocSectionWithoutSectionLink` that does a very specific thing (reset the margin to zero on one type of section). At the very least, can we rename this component to `noMarginOnEmptySections` or something of that nature? 
Another option would be to make to make a generic component  `noMarginBottom` and use that instead. We might want to have a component files containing those generic components that can be used in an ad hoc manner in shapes.

ajluk: @helenemccarron Thank you for suggestions! I have crated generic components file and rename the component. Let me know if I understood you correctly.

styles/designs/carnival/parts/_generic-components.scss
```diff
@@ -0,0 +1,7 @@
+$SectionComponent--NoMarginBottom:(
+    _name: "Section--NoMarginBottom",
+    _subselector: " section",
+    _properties:(
+        margin-bottom: enum('ValueSet:::OPTIONAL'),
```
helenemccarron: I wonder if the values should be hardcoded. The component itself is used only when needed, I don't think we need to make it optional.

ajluk: It is possible but only if there will be a specific class added to these types of sections (which don't have title links as a children, but it will need a recipes change). In the otherwise because the selector applies to all sections in the eoc exercises,  all of them will have bottom-margin removed which doesn't look good in case of sections that have tile links: 
image
Maybe a better solution could be to set the value type to REQUIRED and set it on book specific files to the value we need in the different exercises? But let me know what do you think @helenemccarron?

ajluk:
@helenemccarron I have created new PR including these changes to avoid resolving many conflicts. Here is the new one: https://github.com/openstax/cnx-recipes/pull/2165

Issue #2084, opened 2020-03-24
Check your understanding


MartaSztuka:
@helenemccarron @brittweinstein @lwtchu 
1.Check mark in note is noticeably smaller than in other notes.
2. Should `Solution` header be this big?
![Screenshot from 2020-03-24 11-57-06](https://user-images.githubusercontent.com/47241114/77418481-3bf43100-6dc7-11ea-80cb-5ad9ed9f6659.png)

helenemccarron:
I'll take care of this today

brittweinstein:
This is functioning as it should be.

brittweinstein:
@helenemccarron is taking a look at the "Solution" size.

brittweinstein:
Spec: font-scale 0 line height 0 font-family IBM plex

helenemccarron:
Condensed notes are now one line tall (per #2032 ), and Link to Learning is the only condensed note so it's smaller than  icon notes (2 lines).

Fixed solutions in Link to learning
![2020-03-25_08-27-47](https://user-images.githubusercontent.com/8514591/77541327-87afe480-6e72-11ea-8475-f00aeb8d86c4.png)

MartaSztuka:
tested
![Screenshot from 2020-03-25 16-41-05](https://user-images.githubusercontent.com/47241114/77555376-789f5580-6eb7-11ea-86ac-ffb5b0a2044c.png)

Issue #2082, opened 2020-03-24
Example headers


MartaSztuka:
@helenemccarron @brittweinstein @lwtchu In examples headers for `Solution`, `Strategy` are not consistent.
![Screenshot from 2020-03-24 11-36-05](https://user-images.githubusercontent.com/47241114/77416033-ab682180-6dc3-11ea-9c40-f6b13efa7bb4.png)

helenemccarron:
I'll take care of this today

brittweinstein:
It is possible that this is from outdated content. @helenemccarron will fetch the most recent data and confirm if this is an issue or not.

helenemccarron:
The udpated content fixed a lot of those, I fixed a bug that made them display at font-scale(1) rather than font-scale(0)
![2020-03-24_15-35-46](https://user-images.githubusercontent.com/8514591/77474504-3c9bc000-6de5-11ea-8062-2afb41ed97a3.png)

MartaSztuka:
It's fixed in newest version.
![Screenshot from 2020-03-25 14-34-50](https://user-images.githubusercontent.com/47241114/77542050-e3e02c00-6ea5-11ea-8d08-cb56f884bbf8.png)

PR #2090, opened 2020-03-24
Regressions

helenemccarron:
issue #2084
 issue #2082
convert the preface to two columns


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -1,7 +1,7 @@
 //common components
 $NoteSubtitleComponent: (
   _name: "Subtitle",
-  _subselector: " > .os-subtitle",
+  _subselector: " > .os-subtitle, h4[data-type="title"]",
   _properties: (
     font-size: font-scale(1),
     line-height: line-height(1),
```
wilkus27: @helenemccarron I'm wondering if `font-family` should be also REQUIRED for this component. It's the first subtitle in the note:

image

At this moment font-family is inherited from the root and it is a Alegreya. I'm not sure if should be Muli or IBM Plex Sans.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -229,11 +229,10 @@ $IconNoteMainTitleIconComponent: (
 
 $IconNoteParagraphTitleComponent: (
   _name: "ParagraphSubtitle",
```
wilkus27: @helenemccarron I noticed that previously those titles has also property `font-weight: bold`. Do we want to add it or it was removed because bolded title doesn't look good with smaller font size?

helenemccarron: good catch, I'll add it backt

styles/designs/carnival/parts/notes/_notes-components.scss
helenemccarron: mmh, I'll take a look. I want to rework all those titles, potentially out of the notes and into the titles, so I'll do the minimum to make it work, we can revisit later

helenemccarron: I ended up hard coding it, there too many places to fix at this time

wilkus27:
@helenemccarron I reviewed also last commit. Everything look fine :)

helenemccarron:
thank you!

Issue #1741, opened 2020-01-06
Features have no distinguishing formatting


oscryan:
# Scenario
Concepts of Biology features just look like black headers

# Acceptance Criteria (required)

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?**
* **Images or Screenshots Included:** [ ] Yes  [ ] No  (_attach below_)

<img width="584" alt="Screen Shot 2020-01-06 at 2 13 08 PM" src="https://user-images.githubusercontent.com/1895432/71845822-c661aa00-308e-11ea-9db6-871dabcd90fb.png">

# Environment: (required)
* **Medium**: [ ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ ] staging  [ ] dev  [ ] content
* **OS**: [ ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

wilkus27:
It displays in this way because in Biology 2e such note is styled:

![image](https://user-images.githubusercontent.com/40228854/75143774-f4aa5000-56f5-11ea-9a7b-187984148cf1.png)

But in Concepts Biology note from this issue has `non majors` class:

![image](https://user-images.githubusercontent.com/40228854/75143893-36d39180-56f6-11ea-80f4-590bb578224f.png)

@lwtchu what should be done in this case? Are there different specs for notes with class interactive **and** non-major?

wilkus27:
By Larissa: it will be created new design for this note.

lwtchu:
@wilkus27 we can add this to the biology template 

Design: please use condensed icon note (link to learning) 
Title: Concepts in Action

MartaSztuka:
tested on carnival100 branch![Screenshot from 2020-04-02 13-58-11.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7f7b264b-caf7-4a87-b845-6093546827dd)

Issue #2032, opened 2020-03-09
space saving: note/example styling adjustments for space saving


lwtchu:
# Icon note AND example
<img width="828" alt="Screen Shot 2020-03-09 at 11 35 43 AM" src="https://user-images.githubusercontent.com/11299125/76236125-352bc280-61fa-11ea-8af5-665b67ea2765.png">

size: KEEP font-scale(1) 
line height: KEEP font-scale(0)
**make the text one line**
**icon size: try v-grid(2)?**

# Icon condensed note
<img width="373" alt="Screen Shot 2020-03-09 at 2 21 09 PM" src="https://user-images.githubusercontent.com/11299125/76249441-53e98380-6211-11ea-8f3b-94e64bd41560.png">

size: KEEP font-scale(1)
**make the text one line**
**icon size: v-grid(1)?**

# Boxed note
<img width="1032" alt="Screen Shot 2020-03-09 at 2 26 16 PM" src="https://user-images.githubusercontent.com/11299125/76249727-f6a20200-6211-11ea-9f5c-9d6a8552eba6.png">

**font-size: font-scale(1),**
**line-height: line-height(1),**

**Try removing the dotted line also**

This should impact all books in the carnival theme, but I'm putting the labels for the high prioriy books first.

lwtchu:
Updated card to include a test to remove the dotted line in boxed notes

MartaSztuka:
**College Physics**
![Screenshot from 2020-03-24 15-48-27](https://user-images.githubusercontent.com/47241114/77439184-04957c80-6de7-11ea-9e11-32c0fb38766c.png)

PR #2098, opened 2020-03-25
Style note from concepts biology

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1741


styles/books/biology/book.scss
```diff
@@ -122,6 +122,13 @@
     'Icon:::background': url(toDataUri("svg+xml", "designs/carnival/resources/carnival-linktolearning.svg")) no-repeat top left,
     'TitleText:::max-width': h-grid(10),
   ),
+  ConceptsInActionNote:(
+    _selectors: (".interactive.non-majors"),
+    'Body:::border-bottom-color': (_ref: 'colorMap:::linkToLearningColor'),
+    'Title:::font-family': (_ref: 'typography:::titleFont'),
+    'Icon:::background': url(toDataUri("svg+xml", "designs/carnival/resources/carnival-linktolearning.svg")) no-repeat top left,
+    'TitleText:::max-width': h-grid(10),
```
wilkus27: It looks like that now:

image

When we update code with changes from space-saving branch icon will be smaller and `'TitleText:::max-width': h-grid(10),` has to be removed, so all words from title will be inline.

helenemccarron: per #2032, please remove the max-width

brittweinstein:
@wilkus27  we can merge w/o Helene's review since Alan has approved :)

PR #2106, opened 2020-03-25
Issue-1869 -- broaden fake top-title styling to all carnival books

Stackblocks:
- Extended styling to the remaining carnival titles.
- Cleaned up extraneous entries in the color map, since titles in a sequence use the same coloration.
- Miscellaneous blank-line tidying-up.


styles/books/ap-biology/book.scss
```diff
@@ -3,8 +3,15 @@
...
-    'FirstTableRow:::border-top-color': (_ref: "colorMap:::apbiologyTableColor"),
+    'LastTableRowData:::border-bottom-color': (_ref: "colorMap:::biologyTableColor"),
+    'FirstTableRow:::border-top-color': (_ref: "colorMap:::biologyTableColor"),
+  ),
+));
+@include add_settings((
+  FakeTitleTable: (
+    _selectors: ('.os-table'),
+    'TableTitle:::color': (_ref: "colorMap:::biologyTableColor"),
+    'LastTableRowData:::border-bottom-color': (_ref: "colorMap:::biologyTableColor"),
```
ajluk: I think that if AP Biology is using Biology book.scss file there is no need to add the FakeTitleTable shape here as well.

styles/books/ap-biology/book.scss
```diff
@@ -61,4 +68,6 @@
 @include use('TestPrep', 'carnival:::EocExerciseWithoutSubtitle');
 @include use('SciencePractice', 'carnival:::EocExerciseWithoutSubtitle');
 @include use('LearningObjectiveParagraphShape', 'carnival:::LearningObjectiveParagraphShape');
-@include use('ScientificMethodNote','carnival:::IconNoteShape');
\ No newline at end of file
+@include use('ScientificMethodNote','carnival:::IconNoteShape');
```
ajluk: This @include use is doubled here.

styles/templates/template1/_colors.scss
```diff
@@ -95,10 +95,8 @@
     microbioDiseaseProfile: (_ref: "scheme:::G7"),
     anatomyTableColor: (_ref: "scheme:::B5"),
     chemistryTableColor: (_ref: "scheme:::B4"),
-    biologyTableColor: (_ref: "scheme:::B7"), 
+    biologyTableColor: (_ref: "scheme:::B7"),
     physicsTableColor: (_ref: "scheme:::B5"),
-    hsphysicsTableColor: (_ref: "scheme:::B5"),
-    apbiologyTableColor: (_ref: "scheme:::G7"),
```
ajluk: This color was used for AP Biology in Column Header Tables only (apbiologyTableColor: (_ref: "scheme:::G7"),). Is there a reason why is it removed? Now this color (biologyTableColor: (_ref: "scheme:::B7")) will be used by two table shapes in AP Biology: Column Header Table shape and Fake Title Table shape. Is the changing color intentional?

styles/books/ap-biology/book.scss
Stackblocks: That makes sense. I can remove it here.

styles/templates/template1/_colors.scss
Stackblocks: The color removal was intentional. I had added the AP Biology and HS Physics as part of my previous work on styling this. 

- The G7 was an error on my part in looking up which of the colors were used in Biology, it should have been B7.
- Because both the hsphysicsTableColor and physicsTableColor are the same I removed hsphysicsTableColor because it was a needless duplicate. Same for apbiologyTableColor, which should have matched biologyTableColor.

styles/books/ap-biology/book.scss
Stackblocks: Thanks! I'll remove the extra.

styles/templates/template1/_colors.scss
ajluk: @Stackblocks Thank you for the explanation!

Issue #2068, opened 2020-03-19
answer key display


yblock:
# What happened?

Some answer key sections such as "Conceptual Questions" have each answer in-line next to the last. This make is hard to use. In the old pdf, each answer is on a new line. 

# Reproducibility Details

- **Does it also happen in an incognito window?**: 
- **Does this happen in one place, a few places, or everywhere?**: Answer Key sections 

# Environment:

- **Book(s)**: University Physics
- **URL/Server**: https://ce-pdf-spike.s3.amazonaws.com/col12031-latest-production-242.pdf
- **Location (page number, chapter/section)**: Answer Key 
- **Suggested nearby searchable text**: "No, neither of these two theories is more"
- **Browser**: PDF

# Screenshots

![Screen Shot 2020-03-19 at 9.08.05 AM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/89e8bf8b-8667-4faa-bbb2-bcbbcae6dcb0)

# Additional notes

This is from a PDF generated with COPS and the u-physics style on March 18

openstaxalina:
@lwtchu could you confirm you'd like each answer to start on a new line?  
(This is how it was in oer.exports:)
<img width="855" alt="Screen Shot 2020-03-19 at 10 03 16 AM" src="https://user-images.githubusercontent.com/7863733/77081654-f3083b00-69c8-11ea-82d0-7bdbf8c2bb00.png">

lwtchu:
Yes they should start on a new line for easier reading

wilkus27:
Will be resolved with https://github.com/openstax/cnx-recipes/pull/2107

MartaSztuka:
University physics
![Screenshot from 2020-04-16 10-42-14](https://user-images.githubusercontent.com/47241114/79434677-0231d700-7fcf-11ea-8cc0-e7335d5f18b1.png)
![Screenshot from 2020-04-16 10-43-21](https://user-images.githubusercontent.com/47241114/79434752-1d044b80-7fcf-11ea-824d-7e20d1451414.png)
![Screenshot from 2020-04-16 10-43-45](https://user-images.githubusercontent.com/47241114/79434793-2b526780-7fcf-11ea-92d4-f0300e8f94a4.png)

Issue #2064, opened 2020-03-18
answer key check your understanding issues headers


yblock:
# What happened?

There is odd section number bolding which makes these sections look very confusing. please see screenshot

# Reproducibility Details

- **Does it also happen in an incognito window?**: 
- **Does this happen in one place, a few places, or everywhere?**: all chapter's answer key check your understanding sections

# Environment:

- **Book(s)**: University Physics
- **URL/Server**: https://ce-pdf-spike.s3.amazonaws.com/col12031-latest-production-242.pdf
- **Location (page number, chapter/section)**: Answer Key
- **Suggested nearby searchable text**: "It is probably an underestimate because the density"
- **Browser**: pdf

# Screenshots

![Screen Shot 2020-03-18 at 3.03.56 PM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/38f6fe63-5d51-4d81-a4f2-7f077deca74b)

# Additional notes

Acceptance Criteria: 
1.) font-weight (unbolded)  
2.) spacing 
3.) removal of module titles

 (_For example, steps to reproduce, related issues, more context, etc._)

openstaxalina:
Extra chapter header will be fixed with https://github.com/openstax/cnx-recipes/issues/2061.

I am adjusting this issue to only be about the section bolding.

openstaxalina:
@lwtchu how would you like the section headers, such as 2.1 Scalars and Vectors, styled?

lwtchu:
Section headers should be removed from University PHysics.

ajluk:
This issue has been resolved by https://github.com/openstax/cnx-recipes/pull/2062.

wilkus27:
This will be resolved by https://github.com/openstax/cnx-recipes/pull/2107

MartaSztuka:
College physics vol 1
![Screenshot from 2020-04-16 09-55-23](https://user-images.githubusercontent.com/47241114/79429929-79b03800-7fc8-11ea-9d8b-9b81b58cb190.png)
vol 2
![Screenshot from 2020-04-16 10-06-30](https://user-images.githubusercontent.com/47241114/79430923-f7287800-7fc9-11ea-94e1-be9017e5e737.png)
vol 3
![Screenshot from 2020-04-16 10-19-44](https://user-images.githubusercontent.com/47241114/79432226-d19c6e00-7fcb-11ea-8a10-f76564ee1fd6.png)

Issue #1359, opened 2019-09-06
 Equations


MartaSztuka:
Some of equations look messy (pdf was mathified).
![Screenshot from 2019-09-06 14-43-56](https://user-images.githubusercontent.com/47241114/64428829-00a54580-d0b5-11e9-9d8c-2dda773b569f.png)
![Screenshot from 2019-09-06 13-48-05](https://user-images.githubusercontent.com/47241114/64428835-04d16300-d0b5-11e9-9b8f-f05b12e6b6e6.png)
![Screenshot from 2019-09-06 12-46-01](https://user-images.githubusercontent.com/47241114/64428881-1b77ba00-d0b5-11e9-8778-cfe8bdc94e57.png)

brittweinstein:
@MartaSztuka could you please take another look at this and confirm that it is still occurring?

MartaSztuka:
@brittweinstein Some of equations are still looking weird or cutting lines.
![Screenshot from 2019-09-24 09-51-46](https://user-images.githubusercontent.com/47241114/65492639-8dbc0d00-deb1-11e9-82c6-20038cbb1eae.png)

![Screenshot from 2019-09-24 09-54-31](https://user-images.githubusercontent.com/47241114/65492503-58afba80-deb1-11e9-9b98-c07c3bba1bbd.png)
![Screenshot from 2019-09-24 09-56-46](https://user-images.githubusercontent.com/47241114/65492683-a7f5eb00-deb1-11e9-9912-0b6b62a49399.png)

brittweinstein:
@MartaSztuka could you check if the 'mathify' process has been run and see if this is still happening?

MartaSztuka:
@brittweinstein Yes, it was mathified, issue still occurs.

MartaSztuka:
@brittweinstein Yes, it was mathified, issue still occurs.

wilkus27:
@brittweinstein As I understand at this moment PDFs are generated from files `baked.xhtml` not from `mathified.xhtml` files, so we can't see repaired equations  in PDF.  @helenemccarron am I right?

![image (12)](https://user-images.githubusercontent.com/40228854/67387388-13a5a380-f597-11e9-82bc-66518b0c0cba.png)

MartaSztuka:
Book was mathified, following Alan's instructions.
[...]

PR #2107, opened 2020-03-26
Rework Answer Key

wilkus27:
Issues: https://github.com/openstax/cnx-recipes/issues/1359, https://github.com/openstax/cnx-recipes/issues/2068, https://github.com/openstax/cnx-recipes/issues/2156, https://github.com/openstax/cnx-recipes/issues/2064, https://github.com/openstax/cnx-recipes/issues/2146


styles/designs/carnival/parts/_answerkey-components.scss
```diff
@@ -1,34 +1,40 @@
...
+
+$AnswerKey__Solution--Wrapper: (
+  _name: "SolutionWrapper",
+  _subselector: ' [data-type="solution"]',
   _properties: (
+    display: table,
+  )
+);
+
+// TODO after merge code from g.g branch add SolutionContainerSpecific component
```
brittany-johnson, line 15: Since g.g was merged into master, did you want to include this work in this PR?

wilkus27:
@brittany-johnson @ajluk I updated branch with carnival100. Some additional changes were also needed. But it's ready for review now.

recipes/books/_release.scss
```diff
@@ -328,5 +328,16 @@ $RELEASE_FLAGS: (
     "u-physics",
   ),
 
+  answer-key-h3: enable-for(
```
brittany-johnson, line 331: @wilkus27 Can you explain the problem that this release flag is solving? I want to make sure that I understand. I know that this PR is addressing a few different issues. Thank you so much!

brittany-johnson: Another question, I see that the recipe CSS for chemistry, microbiology, and u-physics has updated but I think the other books enabled by this flag need to be baked as well?

brittany-johnson: Another question, I see that the recipe CSS for chemistry, microbiology, and u-physics has updated but I think the other books enabled by this flag need to be baked as well?

recipes/mixins/_modify.scss
```diff
@@ -934,6 +934,18 @@
         class: attr(class) " first-element";
       }
     }
+
```
brittany-johnson, line 937: @wilkus27 I know that the `modify_addFirstElement` mixin was created in a different PR but I think it should be renamed to be more descriptive on what is happening. Maybe by renaming it to something like `modify_addFirstElementClass` so that it's a little clearer at first glance what is does. What do you think?

styles/designs/carnival/parts/_answerkey-components.scss
```diff
@@ -1,89 +1,88 @@
...
+          $AnswerKey__Solution--Paragraph,
+          $AnswerKey__Image,
+          $AnswerKey__ListComponent--MarginBottomLeftOrdered,
+          $AnswerKey__ListComponent--MarginBottomLeftUnordered,
+      ),
+  );
+}
+
+$AnswerKey__Solution--Container: solutionContainer(' .os-solution-container', null);
+$AnswerKey__Solution--ContainerSpecific: solutionContainer(' .os-solution-container.has-first-element', "Specific");
```
brittany-johnson, line 61: By reading the name of this component, `$AnswerKey__Solution--ContainerSpecific` its not clear to me its purpose. I think this can be solved by either renaming it to be more descriptive or adding a good comment. Whatever you prefer

recipes/books/_release.scss
wilkus27: @brittany-johnson The problem is related with H3 headers from Answer Key. Those titles are inside `<h3>` tag with no atributes. But `H3Title` component created by Helene has selector `h3[data-type="title"]`. To not create separate component or add additional selector to existing one I decided to add missing attribute to those headers by recipe. I think they are now more consistent with other H3 headers from the book.

About books listed in `enable-for`, they are all baked. But we see changes only in files of books which have Answer Key (so Chemistry, Microbio, U-physics). This flag doesn't have impact on outputs of other books from carnival theme. But there are some books from other themes with Answer Key. That's why I created release flag. I'm not sure what impact can have this change on them.

styles/designs/carnival/parts/_answerkey-components.scss
wilkus27: @brittany-johnson as we discussed on Clean Up call we are going to create one function (for exercises and solutions) so this will be renamed during this work.

recipes/mixins/_modify.scss
wilkus27: Yes, I agree. It's confusing now. Thank you for the suggestion :slightly_smiling_face:

wilkus27:
Thank you @brittany-johnson for all your comments! I changed name of the mixin as you suggested and I replied for your questions. It's ready for review now.

recipes/books/_release.scss
brittany-johnson: Ah okay! I understand now. I think that's a great idea to make the H3 headers in the answer key consistent with the others in the book. Thank you!

styles/designs/carnival/parts/_answerkey-components.scss
brittany-johnson: No, I think it would be better to fix it during the cleanup work. Thanks!

PR #2130, opened 2020-03-30
Eoc cleanup continued

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1767


wilkus27:
It's to hard to update this PR with the newest changes. All needed commits from here are copied to https://github.com/openstax/cnx-recipes/pull/2136.

PR #2131, opened 2020-03-30
Change hsphysics collection number

wilkus27:
After content was refreshed on katalyst01 HS Physics has different collection number now.


styles/designs/carnival/parts/_footnotes-components.scss
```diff
@@ -73,11 +73,14 @@ $FootnoteDefaultCallComponent: (
   _name: "DefaultFootnoteCall",
   _subselector: '',
   _properties: (
-    color: enum('ValueSet:::REQUIRED'),
-    line-height: enum('ValueSet:::REQUIRED'),
+    //styling for footnote call taken from reset.scss
```
helenemccarron: We removed the work on the footnotes for now (we will merge them later), is this necessary?

wilkus27:
@helenemccarron you can see BrittneyJ's commits about footnotes because this branch was created from g.g before her changes was reverted. To avoid mistakes I used `cherry-pick` to copy only my commit directly to carnival100. So I'm going to close this PR.

Issue #2141, opened 2020-04-02
Clean up: apply new shapes to microbiology


ajluk:
- apply EOCAssesmentsShape shape for eoc modules with exercises in microbiology
- apply new EOCDefaultCompoundComposite shape for eoc composite chapter in microbiology
- remove unused solution title component in the eoc exercises in hsphyscis

brittany-johnson:
Hi @ajluk ! Looks like this issue can be moved to ready for testing if you're ready.

MartaSztuka:
Microbiology
Issue about blue ribbon has been opened; #2181
![Screenshot from 2020-04-09 09-26-55.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5c2c7d89-38d4-4845-85cd-87966013917b)

PR #2142, opened 2020-04-02
Clean up microbiology

ajluk:
Issue: #2141


ajluk:
@wilkus27 I have enabled flag and added the config settings for images in exercises and added ContentExercises include use. Thanks for catching this up!

ajluk:
Changes in TDS: 
book specific settings has been added for microbiology in these components
GenericEOC

Issue #2125, opened 2020-03-27
Figure caption - spacing discrepancy between number and caption


MartaSztuka:
@helenemccarron @brittweinstein @lwtchu  Space between figure number and caption is too wide (happens in several other locations)
![Screenshot from 2020-03-27 13-25-17](https://user-images.githubusercontent.com/47241114/77755931-8c20fc80-702e-11ea-9ec2-56d60e3cd1dd.png)
Should be like this:
![Screenshot from 2020-03-27 13-26-34](https://user-images.githubusercontent.com/47241114/77755987-a3f88080-702e-11ea-9a43-ee3fd5452d8b.png)

lwtchu:
Might be because it is nested in a note?

brittany-johnson:
<img width="1441" alt="Screen Shot 2020-04-05 at 11 48 02 PM" src="https://user-images.githubusercontent.com/24909007/78523934-f86ecf00-7797-11ea-8ae0-0a92bab19347.png">
![Uploading Screen Shot 2020-04-05 at 11.47.51 PM.png…]()

MartaSztuka:
now they are the same
![Screenshot from 2020-04-10 10-14-45](https://user-images.githubusercontent.com/47241114/78975330-2a738e80-7b14-11ea-9456-92a9dd7b50dc.png)
![Screenshot from 2020-04-10 10-14-18](https://user-images.githubusercontent.com/47241114/78975335-2ba4bb80-7b14-11ea-9126-6dae1cf00550.png)

PR #2162, opened 2020-04-06
issue-2125

brittany-johnson:
I created a new component to remove the margin from the `os-divider` elements in figure captions. 
 https://github.com/openstax/cnx-recipes/issues/2125


wilkus27:
@brittany-johnson it looks fine. I'm just wondering why it has impact only on hs-physics. Output file of other books also should be changed :thinking:

brittany-johnson:
@wilkus27 Thanks for catching that, I just forgot to recompile everything 😱

styles/designs/carnival/parts/_figureimage-components.scss
```diff
@@ -56,6 +57,14 @@ $FigureTitle: (
     )
 );
 
+$FigureCaptionDivider: (
+    _name: "FigureCaptionDivider",
+    _subselector: "> .os-divider",
+    _properties: (
+        margin: 0px,
```
helenemccarron, line 64: `0` is better, we don't use pixels as a unit.

Issue #2129, opened 2020-03-27
Sub-Headers in Performance Task


MartaSztuka:
@helenemccarron @brittweinstein @lwtchu Headers `Materials` and `Procedure` looks weird having same size as `Performance Task` also should be left aligned.
![Screenshot from 2020-03-27 14-31-01](https://user-images.githubusercontent.com/47241114/77761469-64cf2d00-7038-11ea-9ef8-fbeb596bd6e2.png)

lwtchu:
We should look into this next week.

brittweinstein:
"materials" and "procedures" should be smaller than the other headers in this content.

openstaxalina:
@lwtchu could you provide specs for those sub-headers ("Materials", "Procedure")? I think these are all the instances of sub-headers in Performance Task throughout the book:

![Screen Shot 2020-03-30 at 9.57.11 AM.png](https://images.zenhubusercontent.com/5b57372fb01a900cfd923015/384df0d7-30f4-4fda-a841-632f4bcf3b08)

![Screen Shot 2020-03-30 at 9.57.28 AM.png](https://images.zenhubusercontent.com/5b57372fb01a900cfd923015/33cd03f3-73b0-4d75-8d09-c82cb0dcc7ac)!

[Screen Shot 2020-03-30 at 9.57.47 AM.png](https://images.zenhubusercontent.com/5b57372fb01a900cfd923015/72b75fce-7af0-41d0-9866-e5b8763173a0)

lwtchu:
These are marked up as list `<title>`s can we have them be:

font-size: font-scale(0)
font-type: alegreya
font weight: bold

MartaSztuka:
tested ![Screenshot from 2020-04-10 12-16-19](https://user-images.githubusercontent.com/47241114/78983765-213eed80-7b25-11ea-9d4d-ed333667582d.png)

PR #2172, opened 2020-04-08
Style list title

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2129

Issue comes from HS Physics, but I created general list title component and apply to each book. Let me know what do you think. Maybe better solution is to keep all settings OPTIONAL for this component and define them in book.scss file for HS Physics.


brittweinstein:
@wilkus27 i think we can merge!

Issue #2150, opened 2020-04-03
Ribbon in EoB


MartaSztuka:
End of Book needs blue ribbon on margin.
![Screenshot from 2020-04-03 12-58-02](https://user-images.githubusercontent.com/47241114/78353680-c5b3b380-75aa-11ea-8dee-ce7a00b5b896.png)

brittweinstein:
This should just require the enabling of the ribbon.

MartaSztuka:
tested ![Screenshot from 2020-04-10 13-11-22](https://user-images.githubusercontent.com/47241114/78987088-42a3d780-7b2d-11ea-8e0e-edb2d33e8935.png)

PR #2173, opened 2020-04-08
Add bands to eob in anatomy

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2150


brittweinstein:
@wilkus27 i think we can merge!

Issue #2157, opened 2020-04-03
Example headers


MartaSztuka:
Headers in Example are not consistent, e.g. Answer in different font.
![Screenshot from 2020-04-03 16-30-39](https://user-images.githubusercontent.com/47241114/78371780-b5aacc80-75c8-11ea-81a9-61209e680c85.png)

brittweinstein:
@lwtchu please provide expected specs

lwtchu:
It should match the other headers...
IBM Plex Sans
Font size(0) - 9pt
Bold

MartaSztuka:
![Screenshot from 2020-04-14 13-22-48](https://user-images.githubusercontent.com/47241114/79219725-13ee6f80-7e53-11ea-89a9-9d6c1ea5b59d.png)

Issue #2151, opened 2020-04-03
Headers in notes


MartaSztuka:
Headers in notes have inconsistent font and size.
e.g.
1. Grasp check
![Screenshot from 2020-04-03 13-07-16](https://user-images.githubusercontent.com/47241114/78354723-b897c400-75ac-11ea-8c03-d36d7c86d269.png)
![Screenshot from 2020-04-03 13-06-23](https://user-images.githubusercontent.com/47241114/78354724-b9305a80-75ac-11ea-8852-45fa0997604e.png)
2. Tips for Succes
![Screenshot from 2020-04-03 13-26-51](https://user-images.githubusercontent.com/47241114/78355890-d82fec00-75ae-11ea-8507-8137fc15174a.png)
![Screenshot from 2020-04-03 13-26-38](https://user-images.githubusercontent.com/47241114/78355898-d9f9af80-75ae-11ea-878b-a6c2ae324b92.png)
3. Worked Example headers
![Screenshot from 2020-04-03 13-32-07](https://user-images.githubusercontent.com/47241114/78356335-a4a19180-75af-11ea-8edd-450dba34caa6.png)

brittweinstein:
@lwtchu please provide specs

lwtchu:
They should be all     font-size: font-scale(1), - which in the screenshots most are. However they look smaller in the icon notes.

ajluk:
@MartaSztuka The issue about uppercase headers letters has been opened on content side: https://github.com/openstax/content-managers/issues/1660 ane change has been made on production only.
![image](https://user-images.githubusercontent.com/55203598/80396580-1a4e0400-88b5-11ea-8304-34c59a2fe584.png)

MartaSztuka:
![Screenshot from 2020-05-05 11-03-43](https://user-images.githubusercontent.com/47241114/81050353-20993d00-8ec0-11ea-8c0e-c9546591a256.png)
![Screenshot from 2020-05-05 11-02-51](https://user-images.githubusercontent.com/47241114/81050357-22630080-8ec0-11ea-9ace-56c255f13861.png)
![Screenshot from 2020-05-05 11-04-43](https://user-images.githubusercontent.com/47241114/81050428-40c8fc00-8ec0-11ea-9500-b89f889c63ec.png)
![Screenshot from 2020-05-05 11-08-43](https://user-images.githubusercontent.com/47241114/81050745-d2d10480-8ec0-11ea-8a2c-362c26eaf0f1.png)

PR #2175, opened 2020-04-08
Set font-size, family example answer title chemistry

ajluk:
Issue: #2157


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -266,7 +266,7 @@ $IconNoteEquationComponent: (
 
 $IconNoteAnswerTitleComponent: (
   _name: "AnswerTitle",
-  _subselector: " > [data-type='note'] > .os-title > .os-title-label",
+  _subselector: " [data-type='note'] > .os-title > .os-title-label",
```
wilkus27, line 269: @ajluk I'm afraid that this can have impact on other elements. Please check examples from https://github.com/openstax/cnx-recipes/issues/2151. E.g `Grasp Check` is a note inside a note in HS Physics.

ajluk: @wilkus27 I have changed the selector of this one component instead of creating a new one, because I have checked the other notes and it only is present in Examples in Chemistry, in Logo note is added to the shape but unused (not present in the markup). The other icon note shapes doesn't have the answer title with such murkup (not even with the previous selector).

brittweinstein:
@ajluk please go ahead and merge this

Issue #1483, opened 2019-10-21
Key equations have random leading differences.


MartaSztuka:
and it's not caused by equations's margine issue.
![Screenshot from 2019-10-21 13-29-45](https://user-images.githubusercontent.com/47241114/67201705-1e313300-f407-11e9-8863-aac28a6439b3.png)
![Screenshot from 2019-10-21 12-23-10](https://user-images.githubusercontent.com/47241114/67201761-44ef6980-f407-11e9-8d27-a9f3ec3644d5.png)

brittweinstein:
We will re-test this after the katalyst01 refresh.

brittweinstein:
Alina and Larissa will talk about if we should fix this and if so, a CM can migrate a couple of modules.  

We should not do a content refresh b/c we will lose HS Phys.

openstaxalina:
We will migrate content for chs 1 and 3 to katalyst01 by feb 28 (in time for f.f. sprint)

openstaxalina:
See: https://github.com/openstax/content-managers/issues/1456

wilkus27:
I think this is the same problem as in issue: https://github.com/openstax/cnx-recipes/issues/2013.

brittweinstein:
@MartaSztuka please verify if this is still an issue or if it was resolved by https://github.com/openstax/cnx-recipes/issues/2013

wilkus27:
@brittweinstein This issue was just blocked by #2013 . Related PR didn't resolve this problem, it fixed only exercises. Now we need to apply solution created by @ajluk to Key Equations.

wilkus27:
During testing please check also HS Physics. In this book there are section titles in Key Equations. I changed margins for some elements. Please verify if provided changes looks good.

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-04-23 12-40-41](https://user-images.githubusercontent.com/47241114/80090356-aa134b80-855f-11ea-963b-e8e93cda4ddd.png)
Chemistry AF @wilkus27 Key Equations have bullets sticking out, there shouldn't be any.
![Screenshot from 2020-04-23 13-24-48](https://user-images.githubusercontent.com/47241114/80094109-fbbed480-8565-11ea-92a6-1aedab00d9bc.png)
HS Physics looks ok
[...]

PR #2165, opened 2020-04-07
sections margins in exercises

ajluk:
Issue: #2013

PR #2225, opened 2020-04-15
Set margins in Key Equations

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1483


styles/books/hs-physics/book.scss
```diff
@@ -154,6 +154,8 @@
     'Container:::column-count': 2,
     'Container:::column-gap': 2.4rem,
     'Container:::column-width': auto,
+    'Section--MarginBottom:::margin-bottom': column-v-spacing(1),
```
ajluk, line 157: @wilkus27 I have just a question why in HSPhysics this margin need to be set differently than in the other books?

wilkus27: @ajluk In HS Physics in Key Equations there are section titles:

image

So I set bottom margin base on your work for Exercises modules: https://github.com/openstax/cnx-recipes/pull/2165/files

ajluk: @wilkus27 Thanks for the expalanation.

PR #2231, opened 2020-04-15
Note headers hsphysics

ajluk:
Issue: #2151
After discussion with Larissa I have added margins to headers in the Examples where it was needed as well for consistency with cases, where the markup is different and the headers are nested in paragraphs (have inherited margins), which wasn't possible to remove without affecting other paragraphs placed in the icon notes.


styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -5,7 +5,7 @@ $NoteSubtitleComponent: (
   _properties: (
     font-size: font-scale(1),
     line-height: line-height(1),
-    font-family: "IBM Plex Sans", //too many places where it's  optional to fix now, TODO: fix later
+    font-family: "IBM Plex Sans, sans-serif", //too many places where it's  optional to fix now, TODO: fix later
```
ajluk, line 8: I have added the `sans-serif` to font-family property in this component as a temporary solution (was missing it). This property need to be set in the book.scss files.

styles/books/hs-physics/book.scss
```diff
@@ -115,8 +115,12 @@
     'TitleText:::margin-left': h-spacing(1),
     'Subtitle:::font-family': (_ref: 'typography:::sansFont'),
     'Subtitle:::font-weight': bold,
+    'Subtitle:::margin-bottom': .2rem,
     'DivSubtitle:::font-family': (_ref: 'typography:::sansFont'),
     'DivSubtitle:::font-weight': bold,
+    'DivSubtitle:::margin-bottom': .2rem,
+    'DivSubtitle:::text-transform': lowercase,
+    'DivSubtitle--FirstLetter:::text-transform': uppercase,
```
ajluk, line 123: I had to added this additional settings in this book only to overwrite content uppercase letters in header.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -198,16 +199,16 @@ $IconNoteMainTitleTextComponent: (
 );
 
 $IconNoteFirstParagraphTitleComponent: (
-  _name: "FirstTitle",
-  _subselector: " + p > [data-type='title']",
+  _name: "FirstParagraphTitle",
+  _subselector: " p:first-of-type > [data-type='title']",
```
wilkus27, line 203: I'm wondering about selector :thinking: If I understand correctly, previously there was such situation: 

component `$IconNoteMainTitleComponent` had child component `$IconNoteFirstParagraphTitleComponent`, so the selector was ` ... > .os-title +p > [data-type="title"]`. That is, title inside para which is sibling of `.os-title`.

Now `$IconNoteFirstParagraphTitleComponent` is child of `$IconNoteContainerComponent`, so the selector is `.class-of-note p:first-of-type > [data-type='title']`. That is title inside first para from note.

I'm just not sure if this change cover the original title. :thinking:

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -252,6 +254,15 @@ $IconNoteDivTitleComponent: (
...
     padding-right: enum('ValueSet:::OPTIONAL'),
+    margin-bottom: enum('ValueSet:::OPTIONAL'),
+  )
+);
+
+$IconNoteDivTitleComponent--FirstLetter: (
+  _name: "DivSubtitle--FirstLetter",
+  _subselector: " div:not([class]) > [data-type='title']:first-letter, div[data-type='title']:first-letter",
+  _properties: (
+    text-transform: enum('ValueSet:::OPTIONAL'),
```
wilkus27, line 265: Look at my comment from above, as I said there, maybe we can make a change in content rather then fix it in code.

styles/designs/carnival/parts/notes/_notes-components.scss
Stackblocks: Is this something that should be fixed now instead of band-aided so we don't forget about it later?

ajluk: I have thought it should be fixed during the cleaning within new PR for test purposes.

styles/designs/carnival/parts/notes/_notes-components.scss
ajluk: @wilkus27 Yes it does cover the original title. I have created this component for University Physics Examples and now the markup is different  there. That's the reason this title isn't styled now.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -5,8 +5,9 @@ $NoteSubtitleComponent: (
   _properties: (
     font-size: font-scale(1),
     line-height: line-height(1),
-    font-family: "IBM Plex Sans", //too many places where it's  optional to fix now, TODO: fix later
+    font-family: "IBM Plex Sans, sans-serif", //too many places where it's  optional to fix now, TODO: fix later
     font-weight: enum('ValueSet:::OPTIONAL'),
+    margin-bottom: enum('ValueSet:::OPTIONAL'),
```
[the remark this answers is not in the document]
ajluk: @wilkus27 Thanks for bringing this up! I have added the missing comment.

styles/books/hs-physics/book.scss
brittany-johnson: @ajluk @wilkus27 So in the content, the entire title is uppercase, but it should be displayed with only the first letter of each word being uppercase? Example: 

Content -> LEARNING OBJECTIVES 
Desired output -> Learning Objectives 

If this is the case, I think this should be a content change

styles/designs/carnival/parts/notes/_notes-components.scss
ajluk: @wilkus27 I have removed this component and opened an issue about it on content side: https://github.com/openstax/content-managers/issues/1660.

brittweinstein:
@ajluk I think we can go ahead and merge!

Issue #2235, opened 2020-04-16
Fix exercises and Answer Key in Fizyka


wilkus27:
To display exercises from EOC and Answer Key correctly, flags `exercise-first-element` and `answer-key-h3` should be enabled. Description of those flags is on https://docs.google.com/document/d/1ziR2PcSvioOLAHSRNWZga3m-8DlUM_qaskIjpWtoS60/edit#heading=h.hca2jjbhejsz

MartaSztuka:
Exercises first element is not a case in Fizyka. ![Screenshot from 2020-04-23 15-22-51.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/e70bd37b-fc52-4e8f-a42a-93624880e870)
![Screenshot from 2020-04-23 15-20-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5c8bc376-efbe-44bf-81b4-c1e4b7336b02)

PR #2238, opened 2020-04-16
Fix exercises and answer key in Fizyka

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2235


wilkus27:
#### Recipe changes:

- remove doubled `pl-u-physics` from flags `exercise-first-element` and `answer-key-h3` in release.scss file
- add key `type` to config `$Config_BookCompositePages` for accounting, american-government, ap-physics, business-ethics, calculus, chemistry, dev-math, economics, history, microbiology, pl-u-physics, precalculus, u-physics (it changes only pl-u-physics output)
- modify mixins `modify_addFirstElementClass` and `modify_tagExerciseWithFirstElementClass` - change condition to use key `type` from `$Config_BookCompositePages` config. There is needed one condition for all books which have Answer Key. Previously it was the `name`, but in polish u-physics it sounds differently. That's why the new one is added.
- add divider to Check Your Understanding note in Polish U-Physics
- add selectors of figures/images which are first elements of exercises/solutions to Polish U-Physics config.scss file
- remove string `Rozdział` from `$Config_EOBSolutionsChapterSectionTitle` config in `pl/common-stuff.scss` file

Issue #1981, opened 2020-02-28
Symmetry Feedback: Punctuation after math (?) can wrap to next line


openstaxalina:
# Scenario

I'm guessing this is because there is math markup immediately followed by punctuation.  Can anything be done in these instances?

From 2.5:
![Screen Shot 2020-02-28 at 5 45 19 PM](https://user-images.githubusercontent.com/7863733/75595482-3d456e80-5a52-11ea-9623-a0a6abb9c878.png)

From 2.7:
![Screen Shot 2020-02-28 at 5 48 14 PM](https://user-images.githubusercontent.com/7863733/75595580-97deca80-5a52-11ea-9696-07ec4781fa84.png)

# Acceptance Criteria (required)

# Test Plan

Stackblocks:
# In the markup: 

## 2.5
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/3da1eaf7-820c-4d8e-8c6e-63f2d8c55956)
## 2.7
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/6adf21d3-0cd6-430a-aa2d-b6527b72ad86)

brittweinstein:
@openstaxalina will investigate and track down an instance of this happening at the end of a line in the PDF.

openstaxalina:
On today's call, we discussed how this issue had been handled in oer.exports. Perhaps we can implement a similar solution in baked PDF.  

I found an example in oer.exports of math at the end of the line, followed by punctuation, where the punctuation is **not** breaking from the math:

![oerexportsCollegePhysicsmath](https://user-images.githubusercontent.com/7863733/77443404-0a746900-6db9-11ea-8463-2fba116935c0.png)

Because of the differences in font and styles, in the baked PDF, that content does not occur at the end of a line, so it's not a perfect comparison. But I believe the screenshot above shows that there should be a way to handle inline math, followed by punctuation at the end of a line.

brittany-johnson:
![Screen Shot 2020-04-16 at 3 55 10 PM](https://user-images.githubusercontent.com/24909007/79505980-d9bfd200-7ffa-11ea-8c22-cd78cc6c4e8e.png)
![Screen Shot 2020-04-16 at 3 54 25 PM](https://user-images.githubusercontent.com/24909007/79505987-db899580-7ffa-11ea-9ceb-c65765309eec.png)

brittany-johnson:
[...]

Issue #1890, opened 2020-02-10
WAITING TO RESOLVE ALL HIGH PRIO ISSUES: Enable All Release Flags


openstaxalina:
# Scenario

Discussed in today's team meeting:
We will enable *all* remaining release flags (see: https://github.com/openstax/cnx-recipes/blob/master/RELEASE_FLAGS.md).

Content Managers will assist with a full-library regression test.

Alina/Larissa will review issues and prioritize any showstoppers.  (A fallback is to re-enable any flags we can't fix within the sprint.)

Other regressions will be put in future sprints.

# Acceptance Criteria (required)
- [x] all release flags have been enabled (no red "X"s at the above link)
- [x] **developers have run diffs and listed below any specific things that seem to have changed in the books**
- [x] QA has looked in these specific places
- [x] CMs have done regression testing (front/back/three chapters) and opened any issues
- [x] Alina has identified any showstopper issues
- [x] issues have been resolved or those release flags have been re-enabled and will be tracked via cards in subsequent sprints

# Test Plan

brittany-johnson:
**Document outlining all release flags enabled:** 
https://docs.google.com/document/d/1ziR2PcSvioOLAHSRNWZga3m-8DlUM_qaskIjpWtoS60/edit?usp=sharing

brittany-johnson:
Hello @MartaSztuka! I just updated the release flags document with the details to test the `give-composites-ids` release flag. 
Can you re-test this flag with the acceptance criteria in the document? 

Thank you! 

https://docs.google.com/document/d/1ziR2PcSvioOLAHSRNWZga3m-8DlUM_qaskIjpWtoS60/edit#heading=h.dk8rtuc6fky3

MartaSztuka:
`give-composites-ids` flag was retested, work as expected for all books.

brittany-johnson:
My #1 area of concern is the TOC, specifically making sure that everything is numbered and linking properly. 

Other general areas that I think are important to check are the chapter introduction, EOC, and EOB, mainly checking that things are numbering and linking correctly. 

TOC
- unit, chapter, and appendix titles
- EOC and EOB numbering and linking 

Chapter Introduction
- Splash
- Chapter Outline

EOC 
- module titles linking

PR #2245, opened 2020-04-16
issue-1981

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/1981

This change affects all recipes but I didn't put this change behind a release flag because we are enabling all release flags in ignoble.icarus


brittany-johnson:
@wilkus27 Thank you for your review! 
According to this issue: https://github.com/openstax/cnx-recipes/issues/1890 we're going to enable all of the release flags and do library-wide testing. 

I just merged `i.i` into the branch so I think pl-u-physics is up to date now. Thanks for catching that!

brittweinstein:
@brittany-johnson let's go ahead and merge this since we've got 2 approves already.

PR #2246, opened 2020-04-17
enabling flags

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/1890


brittweinstein:
@brittany-johnson let's go ahead and merge this since we've got two approvers

PR #2251, opened 2020-04-21
Issue 2168 - cleanup appendix

Stackblocks:
- Moved H1 over H2 in the appendix-shapes
- Rename old appendix-specific components to be generic
- Swapped in H1Component
- Replaced the chemistryTitleColor used with a generic appendix version
- First step towards breaking border settings out of titles
and into their own components.


styles/output/chemistry-pdf.css
```diff
@@ -882,6 +919,7 @@ a {
   font-weight: bold;
   text-transform: uppercase; }
 
+>>>>>>> ignoble.icarus
```
ajluk, line 922: @Stackblocks There are some unresolved conflicts.

styles/output/chemistry-pdf.css.map
```diff
@@ -116,5 +116,9 @@
...
+>>>>>>> ignoble.icarus
```
ajluk, line 123: See my comment above.

styles/templates/template1/_colors.scss
```diff
@@ -99,5 +105,6 @@
     visualColor: (_ref: "scheme:::B2"),
     watchPhysicsColor: (_ref: "scheme:::B2"),
     workInPhysicsColor: (_ref: "scheme:::BR1"),
+>>>>>>> ignoble.icarus
```
ajluk, line 108: Same here.

styles/output/chemistry-pdf.css
Stackblocks: Thanks for catching these! I guess this is what happens when I try to resolve the conflicts by hand for the first time, lol. I'll resolve these and re-commit. ✅

styles/designs/carnival/parts/_titles-components.scss
```diff
@@ -9,57 +9,57 @@ $H1Component:(
...
+        border-bottom-color: enum('ValueSet:::OPTIONAL'),
+        border-bottom-width: enum('ValueSet:::OPTIONAL'),
+        border-bottom-style: enum('ValueSet:::OPTIONAL'),
         margin-bottom: v-spacing(2),
     )
 );
 
 
 //h1 numbered (appendix)
 //todo rename this to not be appendix specific
```
brittany-johnson, line 26: I think we're safe to remove the todo comment

PR #2255, opened 2020-04-22
bake a chapter

brittany-johnson:
details of the script can be found in comments


wilkus27:
@brittany-johnson I tested it locally:

- I fetched `chemistry-atoms-first-2e`
- Run assemble script
- Run new script and I get this:

image

- I checked `git status` and `temp` css file is created:

image

Stackblocks:
If I try to run it with out `./script/run` I get a "No such interpreter" error, and when I use `./script/run` I get the same error as Natalia.

Additionally, an example of the command should be included in the README

philschatz:
There are linter errors which need to be fixed in order to merge this code. See the [Travis Log](https://travis-ci.org/github/openstax/cnx-recipes/builds/687586348#L470)


Also, merging master into this branch may fix the cnx-easybake not found error that folks were seeing.

Issue #2244, opened 2020-04-16
Give the Word "Chapter" and the Chapter Number Different Indicators


openstaxalina:
# Scenario

Per discussion in Unified-Content:

change from
<.os-number>Chapter 4</> <.os-divider>-</> Kinematics in 7 Dimensions
into the following instead?
<.os-somekindofthrowaway-namethatCEknowsabout>Chapter</> <.os-number>4</> <.os-divider>-</> Kinematics in 7 Dimensions

# Acceptance Criteria (required)
- [ ] REX is able to discard the word "chapter" in the URLs, even when books are baked with recipes that have the "Chapter" in the page title

# Test Plan

philschatz:
Corresponding CE card so we can coordinate: https://github.com/openstax/cnx/issues/972

brittany-johnson:
Expected markup : 

![Screen Shot 2020-05-13 at 10.03.25 AM.png](https://images.zenhubusercontent.com/5b0d6efdd59b032fbd122962/96972768-0ba4-4ae6-9a31-f9c7638b3c42)

omehes:
Discussed with BJ and verified on Prealgebra 
![prealg2.png](https://images.zenhubusercontent.com/5bc8d1f2fe40ab78d94e65fb/2cf1c03e-56a0-4cc2-a1e9-a68a139a04f0)

and Uni Phys 1 on easyvm5

PR #2262, opened 2020-04-27
issue-2244

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2244

This must go out the CE tech team's slug generation work so I will be merging this PR directly into master


styles/templates/template1/parts/_toc-settings.scss
```diff
@@ -145,10 +145,6 @@
       font-size: font-scale(1),
       line-height: line-height(1)
     ),
-    TocChapterNumber: (
```
brittany-johnson, line 148: this is a duplicate of the block on line 32 so I removed it

recipes/books/_common-stuff.scss
```diff
@@ -19,7 +19,7 @@ $Config_PartType_Appendix_TitleContent  : (os-number: counter(appendix, upper-al
 
 
 @if TEMPLATE() == "TEMPLATE1" {
-$Config_PartType_Chapter_TitleContent   : (os-number: "Chapter " counter(chapter), os-divider: ' ' );
+$Config_PartType_Chapter_TitleContent   : (os-chapter-string: "Chapter", os-divider: ' ', os-number: counter(chapter));
 $Config_PartType_Unit_TitleContent   : (os-number:"Unit " counter(unit), os-divider: ' ' );
 $Config_PartType_Appendix_TitleContent  : (os-number: "Appendix " counter(appendix, upper-alpha), os-divider: ' ');
```
wilkus27, line 24: I have a question :slightly_smiling_face: Why it is done just for the `Chapter`? Shouldn't be similar change made for `Unit` and `Appendix` as well? Also I'm thinking about key naming. In other configs we are using `os-text` to define string. It's just my observation. Maybe in this case used name is a better convention.

styles/templates/template1/parts/_toc-settings.scss
wilkus27: I'm not sure if it should be removed :thinking: In line 32 `TocChapterNumber` is defined for `TOC` element. In line 111 starts `UnitToc` element and I think this component should be defined here as well.

styles/output/chemistry-pdf.css
```diff
@@ -1131,13 +1131,14 @@ nav#toc {
...
-  margin-left: 0; }
+  margin-left: 0;
+  display: inline; }
 
 nav#toc > ol > .os-toc-chapter {
   display: block;
   margin-bottom: 1.4rem; }
 
-nav#toc > ol > .os-toc-chapter > a > .os-number {
+nav#toc > ol > .os-toc-chapter > a > .os-chapter-number {
```
wilkus27, line 1141: I'm wondering if all files are generated? Only output for chemistry is changed.

styles/templates/template1/parts/_toc-settings.scss
ajluk: @brittany-johnson I agree with Natalia - two shapes are using this component.

recipes/mixins/_modify.scss
```diff
@@ -1027,3 +1027,25 @@
...
+        move-to: chapterNumber;
+    }
+    > .os-divider {
+        move-to: chapterNumber;
+    }
+  }
+  nav#toc > ol > .os-toc-chapter > a {
+    &:deferred::before {
+      container: span;
+      class: "os-chapter-number";
```
philschatz, line 1047: Is an additional element nesting necessary? 

So far, users of the baked styles assume the title only contains text or an element with only text (no elements with elements...). 

If it is then a very different approach will need to happen in https://github.com/openstax/cnx-common/pull/9 and I think other things may stop working properly.

/cc @rnathuji

brittany-johnson: Hello! 
Originally, it was not my intention to nest the elements, but due to design requirements, I found nesting `.os-chapter-string` `.os-divider` and `.os-number`  in one span allows me to apply the style `display:block`to the span so that they are the only elements on the first line and the.`os-text` and the linking page number (which only appears in the PDF) are on the second line. 

If there is no nesting of elements, I have four spans that are siblings: `.os-chapter-string` `.os-divider` `.os-number` and .`os-text`

`.os-chapter-string` `.os-divider` and `.os-number` need to be on the first line per design requirements and 
.`os-text` and the linking page number (only seen in the PDF) are on the second line. 

Nesting .`os-text` and the linking page number in one span with the recipe so that they can be styled to appear on the second line isn't possible because I don't have access to the page number in the recipe. This is why I thought to group `.os-chapter-string` `.os-divider` and `.os-number`  in one span because I have access to them during baking. 

Without any display properties, this is how the chapter title appears: 
Screen Shot 2020-04-30 at 1 38 38 PM

When seeing this my next thought was to apply a `display: block` to `.os-text` (in this example, the text is "Essential Ideas" so that it would appear on the second line. Doing this does move "Essential Ideas" to the second line but the page number is pushed to the third line because "Essential Ideas" is now taking up the entire width of the line: 
Screen Shot 2020-04-30 at 1 46 04 PM

My next thought was to attempt floating the page number to the right to force it on the same line as "Essential Ideas" but the result does not align with the desired design: 
Screen Shot 2020-04-30 at 1 58 12 PM

I tried a variety of other display properties on "Essential Ideas" and the page number but nothing gave me the result I was looking for. 

This is why I thought to nest "Chapter" and "1" in one span so that I could force it to be a block element, making "Essential Ideas" and the page number "7" to move to the second line and stay "inline" elements

recipes/books/_common-stuff.scss
brittany-johnson: Ahhh yes good point @wilkus27, I'll change the key naming. 
As to why I didn't make a similar change for Appendix and Unit, I honestly didn't think about it since the request for this change was specifically for the TOC. But thank you for pointing it out! 

@philschatz Would it be of use if I also applying this change to the Appendix and Unit here?

styles/output/chemistry-pdf.css
brittany-johnson: Just re-compiled. Thanks for catching that!

styles/templates/template1/parts/_toc-settings.scss
brittany-johnson: Forgot there was a NoUnitToc and a UnitToc. 
Thanks ya'll!

recipes/books/_common-stuff.scss
brittany-johnson: I changed it to `os-part-text` since there is another element in the tag with `os-text` and I don't want to move that one

philschatz: yes, for consistency it would be good to do it for Appendix as well.

recipes/output/accounting.css
```diff
@@ -2507,6 +2507,30 @@
...
+  content: pending(chapterPartString) pending(divider); }
+
+:pass(14) nav#toc > ol .os-toc-appendix > a > .os-part-text {
+  move-to: chapterPartString; }
+
+:pass(14) nav#toc > ol .os-toc-appendix > a > .os-number:deferred::before {
+  container: span;
+  class: "os-part-number";
+  content: pending(chapterPartString) pending(divider); }
+
```
wilkus27, line 2533: I'm wondering that maybe, since those changes are applied to all books (not only from carnival theme), we should remove `if` condition from `common-stuff` file?

image

Now, e.g this book - accounting- has title structure define in lines 16-18.

recipes/books/_common-stuff.scss
```diff
@@ -19,9 +19,9 @@ $Config_PartType_Appendix_TitleContent  : (os-number: counter(appendix, upper-al
 
 
 @if TEMPLATE() == "TEMPLATE1" {
-$Config_PartType_Chapter_TitleContent   : (os-number: "Chapter " counter(chapter), os-divider: ' ' );
-$Config_PartType_Unit_TitleContent   : (os-number:"Unit " counter(unit), os-divider: ' ' );
-$Config_PartType_Appendix_TitleContent  : (os-number: "Appendix " counter(appendix, upper-alpha), os-divider: ' ');
+$Config_PartType_Chapter_TitleContent   : (os-part-text: "Chapter ", os-number: counter(chapter));
```
wilkus27, line 22: Key `os-divider` is missing.

recipes/mixins/_modify.scss
```diff
@@ -1027,3 +1027,17 @@
     }
   }
 }
+
+//takes the string "Chapter" and the chapter number elements in the TOC and wraps them in a span
+@mixin modify_groupTocTitleTextAndNumber($element) {
+  nav#toc > ol #{$element} > a {
+    > .os-part-text {
+        move-to: chapterPartString;
```
wilkus27, line 1035: Since we are using this mixin also to Unit and Appendix now, maybe `chapterPartString` could be changed to `PartString` or to some other, more general name?

recipes/mixins/_modify.scss
```diff
@@ -1027,3 +1027,17 @@
...
+      }
+    > .os-number:deferred::before {
+      container: span;
+      class: "os-part-number";
+      content: pending(chapterPartString) pending(divider);
```
wilkus27, line 1040: I can't find where `divider` is defined. I mean, there is no `move-to: divider` part anywhere. Does this `pending(divider)` needed?

recipes/output/accounting.css
brittany-johnson: oooo Good point, I'll remove the conditional 
Thank you!

recipes/mixins/_modify.scss
brittany-johnson: I was in such a rush on Friday to finish I didn't even notice this ☠️

brittany-johnson:
@wilkus27 Ready for re-review! 
Thank you for reviewing this multiple times😅I really appreciate it.
I was so ready to be done with this issue that I missed a lot ☠️

wilkus27:
You're welcome :relaxed: Happy I can help :wink:

Issue #2112, opened 2020-03-26
Table bottom line


MartaSztuka:
@helenemccarron @brittweinstein @lwtchu Bottom blue line should be table wide.
![Screenshot from 2020-03-26 13-52-41](https://user-images.githubusercontent.com/47241114/77648959-4b12e480-6f69-11ea-86cf-277f5bf42e45.png)

lwtchu:
We should fix this but probably not for this release.

ajluk:
@lwtchu I looked into it and I think it is very hard to resolve it without affecting other rowspan tdies in the other top titled tables (was trying to work around it somehow but I haven't found the right solution). Could it be possible to add a class to this particular td in this table (it is not the last one only of type, so it is impossible to tag it right now)?: 
![image](https://user-images.githubusercontent.com/55203598/80001934-446f8280-84bf-11ea-9d15-66f0300b5fa4.png)

lwtchu:
hmm i'd rather not add another class - can we make it all black then?

ajluk:
Confirmed with Larissa: Top Titled Tables in all books should have changed border bottom book specific color to black.

MartaSztuka:
Colour is fixed, but line width should be consistent.
![Screenshot from 2020-04-24 11-45-05](https://user-images.githubusercontent.com/47241114/80199898-240c0900-8622-11ea-8b67-d10a0b9b6c8e.png)

MartaSztuka:
Fixed
![Screenshot from 2020-05-05 10-21-31](https://user-images.githubusercontent.com/47241114/81047260-56d3be00-8eba-11ea-8a22-f3ce0c7e59a8.png)

PR #2263, opened 2020-04-27
Fake title table bottom-border

ajluk:
Issue: #2112


ajluk:
After discussion with BrittanyJ and Larissa I implemented the safest solution - add black color to the bottom-borders in Fake Titled Tables as a  shorthand bottom-border property for the the whole last table row which also resolves the width issue.

brittweinstein:
@ajluk I think we can go ahead and merge this

PR #2268, opened 2020-04-30
Remove empty section from eoc

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/1885

Here is the continuation of Helene's work from https://github.com/openstax/cnx-recipes/pull/1997.

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -279,7 +279,15 @@
   @include compose_prepEOCcompoundCompositesPages($_chapterPrepCompoundCompositePages);
 }
 :pass(4) {
-
+  $_chapterPrepCompoundCompositePages: ();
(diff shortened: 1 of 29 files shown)


wilkus27:
### Recipe changes

- split mixin `_processComposites` into two parts (`compose.scss`)
  - in `pass(3)` create links with section titles inside `section.#{className}` and move whole section to new container with class `os-#{$className}-container`
  - in `pass(4)` using new mixin `compose_moveSectionSeparted` move `section.#{className}` into new container with class `os-section-area` and then move it to container with class `os-#{$className}`. Final markup is the same as previously. For example:
```
.os-chapter-review-container
    .os-concept-container
        .os-concept
           .os-section-area
               section.concept
                   a
````
- move all changes started from `pass(4)` by one pass (`generator.scss`)

Issue #2224, opened 2020-04-14
CLEANUP: Tables


brittany-johnson:
[Cleanup doc](https://docs.google.com/document/d/10qrKjBt3sGoZR8KaW1xwUSApVZiRPtFxce2WkRP2LJ0/edit#)

- [x] Page-break in the $TableTitle component should be moved to the page break file

- [x] Review component values and update where necessary (required vs optional vs hardcoded) 

- [ ] ~~Separate out borders from tables / create a border component~~

- [ ] ~~Reorganize document to have generic table components up top, then comment-described specialty tables below~~

- [x] Captions are used in tables and figures/images. Create one general caption component to replace the current ones in the tables and figures component files

brittany-johnson:
Related issue: https://github.com/openstax/cnx-recipes/issues/1922

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-05-11 17-16-20](https://user-images.githubusercontent.com/47241114/81578527-34a2da00-93ab-11ea-9063-88635078e005.png)
![Screenshot from 2020-05-11 17-18-04](https://user-images.githubusercontent.com/47241114/81578652-65830f00-93ab-11ea-8a15-d254a29962de.png)
HS Physics
![Screenshot from 2020-05-11 17-23-32](https://user-images.githubusercontent.com/47241114/81580124-4dac8a80-93ad-11ea-9be6-3e206756138b.png)
![Screenshot from 2020-05-11 17-34-04](https://user-images.githubusercontent.com/47241114/81580359-a11ed880-93ad-11ea-910e-dd199b689b50.png)
AP Biology
![Screenshot from 2020-05-11 17-36-52](https://user-images.githubusercontent.com/47241114/81580713-15f21280-93ae-11ea-8cb4-d37c2cccc466.png)

Issue #1717, opened 2019-12-20
CLEAN UP: Rework tables shapes


wilkus27:
Most of tables have such structure:

```
div.os-table
 | div.some-class
 | table.unique-class
 | div.os-caption-container
```
Some of them has additional title and/or caption, but always there is container with class `os-table` which has child - table with its unique class.

Current problem is that all the table settings are pointing at `_selectors: (".os-table")`, so all tables shapes have impact on each other.

wilkus27:
**Conclusion after discussion on the slack:**

per Helene

> Moving the type of table to the container rather than the table itself then changing the setting selectors makes sense. I wonder if maybe copying the table class with a different name (something like os-fakeTitle-container) so we can select the specific table container or table itself would be optimal? The settings then will need to be changed to a unique selector. If we decide not to change the recipes, an alternative would be to create new components with unique classes instead of using only shapes and use the new namespaced components for the shapes.

MartaSztuka:
Chemistry![Screenshot from 2020-05-12 10-18-02.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5c5ef53b-f2c5-414e-8b80-ea5c25cb589d)
![Screenshot from 2020-05-12 10-07-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7a3d9dfb-fe8d-4d0c-8170-6372c13179b7)
HS Physics![Screenshot from 2020-05-12 10-21-05.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f9dea57f-ef51-4a6a-b1a8-13c955d9c7cb)![Screenshot from 2020-05-12 10-19-22.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/9f52c551-ef74-4a80-ba38-e053e6f9faeb)
![Screenshot from 2020-05-12 10-10-34.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/beda6189-bef7-4152-a85f-f2c60e4c9054)![Screenshot from 2020-05-12 10-12-25.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/cfd7d747-4d1b-472c-aec5-5c163e98d244)
[...]

Issue #2221, opened 2020-04-14
CLEANUP: Section Key Terms


brittany-johnson:
[Cleanup doc](https://docs.google.com/document/d/10qrKjBt3sGoZR8KaW1xwUSApVZiRPtFxce2WkRP2LJ0/edit#)

- [x] Investigate if this can be made to use the table components to simplify

- [ ] Replace the ones that match existing table components to reduce uniques

ajluk:
After conversation with Larissa (https://katalysteducation.slack.com/archives/CBW9DFKGF/p1589456742238300) I have removed all the section key terms files, because right now after tables cleaning it is using the Unnumbered Unstyled table shape and the look of the table isn't much different than when we have separate styles for it before the tables cleaning. The only differences were margin-bottom of the table which previously was set to 0 and padding-left of the table data which previously was set to 0 as well. Previously the styles of the tables after cleaning overwrites it as you can see in the output files (there is no change). This is how it looks now in the html file:
![image](https://user-images.githubusercontent.com/55203598/81954539-45508b80-9609-11ea-89e1-cd8c5e0d4f11.png)
and in pdf file:
![image](https://user-images.githubusercontent.com/55203598/81954586-51d4e400-9609-11ea-98ab-2f44b521b00f.png)
How it looked before:
![image](https://user-images.githubusercontent.com/55203598/81954647-64e7b400-9609-11ea-981b-0fd64f5a7aea.png)

ajluk:
@MartaSztuka >> Information for testing purposes: 
Now this par should look like this: 
![image](https://user-images.githubusercontent.com/55203598/82457579-de721d00-9ab5-11ea-9189-9e55bc2d8093.png)

MartaSztuka:
![Screenshot from 2020-05-20 16-58-51](https://user-images.githubusercontent.com/47241114/82463135-bcc86400-9abc-11ea-8b32-6e0d74ef01cb.png)

Issue #2040, opened 2020-03-10
CLEAN UP - EOC: remove named table components from EOC


helenemccarron:
Remove the named table components and add existing table components to shapes in EOC (this includes exercises etc.)

brittweinstein:
@helenemccarron please put points on this when you start work.

wilkus27:
Waiting for merge changes from #2188 and #1717 .

wilkus27:
Done within https://github.com/openstax/cnx-recipes/pull/2300

PR #2270, opened 2020-04-30
Clean up table shapes

ajluk:
Issue: #1717
Changes in this PR will also resolve this issue:  #2224

Recipe changes:

- created a mixin `modify_tagTableParentContainer()` which tags table parent container with inside table and create container's additional classe based on the existing table's class:
 (`os-top-titled-container`, `os-unstyled-container`, `column-header`)
- tables classes which are affected by this changes are placed in the `common-stuff.scss` file under variable $Config_Table_Classes() and these are: `top-titled`, `unstyled`, `column-header`.


ajluk:
I have left some code commented which I want to remove just for review purposes and enable flag that I used for all carnival theme books because this change will affect all of them and changed selectors in settings wouldn't work without it. @brittany-johnson I have added you to this review because this is rather a big change and I'm courious what you and Natalia think about these solutions.

recipes/mixins/_modify.scss
```diff
@@ -979,6 +979,68 @@
...
+  @if release-flag-enabled(tag-table-parent-container) {
+    .os-table {
+      string-set: hasTable " ";
+        table[class] {
+        string-set: hasTable " " attr(class);
+        }
+      &:deferred {
+        class: attr(class) string(hasTable);
+      }
+    }
```
wilkus27, line 992: I'm not sure about this solution. At this moment we have two elements with the same class, tag with `os-table` class and tag `table`. 

In my opinion, better solution is this one, which you commented. I mean, to create such structure:
```
<div clas="os-table os-top-titled-container">
   <table class="top-titled"></table>
</div>
```

But, I think new classes should be created by some loop. We shouldn't define them for each class separately.
Something like that:

```
.os-table {
  class: attr(class) "os- #{$className}-container";
}
```

wilkus27: After discusion with @ajluk : 

we can't use my idea from above. The problem is, that some of the tables have a few classes. E.g there is a table with `class="unnumbered unstyled section-key-terms"`. So in this case when we want to copy class from the tag, the new one is `os-unnumbered unstyled section-key-terms-container`. This is not what we want.

Proposition: create a config with all classes, something like that

`$Config_TableClasses: ("unstyled", "top-titled", "column-header")`

and create loop:

```
@each $class in $Config_TableClasses {
   table.#{$class} {
      &::outside {
          container: div;
          class: "os-#{$class}-container";
      }
   }
}
```
then the markup will be:

```
div.os-table
    | div.os-unstyled-container
          | table.unstyled
```

We are not sure how much impact on styles it will have. Maybe container `div.os-table` will not be needed anymore, so it could be supress. This is to discuss.

ajluk: @wilkus27 Thanks for the suggestions!I have rework the mixin as we disussed. I also have removed Section Key Terms as. Now the Unstyled Table styling is covering the table inside it, but the margin-bottom isn't set to 0 anymore. I'm affraid the same problem is with Key Equations, because there we also need to have margin-bottom removed (set to 0) in the table container. Let me know  if we need to keep the Key Equation shape or both (Section Key Terms as well) as it is to tag this table in Key Equations to prevent table container from inheriting the margin form unstyled table container. Key Equations unlike Section Key Terms (present in HSPhysics only) is present in many books, so it can affect page count.

recipes/mixins/_modify.scss
```diff
@@ -979,6 +979,99 @@
...
+    @if release-flag-enabled(tag-table-parent-container) {
+      .os-table {
+        string-set: hasTable " ";
+          table:not(.unstyled):not(.top-titled):not(.column-header) {
+            string-set: hasTable " os-basic-table-container";
+              }
+        &:deferred {
+          class: attr(class) string(hasTable);
+        }
+      }
```
wilkus27, line 1059: I think that in this case we don't need `os-basic-table-container` class. General settings for basic tables can be added to selector `.os-table:not(.os-unstyled-container)`.

styles/books/hs-physics/book.scss
```diff
@@ -266,9 +266,15 @@
...
-    _selectors: (".os-table"),
+    _selectors: (".os-table.os-top-titled-container"),
     'TableTitle:::color': (_ref: "colorMap:::physicsTableColor"),
   ),
+  // SectionKeyTerms: (
+  //   _selectors: (".os-table.os-section-key-terms-container"),
+  //   'TableContainer:::margin-bottom': 0,
+  //   'Table:::margin-bottom': 0,
+  //   'TableData:::padding': table-v-spacing(1) h-spacing(3) table-v-spacing(1) 0,
+  // ),
```
wilkus27, line 277: I'm wondering is it a good idea to clean up Section Key Terms in this PR. There is separate issue for this element https://github.com/openstax/cnx-recipes/issues/2221. I think it will be easier to decide if we want to remove this shape or to apply components from table-components file after this file will be cleaned.

styles/designs/carnival/parts/_eoc-components.scss
```diff
@@ -125,12 +125,12 @@ $Exercise_Table_Row: (
...
\ No newline at end of file
+// $Key_Equations_Table_Container: (
+//     _name: "KeyEquationsTableContainer",
+//     _subselector: " > .os-table",
+//     _properties:(
+//         margin-bottom: 0,
+//         margin-left: 0,
+//         margin-right: 0,
+//     )
+// );
```
wilkus27, line 136: The same here, also there is separate issue for EOC https://github.com/openstax/cnx-recipes/issues/2040

styles/designs/carnival/parts/_table-components.scss
```diff
@@ -11,34 +11,45 @@ $TableContainer: (
...
-);
-$FakeTitleTable: (
-    _name: "FakeTitleTable",
-    _subselector: " > table.top-titled:not(.unstyled)",
-    _properties: (
-        border-collapse: collapse,
-        width: 100%,
+        border-collapse: enum('ValueSet:::OPTIONAL'),
+        margin-bottom: enum('ValueSet:::OPTIONAL'),
+        width: enum('ValueSet:::OPTIONAL'),
```
wilkus27, line 18: I'm not sure what could be better approach. To have all properties OPTIONAL or to create two components for tables which are centered and for left aligned. For example `DefaultTable` and `UnstyledTable`.

recipes/mixins/_modify.scss
ajluk: @wilkus27 Thanks! I have removed the `os-basic-table-container` class.

styles/books/hs-physics/book.scss
ajluk: @wilkus27 Yes. I think you're right. I have removed changes made for Section Key Terms and Key Equations as well.

brittany-johnson:
@ajluk Also, can you add a bullet point summary of the recipe changes you made in this PR to your PR description? 

I added the `impact: recipe.books` to the issue so that I can collect them during release. 

Thank you!

styles/designs/carnival/parts/_sectionkeyterms-shapes.scss
```diff
@@ -1,4 +1,5 @@
 @import '_sectionkeyterms-components';
+@import '_table-components';
```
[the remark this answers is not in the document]
ajluk: Yes, of course. I've missed it and left after I have reverted the changes I made to this part. Thank you for the catch!

recipes/mixins/_modify.scss
```diff
@@ -979,6 +979,99 @@
...
+    //     class: attr(class) string(hasTable);
+    //   }
+    // }
+  
+//   }
+// }
+
+// Tag table parent container which table and create it's class based on the existing table class   
+@mixin modify_tagTableParentContainer() {
+    @if release-flag-enabled(tag-table-parent-container) {
```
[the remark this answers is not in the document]
wilkus27: Since this mixin just adds additional class to container it shouldn't break anything if we apply those changes to all books. But in this case, maybe better solution is to put all kinds of tables in config in `common-stuff` file. Now they are defined in each `config.scss` file separately. Let me know what do you think :slightly_smiling_face:

styles/designs/carnival/parts/_table-shapes.scss
```diff
@@ -21,8 +24,8 @@
                                     _components: (
                                       map-merge($TableData, (
                                         _components: (
-                                            $TableDataList,
-                                            $TableDataUnstyledList
+                                            $Table__List--MerginLeftUnordered,
```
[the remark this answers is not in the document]
ajluk: Thanks for catching it up!

recipes/mixins/_modify.scss
```diff
@@ -979,6 +979,99 @@
...
+      // .os-table {
+      //   string-set: hasTable " ";
+      //     table:not(.unstyled):not(.top-titled):not(.column-header) {
+      //       string-set: hasTable " os-basic-table-container";
+      //         }
+      //   &:deferred {
+      //     class: attr(class) string(hasTable);
+      //   }
+      // }
+      @each $tableClassSelector in $Config_Table_Classes {
```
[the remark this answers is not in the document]
ajluk: Thank you! :)

styles/designs/carnival/parts/_table-components.scss
```diff
@@ -151,42 +160,42 @@ $LastTableMergedCellsRowData: (
...
-$CaptionContainer: (
-    _name: "CaptionContainer",
-    _subselector: " .os-caption-container",
-    _properties: (
-        display: enum('ValueSet:::OPTIONAL'),
-        caption-side: enum('ValueSet:::OPTIONAL'),
-        prince-caption-page: enum('ValueSet:::OPTIONAL'),
-    )
-);
+// $CaptionContainer: (
```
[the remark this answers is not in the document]
ajluk: Thank you! I have removed the commented code.

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -91,4 +91,21 @@ $Note__ListComponent--MarginLeftBottomOrdered: list--marginLeftBottom(main, " ol
...
+//Tables
+$List--Unstyled: (
+    _name: "List--Unstyled",
+    _subselector: " > ul[data-bullet-style='none']",
+    _properties: (
+        list-style-type: none,
+    )
+);
+// Lists in table cells need to have margin set withs table spacing functions
+$Table__List--MerginLeftUnordered: (
```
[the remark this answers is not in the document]
ajluk: Thank you for catching it up. I have fixed the typo.

styles/templates/template1/parts/_table-settings.scss
```diff
@@ -110,26 +122,25 @@
...
-        ),
+        _selectors: (".os-table.os-unstyled-container"),
+        // TableContainer: (
+        //     margin-bottom: table-v-spacing(2),
+        // ),
         TableData: (
-            padding: table-v-spacing(1) table-h-spacing(1), 
         ), 
         TableHeadCell: (
             padding: table-v-spacing(1) table-h-spacing(1),
```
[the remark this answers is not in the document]
ajluk: Oh, yes, that's right. I have missed it somehow. Thank you!

recipes/mixins/_modify.scss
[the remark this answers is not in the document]
brittany-johnson: Ooooo yes! I agree with you Natalia, I think that's a great idea!

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -91,4 +91,21 @@ $Note__ListComponent--MarginLeftBottomOrdered: list--marginLeftBottom(main, " ol
...
+$List--Unstyled: (
+    _name: "List--Unstyled",
+    _subselector: " > ul[data-bullet-style='none']",
+    _properties: (
+        list-style-type: none,
+    )
+);
+// Lists in table cells need to have margin set withs table spacing functions
+$Table__List--MarginLeftUnordered: (
+    _name: "Table__List--MerginLeftUnordered",
```
brittany-johnson, line 106: oh! Sorry Anna, I forgot to point out this miss-spelling as well. Sorry that I didn't catch it earlier. 
`Mergin` -> 'Margin`

recipes/mixins/_modify.scss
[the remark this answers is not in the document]
ajluk: @brittany-johnson , @wilkus27 I have moved the tables classes config set to `common-stuff.scss file`.

styles/designs/carnival/parts/_lists-components.scss
ajluk: No problem! I have fix this typo as well. Thank you!

recipes/locale/pl/_common-stuff.scss
```diff
@@ -70,6 +70,7 @@ $Config_TargetLabels: (
 $Config_Notes: ();
 $Config_UnnumberedElements: ();
 $Config_FirstElements: ();
+$Config_Table_Classes: ();
```
wilkus27, line 73: I think we should apply classes also here.

styles/output/anatomy-pdf.css
```diff
@@ -926,53 +929,52 @@ nav#toc > .os-toc-title {
...
   width: 100%; }
 
-.os-table > table.top-titled:not(.unstyled) > thead > tr > th {
+.os-table.os-top-titled-container > table > thead > tr > th {
   border-bottom: 0.1rem solid black;
   border-top: 0.1rem solid black;
   font-family: Muli;
   font-weight: 700;
-  padding: 0.7rem 0.7rem; }
+  padding: 0.7rem; }
```
wilkus27, line 952: This is again problem with short hand version of paddings. It doesn't work in `table-components` file :disappointed:

ajluk: @wilkus I have added comment about why it's a temporary solution as we discussed.

recipes/locale/pl/_common-stuff.scss
ajluk: Great catch! Thank you Natalia!

PR #2272, opened 2020-05-01
1973 - Rename ChapterOutline 

Stackblocks:
- Separate out the ChapterOutline into 
`ChaperOutline--Margin` and `ChapterOutline--Fullwidth`
- Make sure that the recipes still build with the new names
- At a future date we will be able to hard-code values
for margin or fullwidth as needed without cross-interference.


styles/templates/template1/parts/_chapterintro-settings.scss
```diff
@@ -147,18 +147,18 @@
     ),
     FullWidthChapterIntro: (
         _selectors: (".introduction"),
-        Outline: (
+        OutlineFullwidth: (
             width: 6.5in, // Width of the page minus margins
             font-family: (_ref: "typography:::sansFont"),
             font-size: font-scale(0),
+            color: (_ref: "colorMap:::chapterOutlineTextColor"),
```
ajluk, line 154: See my comment above. This `color` should be set to book specific in the book.scss files.

styles/output/physics-pdf.css
```diff
@@ -918,13 +918,13 @@ nav#toc > .os-toc-title {
...
 .introduction > .intro-body > .os-chapter-outline {
-  width: 6.5in;
-  margin: 0 0 1.5rem 0;
   font-family: IBM Plex Sans, sans-serif;
+  font-weight: 500;
   font-size: 1rem;
+  color: #FFFFFF;
   line-height: 1.5rem;
-  font-weight: 500;
-  border-bottom-color: #00819A;
```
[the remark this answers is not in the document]
Stackblocks: You're right! I missed naming the files in the book.scss-s. I updated that and the colors should be behaving now.

styles/templates/template1/parts/_chapterintro-settings.scss
Stackblocks: This was a mistake in duplicating the ChapterOutline settings over to both --margin and --fullwidth, since --fullwidth it wasn't needed for fullwidth it was causing problems.

styles/templates/template1/parts/_chapterintro-settings.scss
```diff
@@ -5,18 +5,18 @@
...
+        OutlineMargin: (
+            font-family: (_ref: "typography:::sansFont"),
+            font-weight: 500,
+            font-size: font-scale(0),
+            line-height: line-height(0),
             width: half-fullBleedsSize(2, in),
+            color: (_ref: "colorMap:::chapterOutlineMarginTextColor"),
             background-color: (_ref: "colorMap:::chapterOutlineColor"),
-            color: (_ref: "colorMap:::chapterOutlineTextColor"),
             // min-width: not_implemented(15.6rem),
```
brittany-johnson, line 16: I think we can delete the comments on line 16 and 17

styles/templates/template1/parts/_chapterintro-settings.scss
```diff
@@ -5,18 +5,18 @@
...
         _selectors: (".introduction"),
-        Outline: (
+        OutlineMargin: (
+            font-family: (_ref: "typography:::sansFont"),
+            font-weight: 500,
+            font-size: font-scale(0),
+            line-height: line-height(0),
             width: half-fullBleedsSize(2, in),
+            color: (_ref: "colorMap:::chapterOutlineMarginTextColor"),
             background-color: (_ref: "colorMap:::chapterOutlineColor"),
```
brittany-johnson, line 15: I noticed that this is also being set in the biology book.scss and the values of `chapterOutlineColor` and `biologyHeaderColor` are the same values. I would remove the setting from the biology book since it seems like the value is the same theme wide

brittweinstein:
@Stackblocks i think this can be merged

Issue #2188, opened 2020-04-09
CLEAN UP: EOC (the grand finale)


helenemccarron:
- [x] Move all shapes out of components
- [x] Move problems/exercises related components to _exercisesComponents (most of this should already be handled by the `EOCAssementShape`, verify if anything is missing for all books
- [x] Use `EOCAssessments` for all EOC exercises  (already done by Natalia in https://github.com/openstax/cnx-recipes/pull/2136, https://github.com/openstax/cnx-recipes/pull/2140, https://github.com/openstax/cnx-recipes/pull/2142, https://github.com/openstax/cnx-recipes/pull/2147)
- [ ]  Move all named tables to `_tablesComponents`, use default tables, create new components only if absolutely necessary
- [x] Fix comments: delete useless comments, document what the different shapes and components are doing, in the components, shapes and book files if necessary
- [x] Delete the EOC folder and everything in it, no longer used.
- [x] Move any remaining headers to titles file (this might be already done), I think all the headers used are now H2Title-H4Title
- [x] Remove anything not longer used in templates/template1/_eoc_settings.scss, remove the file if empty.
- [x] Clean up shapes: delete `EocExercise`, `EocExerciseWithoutSubtitle`, `EocReviewQuestions`, `EocCompChapExercises` + all commented out shapes

wilkus27:
Point about tables components will be resolved within https://github.com/openstax/cnx-recipes/issues/2040 after work from https://github.com/openstax/cnx-recipes/pull/2270 will be merged.

wilkus27:
@MartaSztuka there should be no visible changes in PDF

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-05-11 14-15-00](https://user-images.githubusercontent.com/47241114/81560600-095fc100-9392-11ea-8f12-5357cabb67ff.png)
Chemistry AF
![Screenshot from 2020-05-11 15-57-15](https://user-images.githubusercontent.com/47241114/81569860-1daaba80-93a0-11ea-8269-a2ead9312104.png)
Anatomy
![Screenshot from 2020-05-11 16-36-13](https://user-images.githubusercontent.com/47241114/81574034-8d6f7400-93a5-11ea-96f0-c52c22bcabc6.png)
Biology 2e
![Screenshot from 2020-05-11 16-37-22](https://user-images.githubusercontent.com/47241114/81574139-b6900480-93a5-11ea-98a0-3fe759e21671.png)
[...]

PR #2274, opened 2020-05-04
Clean up eoc

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2188


styles/books/anatomy/book.scss
```diff
@@ -144,6 +144,8 @@
     'Container:::column-gap': 2.4rem,
     'Container:::column-width': auto,
     'ProblemNumber:::color': (_ref: 'colorMap:::fontBodyColor'),
+    'ProblemContainerSpecific:::vertical-align': bottom,
+    'ProblemContainerSpecific:::padding-top': half-v-grid(1),
```
wilkus27, line 148: Settings from those properties were inherited from shape ContentExercises. So, even there are changes in output everything should look the same in PDF. It' better to define them separate for EOCAssesmentsShape.

Issue #2199, opened 2020-04-10
CLEAN UP: modules


helenemccarron:
Move `ModuleParagraphTitleComponent` to the titles components

ajluk:
@brittweinstein This issue will be resolved within this PR: #2279 I have only merged previous PR to another one to resolve clean up issues regarding modules and chapters together.

ajluk:
Information for testing purposes about font-family in Modules and Chapters (form Larissa):

    H2-H5 titles should be all set to Muli except H1 which should be IBM plex sans

MartaSztuka:
**HS Physics**
![Screenshot from 2020-05-14 13-55-58](https://user-images.githubusercontent.com/47241114/81934593-7e7c0200-95ef-11ea-9865-ec57dd93e098.png)
**Chemistry 2e**
![Screenshot from 2020-05-14 14-53-21](https://user-images.githubusercontent.com/47241114/81936488-b0429800-95f2-11ea-8571-f471a52cc5b6.png)
**Chemistry AF**
![Screenshot from 2020-05-14 15-03-28](https://user-images.githubusercontent.com/47241114/81937583-17ad1780-95f4-11ea-8cc2-064de2dd2ae4.png)
**Anatomy**
![Screenshot from 2020-05-14 15-17-59](https://user-images.githubusercontent.com/47241114/81939060-209ee880-95f6-11ea-9113-849db687b6af.png)
**Biology 2e**
![Screenshot from 2020-05-14 15-44-22](https://user-images.githubusercontent.com/47241114/81941971-cf90f380-95f9-11ea-996f-9c5d522f8c58.png)
**AP Biology**
![Screenshot from 2020-05-14 15-58-22](https://user-images.githubusercontent.com/47241114/81943551-d0c32000-95fb-11ea-8161-ad299139540a.png)
**Concepts of Biology**
![Screenshot from 2020-05-14 16-03-46](https://user-images.githubusercontent.com/47241114/81944103-88f0c880-95fc-11ea-866c-fde6dd5a3663.png)
**Microbiology**
![Screenshot from 2020-05-14 16-42-54](https://user-images.githubusercontent.com/47241114/81948407-fc490900-9601-11ea-89d7-95a325a9e226.png)
**College Physics**
![Screenshot from 2020-05-14 17-49-00](https://user-images.githubusercontent.com/47241114/81956151-42569a80-960b-11ea-8558-fcb51eb6325a.png)
**University Physics**
[...]

Issue #2170, opened 2020-04-07
CLEAN UP: chapters


helenemccarron:
- [x] Remove the shapes from the components
- [x] Remove the page from the shape
- [x] Move the title to the titles component 
- [x] assess if we need a shape for chapters at all

ajluk:
Information for testing purposes about font-family in Modules and Chapters (form Larissa):

H2-H5 titles should be all set to Muli except H1 which should be IBM plex sans

MartaSztuka:
Word `chapter` lack of style is caused by regression, will be fixed.
![Screenshot from 2020-05-14 15-05-52](https://user-images.githubusercontent.com/47241114/81937864-72467380-95f4-11ea-9c9e-26496c121653.png)

**HS Physics**
![Screenshot from 2020-05-14 13-52-53](https://user-images.githubusercontent.com/47241114/81934540-699f6e80-95ef-11ea-8b79-916a93ade070.png)
**Chemistry 2e** ![Screenshot from 2020-05-14 14-49-49](https://user-images.githubusercontent.com/47241114/81936231-390d0400-95f2-11ea-9272-f01479e4f94b.png) 
**Chemistry AF**
![Screenshot from 2020-05-14 15-02-58](https://user-images.githubusercontent.com/47241114/81937535-08c66500-95f4-11ea-9830-c7c2c7fd3d37.png)
**Anatomy**
![Screenshot from 2020-05-14 15-17-21](https://user-images.githubusercontent.com/47241114/81939003-0bc25500-95f6-11ea-9af3-9f3c0d5c9f47.png)
**Biology 2e**
![Screenshot from 2020-05-14 15-43-52](https://user-images.githubusercontent.com/47241114/81941922-bee07d80-95f9-11ea-8bc5-6ecf00234af9.png)
**AP Biology**
![Screenshot from 2020-05-14 15-57-41](https://user-images.githubusercontent.com/47241114/81943436-abcead00-95fb-11ea-96b4-7272d6c2f1fa.png)
**Concepts of Biology**
![Screenshot from 2020-05-14 15-59-42](https://user-images.githubusercontent.com/47241114/81943893-3fa07900-95fc-11ea-8229-d1bb302e40bc.png)
**Microbiology**
![Screenshot from 2020-05-14 16-40-55](https://user-images.githubusercontent.com/47241114/81948363-f05d4700-9601-11ea-82e0-3b5dfdd29521.png)
**College Physics**
[...]

PR #2279, opened 2020-05-06
Clean chapter components

ajluk:
Issues:  #2170, #2199

I have removed chapter-components/shapes/settings and modules-components/shapes/settings files, because they're not need anymore.
The titles have been moved to title-components.scss file and based on discussion with Natalia, BrittanyJ and Alan the new shape for Module and Chapter titles together was created.
I have also added within this PR H5 components to the modules based on the comment in title-components.scss file and removed accidentally added SolutionTitle setting in some book.scss files form Modules (it is only used by Exercises shape) and added missing properties to ExerciseSolutionTitle component  (probably accidentally deleted as well).

Information about font-family in Modules and Chapters (form Larissa):
- H2-H5 titles should be all set to Muli except  H1 which should be IBM plex sans


styles/designs/carnival/parts/_titles-shapes.scss
```diff
@@ -0,0 +1,37 @@
...
+                        $SubSection__ListComponent--MarginLeftOrdered,
+                        $TitleOverFigure,
+                        )
+                        )),
+                map-merge($TitleContainer, (
+                    _components: (
+                        $TitleTextComponent,
+                        $TitleNumberComponent,
+                        )
+                    )),
```
wilkus27, line 33: Hmm it doesn't change anything, but couldn't be this part move at the beggining, before `map-merge($ModuleContainerComponent, (`? Then shape will have structure as markup in HTML (first is chapter title then module). I think it's quite confusing now :thinking:

styles/templates/template1/parts/titles-settings.scss
```diff
@@ -0,0 +1,41 @@
...
+            //color: book specific, defined in book.scss file
+        ),
+        TitleText: (
+            display: block,
+            font-family: (_ref: 'typography:::sansFont'),
+            font-size: font-scale(7),
+            line-height: line-height(7),
+            font-weight: lighter,
+            
+        ),
```
wilkus27, line 39: Same here. Can those components be moved at the beginning?

styles/designs/carnival/parts/_titles-components.scss
```diff
@@ -181,3 +182,23 @@ $H5Component: (
...
+    _name: "TitleOverFigure",
+    _subselector: ' h4 + div.os-figure > figure',
+    _properties:(
+        padding-top: v-spacing(1),
+    )
+);
+
+//Modules and chapters empty containers
+$ModuleContainerComponent: empty_wrapper(Container, ' > [data-type="page"]:not(.introduction)');
+$ChapterContainerComponent: empty_wrapper(Container, '');
```
wilkus27, line 204: I think the names should be different. For chapter is ok, because it's main parent. And it is consistent with other files. But for module maybe it could be just `Module`? The final name created by `empty_wrapper` function will be `ModuleWrapper`.

styles/books/anatomy/book.scss
```diff
@@ -21,11 +21,14 @@
     'H5Title:::color': (_ref: "colorMap:::anatomyHeaderColor"),
   ),
 ));
-
 @include add_settings((
-  Chapter: (
+  ChapterWithModuleTitles: (
```
wilkus27, line 25: I'm wondering about name. When Helene created shape for titles in EOC she named it `GenericEOC`. What do you think to use this convention and name it `GenericChapter` or something like that?

ajluk: @wilkus27 maybe it would be a good idea to add a word title to the shape name in this case ad as well to `GenericEoC`. Something like `GenericChapterTitles`, `GenericEoCTitles`. I was also wondering if the word `Generic` isn't misleading regarding the titles shapes name. There is a file `_generic-components`, but I think that titles shaoes should rather be placed in the `titles-shapes` file. What do you think? @brittany-johnson , @Stackblocks what is your opinion? Should we keep the generic naming  convention here?

Stackblocks: I agree that giving the name context is important, as `GenericEOC` is a little vague. As for the placement of generics, we may want to rename the `_generic-components` file, as the only thing in it currently is a `SectionComponent--MarginBottom`, which leaves things like `GenericTitle` free to be placed in the Titles file, etc.

styles/designs/carnival/parts/_titles-shapes.scss
Stackblocks: I agree with Natalia. I feel that where possible the structure of the shapes should mirror the actual thing.

styles/designs/carnival/parts/_exercises-components.scss
```diff
@@ -119,5 +119,8 @@ $Exercise__Solution--Title:(
     _properties:(
         font-family: enum('ValueSet:::REQUIRED'),
         font-weight: bold,
+        //these properties above were probably accidentally removed during the changes that has been made to exercises shapes
```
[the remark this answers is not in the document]
ajluk: Yes, that's right. It was only left for code review.

styles/designs/carnival/parts/_titles-components.scss
ajluk: Yes, you're right. I have also changed the `$ChapterContainerComponent` name to ChapterWrapper, so it won't be cofusing.

styles/templates/template1/parts/titles-settings.scss
```diff
@@ -0,0 +1,42 @@
+
```
brittany-johnson, line 1: this is not a big deal, but there are three lines of empty space here

styles/books/anatomy/book.scss
brittany-johnson: I think the name `Generic` may be a little misleading as well because we have only one shape now for all titles. I think the name `ChapterAndModuleTitles` or `ChapterModuleTitles` could work? What do you all think?

ajluk: I have renamed the shape name to `ChapterModulesTitles` as BrritanyJ suggested above and keeped the `titles-components` and `titles-shapes` files. I have also left the `generic-compoents` file as it was. There is only one component there but after the cleaning maybe can be more, so we can decide if we want to keep that file in the future.

Issue #2178, opened 2020-04-08
CLEAN UP: create a function.scss file, refactor all functions and store them in that file


helenemccarron:
Move all helper functions to a commonplace so that they can be used across files.
- [x] Implement the `empty_wrapper` function,  use it everywhere we have an empty container.
```
//in _functions.scss
@function empty_wrapper($name, $selector) {
   @return (
     _name: "#{$name}Wrapper", // IntroBodyWrapper
     _subselector: $selector, // > .intro-body 
     _properties: ()
   );
 }
 ```
```
//in _chapterintroComponent.scss
 $ChapterIntroBody: empty_wrapper(IntroBody, ' > .intro-body');
 ```
```
//in _chapterIntroShape
...
 … map-merge($ChapterIntroBody, () ….

```

- [x]  move `titleBottomBorder` from `chapterIntroComponents` function to funtions.scss

wilkus27:
Parts of book, where new function is applied:

- Answer Key
- Introduction page
- Exercises
- Learning Objectives
- Module
- References
- Section Key Terms
- Tables
- Titles
- TOC
- Notes

I think it will be ok to test just one book which contains most of those parts.

MartaSztuka:
Chemistry 2e
![Screenshot from 2020-05-08 10-28-03](https://user-images.githubusercontent.com/47241114/81387298-be864500-9116-11ea-9f5d-7bd0ae66503a.png)
![Screenshot from 2020-05-08 10-29-14](https://user-images.githubusercontent.com/47241114/81387334-ca720700-9116-11ea-864a-acd99f810196.png)
![Screenshot from 2020-05-08 10-30-31](https://user-images.githubusercontent.com/47241114/81387435-f8574b80-9116-11ea-8390-f961d54a0792.png)
![Screenshot from 2020-05-08 10-32-32](https://user-images.githubusercontent.com/47241114/81387619-43715e80-9117-11ea-993d-5cc680f7bb31.png)
![Screenshot from 2020-05-08 10-31-31](https://user-images.githubusercontent.com/47241114/81387622-44a28b80-9117-11ea-80ff-ed3a06ca9398.png)
![Screenshot from 2020-05-08 10-34-16](https://user-images.githubusercontent.com/47241114/81387762-86333680-9117-11ea-907e-26f09f16ed47.png)

PR #2284, opened 2020-05-07
Clean up functions

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2178


styles/designs/carnival/parts/_chapter-components.scss
```diff
@@ -37,7 +37,7 @@ $chapterTitle: (
   )
 );
 
-$chapterContainer: (
+$chapterContainer: (   //ToDo: apply empty_wrapper function
```
wilkus27, line 40: I just left a comment, because this component has nested component and it will be clean up within another issue.

Issue #2227, opened 2020-04-15
CLEANUP: Learning Objectives


brittany-johnson:
[Cleanup doc](https://docs.google.com/document/d/10qrKjBt3sGoZR8KaW1xwUSApVZiRPtFxce2WkRP2LJ0/edit#heading=h.3i7y3p43cq0m)

- [x] Investigate which learning objectives we have to verify which shapes we need ( see the doc for more details)

- [x] Can $LearningObjectiveTitleComponent be replaced with a component from the titles file?

- [x] Empty NoteBody can be replaced by no props function

- [x] The name of the shapes should match the component variable names (or at least be closer to the name)

- [x] Check if the lists components can be moved to _list-components.scss file

wilkus27:
### Decision

Learnig Objectives in all books should be styled in one way, like it is done in HS Physics.
(Link to discusion: https://katalysteducation.slack.com/archives/CBW9DFKGF/p1588938854198800)

Therefore, some of the content changes are needed:

- add title "Learning Objectives" in AP Biology
- wrap text inside `[data-type="abstract"]` into `<p>` in Biology, Chemistry, Anatomy, U-Physics, Pl-U-Physics
- add `<p>` with text "By the end of this section..." in Microbiology

Those changes will be done on production. On katalyst01 it will be changed only one chapter in each book.

wilkus27:
@MartaSztuka content is changed in Anatomy in module `Overview of Anatomy and Physiology`. There it should display properly. Everywhere else some parts of Learning Objectives could be unstyled.

Please, create content issue for polish u-physics in `openstax-poland/content` repo.

Here is the needed change.
Text from the screenshot should be wrapped into `<p>`:
![image](https://user-images.githubusercontent.com/40228854/82031471-39f97080-969a-11ea-84e2-2bf0d2431983.png)

wilkus27:
### Recipe changes
Add title "Learning Objectives" in Biology, Anatomy and Chemistry in their `config.scss` files.

MartaSztuka:
Anatomy
![Screenshot from 2020-05-15 16-37-48](https://user-images.githubusercontent.com/47241114/82062467-70e77a80-96ca-11ea-8f0f-769103ddc4a3.png)
Chemistry 2e
![Screenshot from 2020-05-15 15-42-51](https://user-images.githubusercontent.com/47241114/82056834-c91a7e80-96c2-11ea-899e-b420d3fcea5d.png)
Chemistry AF
[...]

PR #2293, opened 2020-05-08
Clean Up Learning Objectives

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2227


recipes/books/anatomy/_config.scss
```diff
@@ -162,3 +162,5 @@ $Config_FirstElements: (
   (selector: ' > .os-figure:first-child'),
   (selector: ' > div:first-child > p:first-child img:first-child'),
 );
+
+$Config_abstractTitle: "Learning Objectives";
```
wilkus27, line 166: I noticed that for each book which has `[data-type="abstract"]` title should be added:

image

But it has to be defined in `config.scss` file. It was missing in Anatomy, Biology and Chemistry.

brittweinstein:
@ajluk we can merge this :)

Issue #2114, opened 2020-03-26
Footnote in note that breaks across pages


lwtchu:
# What happened?
When a footnote call out appears within a paragraph or table that breaks across pages - it gets pushed to the second page even if it's on the first page.

Also when there's no more room at the bottom, it breaks.

~- within a note~
~- the note breaks across two pages~
~- it appears before the page break~

TO SEARCH: look for the word `swine farms` 🐷 , `ecosystem environment`, `a study in jordan`
 
<img width="715" alt="Screen Shot 2020-03-26 at 9 47 17 AM" src="https://user-images.githubusercontent.com/11299125/77660402-450c0c00-6f47-11ea-8b2b-2ca67ba1a1a5.png">
<img width="598" alt="Screen Shot 2020-03-26 at 9 47 12 AM" src="https://user-images.githubusercontent.com/11299125/77660409-45a4a280-6f47-11ea-84e3-c975448b60ab.png">

# Reproducibility Details

- **Does it also happen in an incognito window?**: 
- **Does this happen in one place, a few places, or everywhere?**:

# Environment:

- **Book(s)**:
- **URL/Server**: 
- **Location (page number, chapter/section)**:
- **Suggested nearby searchable text**: 
- **Browser**:

# Screenshots

(_attach here (please clearly label what environment each screenshot came from)_)

# Additional notes

 (_For example, steps to reproduce, related issues, more context, etc._)

lwtchu:
frog scream: i found a case where it's working
<img width="553" alt="Screen Shot 2020-03-26 at 10 56 04 AM" src="https://user-images.githubusercontent.com/11299125/77667594-6de4cf00-6f50-11ea-9820-60dca460b2de.png">

lwtchu:
<img width="554" alt="Screen Shot 2020-03-26 at 10 56 59 AM" src="https://user-images.githubusercontent.com/11299125/77667687-894fda00-6f50-11ea-90a9-c1a54cf33a2f.png">
found this instance. 

i think it's more of when the paragraph breaks

lwtchu:
<img width="590" alt="Screen Shot 2020-03-26 at 11 02 22 AM" src="https://user-images.githubusercontent.com/11299125/77668659-ae911800-6f51-11ea-9344-38009f653357.png">
also footnotes get pushed if there's no room

brittweinstein:
[...]

PR #2302, opened 2020-05-12
[WIP] Issue 2114

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2114


wilkus27:
@brittany-johnson I found out why it doesn't work.

This is how footnotes looks like in baked html:
image

This text is inside tag `aside[role="role-docfootnote"]` and for this tag we applied property `float: footnote`. So in this case Prince moves tag to the end of page and instead creates footnote call in this place (which we remove in styles by `content: none` property).

And theoretically it's working:

image

Footnote is at the same page as footnote call, but this one created by Prince, not our in `a[role="doc-noteref"]`.

brittany-johnson:
It took some time for me to be able to work on this again, sorry for the delay! 

As a test I removed the `content: none` from `aside[role="doc-footnote"]::footnote-call` and added `prince-footnote-policy: keep-with-line` to `div.os-note-body > p`. The result is the screenshot below. 

I see what you mean that Prince is still making a place for the footnote that we moved with `float: footnote`, but Prince's footnote and our footnote are both on the following page instead of the page where the call occurs. 

Screen Shot 2020-05-19 at 3 42 57 PM

wilkus27:
@brittany-johnson on the first screenshot this `1` at the bottom of the note is not a place for the footnote that we moved with `float`, it is a footnote call automatically created by Prince. So it's working as intendent - footnote call is on the same page as footnote marker. 

Here is how prince is working: (https://www.princexml.com/doc/footnotes/#footnotes).

image

This blue `1` on your screenshot it is a footnote call `hard coded` by us, we create it by recipes.

wilkus27:
This is our markup in baked html:

image

So when we apply `float: footnote` for `aside` and run Prince, it moves `aside` to the end of the PDF page and creates in this place footnote call. Prince ignores `a[role=doc-noteref]` because it treats it as regular element, not part of the footnote.

wilkus27:
If we want to use all footnotes properties dedicated by Prince `aside` should be in text next to the `a[role=doc-noteref]`. Then footnote call created by Prince will be at the same place as footnote call hard coded by us. But it requires additional work in Rex :disappointed: Another solution could be do not allow page breaks inside notes, but then there is a lot of empty spaces in the PDF.

brittany-johnson:
@wilkus27 I suspected we would need a recipe change but I was hoping with all of my heart we somehow wouldn't LOL!! 

I will move this to the icebox until the current footnote work is at least on production. 

Thank you so much again Natalia for investigating this with me!!

MartaSztuka:
##### PASSED

PR #2305, opened 2020-05-14
Clean section key terms

ajluk:
Issue: #2221


ajluk:
After conversation with Larissa (https://katalysteducation.slack.com/archives/CBW9DFKGF/p1589456742238300) I have removed all the section key terms files, because right now after tables cleaning it is using the Unnumbered Unstyled table shape and the look of the table isn't much different than whe we have separate styles for it. The only differences were margin-bottom of the table which previously was set to 0 and padding-left of the table data which previously was set to 0 as well. This is how it looks now in the html file: 
image and pdf:

brittweinstein:
@ajluk we can merge this one!

Issue #2220, opened 2020-04-14
CLEANUP: References


brittany-johnson:
[Cleanup doc](https://docs.google.com/document/d/10qrKjBt3sGoZR8KaW1xwUSApVZiRPtFxce2WkRP2LJ0/edit#heading=h.njxq5e8rshj0)

- [x] How are titles working if the titles aren’t imported? (Test / fix)

- [x] Delete the two commented out components ($ReferencesTitleComponent & $ReferencesSubtitleComponent) 

- [x] Should $ReferencesTextComponent be made into two columns? Ask Andy

ajluk:
I have checked how the H1, H2 titles work and they should be kept in the shape. The current look after Helene's cleaning of the titles I have consulted with Larissa and it doesn't need any modification , so I  only removed commented components.

ajluk:
Information for testing purposes: There were no changes in the design of References within this PR. It should look the same as it was previously.

MartaSztuka:
Anatomy
![Screenshot from 2020-05-19 11-58-54](https://user-images.githubusercontent.com/47241114/82313063-2450b780-99c8-11ea-9f64-183ad0c144c1.png)
![Screenshot from 2020-05-19 12-07-50](https://user-images.githubusercontent.com/47241114/82313971-6a5a4b00-99c9-11ea-8903-2325db932d2b.png)

PR #2306, opened 2020-05-15
remove commented compoents

ajluk:
Issue: #2220


ajluk:
I have checked how the H1, H2 titles work and they should be kept in the shape. The current look after Helene's cleaning of the titles I have consulted with Larissa and it doesn't need any modification , so I only removed commented components.

Issue #2196, opened 2020-04-10
CLEAN UP: Figures and images


helenemccarron:
- [x] Rename components to follow our conventions
- [x] Fix the `PicCaption`,  you may be able to solve the problem using a generic function that removes the empty element that is problematic.
- [x] Rework the `figureContainer` to be "pure", 
   - a "pure component" does not define global styles for all the components nested inside, only what governs the properties of the component itself (like the ability to convert it into columns)
  - Remove and move to inner components the following:
       ```font-size: font-scale(-1),
       line-height: line-height(0),
       margin-bottom: v-spacing(1),```

ajluk:
@MartaSztuka >>> Information for testing purposes: There should be no visible changes in Figures after the cleaning.

MartaSztuka:
@ajluk Figures become left aligned while should be centered.
AP Biology
before:
![Screenshot from 2020-05-25 13-54-55](https://user-images.githubusercontent.com/47241114/82812384-2a94d700-9e93-11ea-9f89-b69e1623590e.png)
after:
![Screenshot from 2020-05-25 14-21-49](https://user-images.githubusercontent.com/47241114/82812391-2e285e00-9e93-11ea-9857-a98e93ee3f49.png)

MartaSztuka:
Chemistry 2e![Screenshot from 2020-05-29 16-44-30.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/1c017c22-bc18-4154-92ad-831b05e35866)
Chemistry AF![Screenshot from 2020-05-29 16-47-15.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/73ec5aca-69d9-42b2-8955-f67ec49492fd)
Anatomy![Screenshot from 2020-05-29 16-49-37.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6b13987c-0b1c-4eb0-b219-3c24235a8fe6)
Biology 2e![Screenshot from 2020-05-29 16-50-46.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/bf986e06-0419-4d46-929f-ac52208a1e79)
AP Biology![Screenshot from 2020-05-29 16-51-49.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/bb141a10-09c1-451b-8c39-30453df7403d)
[...]

PR #2307, opened 2020-05-15
Cleanup figures

ajluk:
Issue: #2196


styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -8,20 +8,23 @@ $CaptionContainer: (
...
-$CaptionTitleLabel: (
+$Caption__TitleLabel: (
     _name: "CaptionTitleLabel",
     _subselector: " > .os-title-label",
     _properties: (
         font-family: enum('ValueSet:::OPTIONAL'),
         font-weight: bold,
         color: enum('ValueSet:::OPTIONAL'),
+        font-size: enum('ValueSet:::OPTIONAL'), // used in Figures (left OPTIONAL, need to be checked if it is also needed f.ex. in Chapter Intro)
+        line-height: enum('ValueSet:::OPTIONAL'), // used in Figures (left OPTIONAL, need to be checked if it is also needed f.ex. in Chapter Intro)
```
wilkus27, line 23: Is it necessary to define `font-size` and `line-height` for each part of caption? Couldn't be added to `$Caption__Container` which is the parent of those elements? Then they should inherit these properties.

styles/designs/carnival/parts/_figureimage-components.scss
```diff
@@ -48,7 +40,7 @@ $FigureTitle: ( //used onlu in Chapter Intro components need to be reame cleane
     )
 );
 
-$Figure:(
+$Figure__Component:(
     _name: "Figure",
```
wilkus27, line 44: I'm wondering about naming convention :thinking:  In my PR with exercises clean up I move one more component to this file:

image

So going with this convention I should name it `Figure__Image`. I'm not sure if it's ok. Sometimes images are inside figures, but not always. There are also images without figure parent. So it will be misleading. 

What do you everybody think? I don't have an idea how it could be named differently.

Maybe we could use name from the file name, so it will be `FigureImage__Container`, `FigureImage__Figure`, `FigureImage__Image`. But I think it is also odd :smiley:

Stackblocks: So for BEM it would be `Image` for the base image, and `Image--Figure` for the image as part of a figure, because it is still an image, but is an image of a specific and different type.

styles/designs/carnival/parts/_caption-components.scss
ajluk: Yes, that would be much simpler. I thought we want to remove those properties from this container as well.

styles/designs/carnival/parts/_figureimage-components.scss
ajluk: I decided to set this component name to just `Figure`, because it is the `figure` tag itself. Then the `image` component can be named as Alan suggesting above to just `Image` for images that aren't part of the figures and `Image--Figure` or maybe `Figure__Image` in the case when they are inside a `figure`. What do you think? :)

styles/designs/carnival/parts/_table-shapes.scss
```diff
@@ -103,10 +103,10 @@
                         )),
                     )
                 )),
-                map-merge($CaptionContainer, (
+                map-merge($Caption__Container, (
                     _components: (
-                        $CaptionTitleLabel,
-                        $CaptionNumber,
+                        $Caption__TitleLabel,
```
Stackblocks, line 108: Should `Caption__TitleLabel` be decomposed further into either `...Title__Label` or `...Title--Label`?

brittweinstein:
@ajluk we can merge this one too

Issue #2167, opened 2020-04-07
CLEAN UP: answer key


helenemccarron:
- [x] Remove the `exerciseComponentsDisplay` function, move it to a generic function common folder of some sort if it's still needed.
- [x]  Replace the function with a generic exercise component imported from exercises components
- [x] Remove `AnswerKeyImageComponent` and replace with generic `image` component from the figures/images components. 
- [x] Same with number, is this number necessary, can it be rolled into a generic component?
- [x] Move the shapes out of the component

wilkus27:
@MartaSztuka during clean up I added padding to images/figures which are first elements of exercises/solutions. This padding was defined in code, but by old, not existing function. So please check those places carefully, if it doesn't break anything. This changes, as I said,  affect not only solutions in Answer Key, but also exercises (from regular modules and EOC). Please check them as well.

MartaSztuka:
AP Biology
![Screenshot from 2020-05-26 14-58-19](https://user-images.githubusercontent.com/47241114/82903521-5bdbd880-9f61-11ea-8146-9aab9e5b46e8.png)
Chemistry 2e
![Screenshot from 2020-05-26 15-25-55](https://user-images.githubusercontent.com/47241114/82906252-3f419f80-9f65-11ea-8473-e183205dc278.png)
Chemistry AF
![Screenshot from 2020-05-26 16-31-33](https://user-images.githubusercontent.com/47241114/82913289-6bade980-9f6e-11ea-883e-ef2834954173.png)
HS Physics
![Screenshot from 2020-05-26 15-41-05](https://user-images.githubusercontent.com/47241114/82907784-5bded700-9f67-11ea-957f-25378b3165d2.png)
College Physics
![Screenshot from 2020-05-26 16-21-56](https://user-images.githubusercontent.com/47241114/82912181-0efdff00-9f6d-11ea-9bb7-2b674b748dce.png)
University Physics
![Screenshot from 2020-05-26 16-38-05](https://user-images.githubusercontent.com/47241114/82914029-51c0d680-9f6f-11ea-8476-9334e09b858d.png)

Issue #1983, opened 2020-02-28
Symmetry Feedback: Text Cutoff When Figure Follows Linked Section Header


openstaxalina:
# Scenario

Note that the bottom of the "p" in "Displacement" doesn't appear:
![Screen Shot 2020-02-28 at 5 51 55 PM](https://user-images.githubusercontent.com/7863733/75595720-23f0f200-5a53-11ea-9761-9caa5e81ab21.png)

# Acceptance Criteria (required)

# Test Plan

lwtchu:
Is this tied to the no line-height issue mentioned in #1978

ajluk:
The space under the section links will be resolved in this PR: https://github.com/openstax/cnx-recipes/pull/2014. Now it looks like this:
![image](https://user-images.githubusercontent.com/55203598/75988398-cc8ad000-5ef1-11ea-8028-9315af78b690.png)

openstaxalina:
Here is another test case for this, from HS Physics:
![Screen Shot 2020-03-05 at 3 22 38 PM](https://user-images.githubusercontent.com/7863733/76026894-44a6c500-5ef5-11ea-95f5-7bbf3a35d806.png)

ajluk:
I have added line-height: line-height(3) property (it gives line-height:3rem in the output) to the the section links but while it resolves this issue with images, when there are few lines in the section title links it looks poor: 
![image](https://user-images.githubusercontent.com/55203598/76073564-18914f80-5f9a-11ea-84b4-29a6eb93a845.png). The line-height before my changes for these elements was set to line-height(1) = 1.5 rem (inherited form the root). I tried to set it to line-height(2), but it gives the same output (1.5rem).

ajluk:
I guess this case https://github.com/openstax/cnx-recipes/issues/1983#issuecomment-595454555 is connected to works on the epic issue: https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/2015 and especially: https://github.com/openstax/cnx-recipes/issues/1956.

ajluk:
Now in College Physics the image under section title link looks like this: 
![image](https://user-images.githubusercontent.com/55203598/76104847-56ab6500-5fd4-11ea-9db8-d82e83b7a46f.png)

ajluk:
Changes in TDS for this issue are already included in #1978 (they are realted).

MartaSztuka:
[...]

PR #2308, opened 2020-05-15
Answer Key clean up

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2167


styles/designs/common/pdf/_functions.scss
```diff
@@ -46,7 +46,6 @@
       _properties:(
           display: table-cell,
           vertical-align: bottom,
-          padding-top: half-v-grid(1),
```
wilkus27, line 49: I removed this because `half-v-grid` function doesn't exist anymore. I checked and this property was added to solve issue: https://github.com/openstax/cnx-recipes/issues/1983 in HS Physics. But since Helene added bottom margin to section titles during her work on titles, this padding-top from above is not necessary now.

styles/output/hs-physics-pdf.css
```diff
@@ -2289,11 +2289,11 @@ iframe {
...
 .os-concept-container [data-type='problem'] > .os-number {
   display: table-cell;
+  text-decoration: none;
   color: #000000;
-  font-weight: bold;
-  text-decoration: none; }
+  font-weight: bold; }
 
-.os-concept-container [data-type='problem'] .os-divider {
+.os-concept-container [data-type='problem'] > .os-divider {
```
ajluk, line 2296: @wilkus27 I have only a question about changing this selector. Is there a reason that it was more general before?

wilkus27: @ajluk I'm not sure. I checked and everything seems to be ok in PDF, but I will ask Marta to pay attention on this element during testing.

brittweinstein:
@wilkus27 we can go ahead and merge

Issue #2202, opened 2020-04-10
CLEAN UP: change conventions in the framework


helenemccarron:
Remove the warnings to allow for BEM notation

lwtchu:
a) we did this
b) it's ..... significantly old
-alan cooke

PR #2309, opened 2020-05-15
add connector characters to characters list

ajluk:
Issue:  #2202


ajluk:
@brittany-johnson I have added you to this review, because I'm not sure if this solution is acceptable.

brittany-johnson:
This looks really good @ajluk ! I ran it locally and it's nice not seeing all of those warnings.

styles/framework/design-shape/_init.scss
```diff
@@ -122,10 +122,15 @@ $FRAMEWORK_SPACE_PLACE: ();
       @return $component_properties;
     }
 
-    @if not is_camel_case($component_name) {
-      @warn 'Convention for component names is CamelCase. Given `#{$component_name}`.'; // coverage:disable-line
+    // @if not is_camel_case($component_name) {
+    //   @warn 'Convention for component names is CamelCase. Given `#{$component_name}`.'; // coverage:disable-line
+    // }
+
+    @else if not is_bem_camel_case($component_name) {
```
wilkus27, line 129: Why do you use here `else`? Previously there was just `if`.

styles/framework/lint/_lint.scss
```diff
@@ -37,7 +41,20 @@
...
+    }
+  }
+  @if not str_is_upper(char_at_index($string, 1)) {
+    @return false;
+  }
+  @return true;
+}
+
+@function is_bem_camel_case($string) {
+  $allowed: join(join(join(upper_chars(), lower_chars()), numeric_chars()), connector_chars(),);
```
wilkus27, line 57: What are the cases, where we still need just Camel Case, not Bem? I see you left both functions.
Also I noticed that Camel Case is used in a few other places. Shouldn't those mixins be also modify in some way? :thinking: 

image

image

Screenshots are from the `init.scss` file from above.

styles/framework/design-shape/_init.scss
ajluk: I have left it by accident. Already removed.

styles/framework/lint/_lint.scss
ajluk: Yes I have left this function, because it supports creating Shapes and Subshapes names, here we currently don't use BEM.

lwtchu:
@wilkus27 @brittany-johnson hi! this seems to be an old PR relic of the past - is this something we should close out or merge?

PR #2310, opened 2020-05-15
2201-rename-components

Stackblocks:
- Remove "Component" from component names
- Rename components using BEM syntax
- Align shapes with the updated component names


brittany-johnson:
Documentation of how we're using BEM syntax with an example should be added to the documentation. The docs in the repo need to be cleaned up but just adding info on BEM would be good for now. Thanks!
Either here: https://github.com/openstax/cnx-recipes/blob/master/styles/README.md
Or here: https://github.com/openstax/cnx-recipes/blob/master/styles/styling-bakedpdf.md

Stackblocks:
Ok. So i don't have to turn it up later, here is the source article we've been using for BEM: https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -176,17 +176,17 @@ $ChapterObjectivesListItem: (
...
     _properties: (
         color: enum('ValueSet:::REQUIRED'),
     )
 );
 
-$ChapterIntroContent: empty_wrapper(IntroContent, ' > .intro-text');
+$ChapterIntro__Content: empty_wrapper(IntroContent, ' > .intro-text');
 
-$ChapterIntroContentHeader: (
+$ChapterIntro__Content--Header: (
```
brittany-johnson, line 189: I think this should be written as ` $ChapterIntro__Content__Header`

styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -57,41 +57,41 @@
...
-$SubSubSection__ListComponent--MarginLeftUnordered: list--marginLeft(main, " > section > section > section > ul");
-$Nested__ListComponent--MarginLeftUnordered: list--marginLeft(nested, " > li > ul");
-$Nested__ListComponent--MarginLeftOrdered: list--marginLeft(nested, " > li > ol");
-$Summary__Section__ListComponent--MarginLeftUnordered: list--marginLeft(main, " section > ul");
-$Summary__Section__ListComponent--MarginLeftOrdered: list--marginLeft(main, " section > ol");
-$SubSection__Container__ListComponent--MarginLeftBottomUnordered: list--marginLeft(main, " > section > section > div > ul");  //added as a temporary solution for polish University Physics, before we start working on lists cleaning 
+$List--MarginLeftUnordered: list--marginLeft(main, " > ul");
+$List--MarginLeftOrdered: list--marginLeft(main, " > ol");
+$Section--List--MarginLeftUnordered: list--marginLeft(main, " > section > ul");
+$Section--List--MarginLeftOrdered: list--marginLeft(main, " > section > ol");
```
brittany-johnson, line 63: I think for these lists in sections (ln 62 & 63) it should be `$Section__List--blah` 

Also all of these lists, I think we should consider renaming the `$List` part to include the type of list. For example, line 63 would read `$Section__OrderedList--MarginLeft` since the type of list, unordered or ordered is the element type, not a css property. 

What do you think @ajluk and @Stackblocks

Stackblocks: In that case I think it would be `$Section__List--Ordered--MarginLeft` because it is a section that contains(__) a List that is (--) an Ordered List, which is (--) a MarginLeft-type Ordered list." 

Does that sound right to y'all?

styles/designs/carnival/parts/_chapter-components.scss
```diff
@@ -0,0 +1,48 @@
+$Chapter__Title__Text: (
```
brittany-johnson, line 1: This file was deleted here: https://github.com/openstax/cnx-recipes/pull/2279/files?file-filters%5B%5D=.css&file-filters%5B%5D=.map&file-filters%5B%5D=.scss

styles/designs/carnival/parts/_chapter-shapes.scss
```diff
@@ -0,0 +1,13 @@
+@import 'chapter-components';
```
brittany-johnson, line 1: This file was deleted here: https://github.com/openstax/cnx-recipes/pull/2279/files?file-filters%5B%5D=.css&file-filters%5B%5D=.map&file-filters%5B%5D=.scss

styles/designs/carnival/parts/_module-components.scss
```diff
@@ -0,0 +1,17 @@
+$Module__Container: empty_wrapper(Container, '');
```
brittany-johnson, line 1: This file was deleted here: https://github.com/openstax/cnx-recipes/pull/2280/files

styles/designs/carnival/parts/_module-shapes.scss
```diff
@@ -0,0 +1,33 @@
+@import 'module-components';
```
brittany-johnson, line 1: this file was deleted here: https://github.com/openstax/cnx-recipes/pull/2280/files

styles/designs/carnival/parts/_lists-components.scss
ajluk: @Stackblocks @brittany-johnson  I think you both are right about using the convention :) . As I understood from the BEM documentation we should use "__" separator for children elements and  "--" separator for modifiers (specific css properties added to this elements which they are characterized by) or variations (in this case ordered list can be a variation of the list components if we use word `List` in the name). So it could be either `$Section__List--Ordered--MarginLeft` or `$Section__OrderedList--MarginLeft` depends what convention we prefer to use. I would rather prefer the second proposal:  $Section__OrderedList--MarginLeft` because ordered list is also an element itself so it will be less confusing and shorter.

styles/designs/carnival/parts/_indent-components.scss
```diff
@@ -1,6 +1,6 @@
-$Indent__Container: empty_wrapper(Container, '');
+$Indent--Container: empty_wrapper(Container, '');
```
ajluk, line 1: I guess these components should also rather use "__" as a separator as it was done before.

styles/designs/carnival/parts/_indent-components.scss
```diff
@@ -1,6 +1,6 @@
...
 
-$Indent__Paragraph: (
+$Indent--Paragraph: (
```
ajluk, line 3: See the comment above.

styles/designs/carnival/parts/_indent-components.scss
```diff
@@ -9,4 +9,4 @@ $Indent__Paragraph: (
     ),
 );
 
-$Indent__Section: empty_wrapper(Section, ' > section');
+$Indent--Section: empty_wrapper(Section, ' > section');
```
ajluk, line 12: Same here.

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -230,7 +230,7 @@ $IconNoteParagraphTitleComponent: (
   )
 );
 
-$IconNoteDivTitleComponent: (
+$Note--Icon__Title--Div: (
```
ajluk, line 233: Should it be rather `$Note--Icon__Div__Title` in this case?

styles/designs/carnival/parts/notes/_notes-components.scss
```diff
@@ -11,14 +11,14 @@ $NoteSubtitleComponent: (
   )
 );
 
-$NoteBodyComponent:  empty_wrapper(Body, ' > .os-note-body');
+$Note__Body:  empty_wrapper(Body, ' > .os-note-body');
 
-$NoteProblemComponent: empty_wrapper(Problem, ' > [data-type="problem"]');
+$Note--Problem: empty_wrapper(Problem, ' > [data-type="problem"]');
```
ajluk, line 16: Shoud it be rather `$Note__Problem` here and $Note__Problem__Container in line 18?

styles/designs/carnival/parts/_toc-components.scss
```diff
@@ -227,15 +216,26 @@ $TocContentsTitleWord: (
...
+  _subselector: ' > ol',
+  _properties: (
+    list-style: none,
+    padding-left: 0,
+    margin-left: 0,
+    display: inline,
+  )
+);
+
+$Toc__List--Unit: (
```
ajluk, line 238: Should it be $Toc__Unit__List here?

styles/designs/carnival/parts/_toc-components.scss
```diff
@@ -62,25 +62,25 @@ $TocChapterPage: (
...
   _properties: (),
   _components: (
-    $TocEocCompositeChapterList,
+    $Toc__Eoc__List--Chapter--Composite,
     tocLink('TocEocCompositeChapter'),
   )
 );
 
-$TocChapterList: (
+$Toc__List--Chapter: (
```
ajluk, line 83: Shoul it be `$Toc__Chapter__List` here?

styles/designs/carnival/parts/_toc-components.scss
```diff
@@ -62,25 +62,25 @@ $TocChapterPage: (
...
 );
 
-$TocEocCompositeChapter: (            //ToDo: apply empty_wrapper function
+$Toc__Eoc__Chapter--Composite: (            //ToDo: apply empty_wrapper function
   _name: 'TocEocCompositeChapter',
   _subselector: '> .os-toc-composite-chapter',
   _properties: (),
   _components: (
-    $TocEocCompositeChapterList,
+    $Toc__Eoc__List--Chapter--Composite,
```
ajluk, line 78: Same here.

styles/designs/carnival/parts/_toc-components.scss
```diff
@@ -62,25 +62,25 @@ $TocChapterPage: (
   )
 );
 
-$TocEocCompositeChapterList: (
+$Toc__Eoc__List--Chapter--Composite: (
```
ajluk, line 65: Should it be rather `$Toc__Eoc__Chapter--Composite__List`

styles/designs/carnival/parts/_titles-components.scss
```diff
@@ -130,7 +130,7 @@ $H3LinkComponent:(
     )
 );
 
-$H3ListComponent:(
+$H3--List:(
```
ajluk, line 133: Should these components in lines 121 and 133 be rather named to `$H3__Link`, `H3__List`?

styles/designs/carnival/parts/_indent-components.scss
Stackblocks: For all of the indent items i'll need to look into whether or not they (e.g. Containter, Paragraph, Section) all together form an `Indent`, or if each of them is an `Indent` used in a specific situation. That will determine which divider should be used.

styles/designs/carnival/parts/notes/_notes-components.scss
Stackblocks: Yup! I'll adjust it to correct the mistake.

styles/designs/carnival/parts/_toc-components.scss
Stackblocks: For the Toc__List items, I believe they should stay as is, because `Chapter`, `Unit`, and `Chapter (Composite)` are all the types of list.

styles/designs/carnival/parts/_titles-components.scss
Stackblocks: I'll need to double check that these are in fact "An H3 that is a list/link" instead of "A list/link that is part of H3s"

PR #2250, opened 2020-04-21
issue-2243

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2243

Recipe changes: 
- added additional selector for footnote and footnote call numbering 

The footnote work was previously in the g.g. branch but had to be removed until the transforms work is deployed. For the recipe change, we need both the old and new selectors until the transforms work is deployed. 

I also created a `_pages.scss` file to start putting our page specific styles and edited some footnote components, shapes and settings

Previous related PRS 
https://github.com/openstax/cnx-recipes/pull/2076
https://github.com/openstax/cnx-recipes/pull/2030
https://github.com/openstax/cnx-recipes/pull/2104

Issue #566, opened 2018-08-20
Footnotes appear in wrong sections


uskka:
Footnotes in webview appear in wrong section of the book. Footnotes 1,2,3 refer to the "Homework" content, in PDF these footnotes are in Homework section. 4 is in proper section but should have 1 number because it should appear as first footnote in this module, than 5 should be in "Chapter Review".

(1,2,3) https://katalyst01.cnx.org/contents/tWu56V64@34.15:PhCRzjQp@15/12-Data-Sampling-and-Variation

(4,5) https://katalyst01.cnx.org/contents/tWu56V64@34.15:3UZ5oaPG@6/14-Experimental-Design-and-Eth

Webview
![screenshot-katalyst01 cnx org-2018 08 20-18-58-23](https://user-images.githubusercontent.com/40228918/44354803-15fc0180-a4ab-11e8-83c3-53b5a5baca87.png)

Webview
![screenshot-katalyst01 cnx org-2018 08 20-18-45-05](https://user-images.githubusercontent.com/40228918/44354300-c0732500-a4a9-11e8-90b0-acb4015c859c.png)

PDF(Homework)
<img width="855" alt="screen shot 2018-08-20 at 18 45 46" src="https://user-images.githubusercontent.com/40228918/44354304-c406ac00-a4a9-11e8-96aa-71560629a856.png">

### QA Testing (added 5.20.20)

1.  run neb publish on each book with a trivial change in each module so that the version of each module changes.

example change: `</md:title>` to `</md:title>(add space)`

**Then we should:**

2. test that the previous version of each book (representative sample) still has old-style footnotes

3. test that the new version of each book has the new-style footnotes (representative sample)

**Potentially broken elements to look for:**

- disappearing footnotes (medium likelihood)
- pages that end abruptly (low likelihood)
- broken link between the footnote and the footnote marker in the content (low likelihood)

**Likely places a footnote will disappear:** 
inside a table header, a list, a table caption, a figure caption, a term definition, an exercise. (I'm not sure they occur in all of those places but I recall footnotes in table headers being a surprise)

wilkus27:
The same problem in Introductory Statistics:

1) https://katalyst01.cnx.org/contents/MBiUQmmY@20.1:CY4aJuYS@17/12-Data-Sampling-and-Variation -> **1,2,3** lead to **Homework**

[...]

PR #2318, opened 2020-05-21
remove the old footnote format

philschatz:
:broom: This completes the work in #2250 by removing the now-dead code.

This should be released once the book viewers all use the new footnote format.

refs #566


brittany-johnson:
@philschatz Hello! Thanks for creating this PR. When you say all book viewers, that includes Tutor as well?

Issue #2327, opened 2020-05-26
Prealgebra - style Preface


wilkus27:
Style Preface with specs from https://github.com/openstax/cnx-recipes/issues/2320

wilkus27:
:exclamation:  Blocked by https://github.com/openstax/content-managers/issues/1720

lwtchu:
@wilkus27 just published the change on katalyst01

MartaSztuka:
Tested
![Screenshot from 2020-06-09 11-33-11](https://user-images.githubusercontent.com/47241114/84131605-0dd8d180-aa45-11ea-81fa-96ecf4326462.png)
![Screenshot from 2020-06-09 11-32-29](https://user-images.githubusercontent.com/47241114/84131608-0f09fe80-aa45-11ea-9239-19be23c05476.png)

PR #2340, opened 2020-05-27
Style preface in prealgebra

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2327

Also I'm adding here style for basic lists.


styles/designs/corn/parts/_lists-shapes.scss
```diff
@@ -0,0 +1,40 @@
+@import 'lists-components';
+
+@include create_shape('corn:::ListsShape',(
```
wilkus27, line 3: I created general shape for lists in regular pages. It doesn't cover specific cases, like lists in exercises or notes. I think we can add new components or additional selectors to existing ones during work on those elements. Please, let me know what do you think. It's quite different approach then in carnival theme.

ajluk: I think it is a good idea. We can create different shapes for exercises and notes if it will be needed and check if we could overwrite the basic one that way.

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -0,0 +1,37 @@
...
+    )
+);
+
+$OrderedList--SecondLevel: (
+    _name: "SecondLevelOrderedList",
+    _subselector: " > li  > ol",
+    _properties:(
+        margin-left: h-spacing(2),
+    )
+);
```
wilkus27, line 37: Based on discussion with @ajluk , comment of @Stackblocks (https://github.com/openstax/cnx-recipes/pull/2385#discussion_r435297001) and convention from Carnival Theme (https://github.com/openstax/cnx-recipes/blob/9dd79dbd3a57999bd1207ff27430071a60d0e15c/styles/designs/carnival/parts/_lists-components.scss) I changed names of lists components. Let me know what do you think :slightly_smiling_face: 

I'm not sure about : 
```
$List__Container
$List__Wrapper--FirstLevel
$List__Wrapper--SecondLevel
$List__Wrapper--ThirdLevel
```
Maybe it should be
```
$ListContainer
$ListWrapper--FirstLevel
$ListWrapper--SecondLevel
$ListWrapper--ThirdLevel
```
or maybe it is ok, but we should add `List` part to other components, e.g
```
$List__UnorderedList--FirstLevel
``
Between `List` and `UnorderedList` is `__` not `--` because it's not modifier, but separate tag.

What is your opinion?

File: styles/books/dev-math/book.scss [hunk 3]
```diff
@@ -14,3 +14,5 @@ $ChapterIntroType: fullWidth;
 
 @include use('BookRoot', "common:::BookRoot");
 @include use('ModuleTitles', 'corn:::ModuleTitlesShape');
+@include use('Preface', 'corn:::PrefaceShape');
```
brittany-johnson, line 17: Hello! This is the first time I'm reviewing a PR from the corn theme. Can we add a comment at the top of this file with the specific books dev math includes? `Prealgebra 2e, Intermediate Algebra 2e & Elementary Algebra 2e`

styles/designs/corn/parts/_lists-components.scss
brittany-johnson: I think `$ListWrapper--FirstLevel` is better because in this instance, `List` and `Wrapper` are not two separate elements. 
I think the naming in this file is very clear, I like it!

wilkus27:
@ajluk @brittany-johnson Thank you for all your comments and suggestions! PR is ready for re-review :slightly_smiling_face:

Issue #2383, opened 2020-05-29
Prealgebra 2e - style Learning Objectives


wilkus27:
Style Learning Objectives with specs from https://github.com/openstax/cnx-recipes/issues/2190

wilkus27:
:exclamation: blocked by https://github.com/openstax/content-managers/issues/1726

lwtchu:
@wilkus27 i added the para tags on katalyst01

MartaSztuka:
Tested
![Screenshot from 2020-06-15 09-08-43](https://user-images.githubusercontent.com/47241114/84627835-e6bc4d00-aee7-11ea-893e-810c4a31ad37.png)
![Screenshot from 2020-06-15 09-04-12](https://user-images.githubusercontent.com/47241114/84627840-e7ed7a00-aee7-11ea-9cc4-62075594a5a6.png)

PR #2385, opened 2020-06-01
Style learning objectives in Prealgebra

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2383


styles/designs/corn/parts/_learning-components.scss
```diff
@@ -0,0 +1,37 @@
...
+    width: 0.5rem,
+    display: inline-block,
+    position: relative,
+    content: "''",
+    margin-right: h-spacing(1),
+    bottom: 0.1rem,
+  )
+);
+
+$Lists__FirstLevel__UnorderedList: (
```
ajluk, line 30: I'm wondering if the name should rather be `$Lists--FirstLevel__List--Unordered`?

wilkus27: Thank you for catching that! :slightly_smiling_face: I copied this component from PR https://github.com/openstax/cnx-recipes/pull/2340/files, but I forgot that later I totally changed properties for this list. So probably it should be named differently, maybe `$Lists--FirstLevel__List--Unordered--LeraningObjectives`? But it seems to be too long name :thinking:  What do you think about such convention in `lists-components.scss` file from PR 2340: 
- `$List__Wrapper--FirstLevel`
- `$List--Unordered--FirstLevel`
- `$List--Ordered--FirstLevel`
and then name for list from this PR it could be: `$List--Unordered--LearningObjectives`? 
Let me know your thoughts. @Stackblocks you have already approved this PR, but I'm courious also about your opinion. What do you all think? :relaxed:

Stackblocks: I agree that `$Lists--FirstLevel__List--Unordered--LeraningObjectives` is a bit long for a name 😆 

I like the alternatives you've proposed, so long as the `--Modifier` part is directly attached to the part that is modified. Keeping the modifiers clear makes sure new readers and people that are working with us can tell that `$List__Wrapper--FirstLevel` and `$List--FirstLevel__Wrapper` are different things.

wilkus27: @Stackblocks @ajluk Thank you for all your comments and suggestions! I changed names of components based on this and discussion from https://github.com/openstax/cnx-recipes/pull/2340#discussion_r435847735. :slightly_smiling_face:

brittweinstein:
@wilkus27 we can merge!

PR #2387, opened 2020-06-01
Pin lxml to 4.4.3 to avoid bug in versions >= 4.5.0

tomjw64:


requirements.txt
```diff
@@ -1,3 +1,4 @@
+lxml==4.4.3
```
m1yag1, line 1: I'm not sure this actually solves the problem. I did this and still ended up with the version conflict b/c when neb runs it does a `pip install -U -r requirements.txt` step which tries to update everything.

https://github.com/openstax/cnx-recipes/blob/master/script/assemble-book#L90

tomjw64: my understanding of `pip -U` is that it would force update to the version specified by the pin, i.e. if 4.5.1 is installed, use the pinned version 4.4.3 instead. In fact, I don't think this change would work without `pip -U`

m1yag1: yea, idk. that's what I had to do to get it to work. I can see it uninstalling the older version and upgrading it. Either way, it seems to work for Ripal so I think it's a problem on my end.

Issue #2386, opened 2020-06-01
Prealgebra 2e - style Be Prepared note


wilkus27:
Style `Be Prepared` with specs from https://github.com/openstax/cnx-recipes/issues/2179

MartaSztuka:
Tested
![Screenshot from 2020-06-04 12-40-28](https://user-images.githubusercontent.com/47241114/83747245-a1cf2580-a660-11ea-9fe7-b7cb376b5560.png)
![Screenshot from 2020-06-04 12-37-31](https://user-images.githubusercontent.com/47241114/83747247-a398e900-a660-11ea-9d91-473be67ca1b8.png)

PR #2388, opened 2020-06-02
Corn Be Prepared note

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2386


wilkus27:
### Changes in recipes

- In `config.scss` file for dev-math settings for **Be Prepared** note was applied two times. 
  Because of that, this note has two tags with `os-note-body` class.

Issue #2333, opened 2020-05-26
Learning Objective section numbers 


yblock:
# Scenario
The Learning Objective section number are doubling for each subsequent section. For example, what 1.2.1 is showing as 1.4.1, and 1.5.1 is showing as 1.10.1, etc. 

# Acceptance Criteria (required)
Section numbers should count up by one

# Steps to Reproduce (required)
* **How many times were you able to reproduce the issue?** 
     - all chapters/sections
* **Images or Screenshots Included:** [ x ] Yes  [ ] No  (_attach below_)
In this screenshot, the Learning Objective section numbers are showing 1.8.x where they should be 1.4.x, the section is being doubled
![Screen Shot 2020-05-26 at 10 23 18 AM](https://user-images.githubusercontent.com/5865292/82920408-d8a88b80-9f3c-11ea-8b32-1abf25c626b4.png)

# Environment: (required)
* **Medium**: [ x ] REX [ ] PDF  [ ] cnx.org
* **Instance**: [ ] production  [ ] qa  [ x ] staging  [ ] dev  [ ] content
* **OS**: [ x ] Mac OSX  [ ] Windows  [ ] iOS  [ ] Android
* **Browser**: [ ] Chrome  [ x ] Firefox  [ ] Safari  [ ] Opera  [ ] IE  [ ] Edge

# Impact
* **Severity**: [ x ] High [ ] Medium  [ ] Low
* **Frequency of Occurrence**: [ x ] High [ ] Medium  [ ] Low

# Next Steps for User
If the card is accepted or rejected, what does user need to know now?

# Test Plan

openstaxalina:
This is a showstopper - we cannot release without this fix.

I confirmed that this is NOT an issue in the live content: https://openstax.org/books/calculus-volume-1/pages/1-2-basic-classes-of-functions

This does seem to be introduced by recent code -- the bake before the Test Fest (march 17) did not have the issue: https://vendor.cnx.org/contents/i4nRcikn@12.6:056cQH8B@12/1-2-Basic-Classes-of-Functions

openstaxalina:
Here's a link to some content on a CNX instance (that does not redirect to REX) that shows the issue: https://vendor-staging.cnx.org/contents/i4nRcikn@13.2:fP4FUrwS@9/1-1-Review-of-Functions

Stackblocks:
Doubling is happening even in locally-baked content using current books.txt and looming.liberace branch

[...]

PR #2391, opened 2020-06-02
Issue 2333- LO numbering

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2333

- The release flag 'increment-section-counter-for-lo` was the source of the LO section numbering issues. The flag was counting the sections again. 
- I think its safe to remove this code and that it will not negatively affect other books because: 
  - The name of the flag clearly indicates what is was used for
  - Calculus is the only book with numbered learning objectives in sections in this way. 
  - The sections are being numbered in another location in the generator file


recipes/books/_generator.scss
```diff
@@ -462,9 +462,6 @@
   @include count_countChapters(chapter);
   @if $Config_SectionCompositePages != () { @include count_countSections(section); }
   @include count_countAppendices(appendix);
-  @if release-flag-enabled(increment-section-counter-for-lo) {
-    @include count_countSections(section);
-  };
```
wilkus27, line 467: It looks good. I was just wondering why this flag was created. I did some investigation and it was added for accounting first:
(https://github.com/openstax/cnx-recipes/pull/810/commits/7b01e593cf1383057f33c777b9c1eb9dcb29fb23)
image

I guess that in this book `count_countSections` include was needed in higher pass. But it's just my assumption. I think we should test that, before we merge those changes.

brittany-johnson: Thank you @wilkus27 for catching this! The LO section numbering is used in the EOC in accounting and was being negatively affected by removing the flag.
I've updated the PR. 

I didn't add the release flag back because I don't think other books need this. Instead, I added a config variable.

Issue #2396, opened 2020-06-03
Prealgebra 2e: Style Column Header Table


ajluk:
Issue with specs: #2204

MartaSztuka:
Tested
![Screenshot from 2020-06-08 12-47-28](https://user-images.githubusercontent.com/47241114/84022286-4bc2f080-a986-11ea-9274-a7b21107ec4d.png)
![Screenshot from 2020-06-08 12-49-20](https://user-images.githubusercontent.com/47241114/84022436-8af14180-a986-11ea-8575-291da9a693eb.png)

PR #2398, opened 2020-06-04
style column-header-table

ajluk:
Issue: #2396


styles/templates/template2/parts/_table-settings.scss
```diff
@@ -31,4 +33,45 @@
...
+            border-spacing: 0,
+        ),
+        TableData: (
+            font-family: (_ref: "typography:::baseFont"),
+            border-width: 0.09rem, 
+            border-style: solid,
+            border-color: (_ref: "colorMap:::basicTableCellBorderColor"),
+            color: (_ref: "colorMap:::tableColor"),
+            padding-right: 1rem,
+            padding-left: 1rem,
```
ajluk, line 57: I needed to decrease the paddings in `td` for this type of table to make the cells be more square and equal.

Issue #2395, opened 2020-06-03
Prealgebra 2e: style How To note


wilkus27:
Style note How To with specs from https://github.com/openstax/cnx-recipes/issues/2208

MartaSztuka:
Tested
![Screenshot from 2020-06-16 11-06-23](https://user-images.githubusercontent.com/47241114/84755012-6ff38280-afc1-11ea-8704-99f5efbd20cc.png)
![Screenshot from 2020-06-16 11-07-55](https://user-images.githubusercontent.com/47241114/84755212-a8935c00-afc1-11ea-85ab-7ed4e43a2611.png)

PR #2400, opened 2020-06-04
Style how to note in prealgebra

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2395


styles/designs/corn/parts/notes/_notes-iconBoxed-shapes.scss
```diff
@@ -9,7 +9,11 @@
                         $Notes--IconBoxed__Title__Icon,
                     )
                 )),
-                $Notes--IconBoxed__Body,
+                map-merge($Notes--IconBoxed__Body, (
+                    _components: (
+                        $Notes--IconBoxed__Subtitle,
+                    )
+                )),
```
wilkus27, line 16: I just added subtitle to existing IconBoxedShape. Please, let me know if you think that better solution is to create separate shapes for notes with and without subtitle.

ajluk: I think there is no need to create another shape just for one additional title, but let me know what you all think?

Stackblocks: I agree, I don't think there is a need to create another shape.

Issue #2404, opened 2020-06-05
Prealgebra 2e: Style Key Terms


ajluk:
Issue with specs: #2321

brittweinstein:
@ajluk please put points on this

MartaSztuka:
Tested
![Screenshot from 2020-06-30 15-58-48](https://user-images.githubusercontent.com/47241114/86135083-a6192200-baea-11ea-916d-a9bb56b60926.png)
![Screenshot from 2020-06-30 15-56-07](https://user-images.githubusercontent.com/47241114/86135093-a7e2e580-baea-11ea-9417-4cc8c035b275.png)

PR #2405, opened 2020-06-05
add styles to key terms prealgebra

ajluk:
Issue:  #2404


styles/designs/corn/parts/_titles-components.scss
```diff
@@ -86,6 +86,7 @@ $H2: (
         font-size: font-scale(3),
         line-height: line-height(3),
         margin-bottom: v-spacing(1),
+        text-transform: enum('ValueSet:::OPTIONAL')
```
wilkus27, line 89: I don't see it specified in the settings file, is this needed?

ajluk: Yes, I guess so. I don't know where was it came from. Thanks Natalia :wink:

Issue #2413, opened 2020-06-10
Prealgebra 2e: Style Appendix


ajluk:
Related issue with specs:  #2320

MartaSztuka:
Prealgebra 2e
![Screenshot from 2020-06-30 15-10-30](https://user-images.githubusercontent.com/47241114/86130137-08baef80-bae4-11ea-8ab8-f475fad4b723.png)
![Screenshot from 2020-06-30 15-10-22](https://user-images.githubusercontent.com/47241114/86130142-0a84b300-bae4-11ea-88ef-2ecce71262a8.png)
![Screenshot from 2020-06-30 15-10-11](https://user-images.githubusercontent.com/47241114/86130143-0b1d4980-bae4-11ea-9392-cb757b36c5c3.png)
![Screenshot from 2020-06-30 15-09-22](https://user-images.githubusercontent.com/47241114/86130148-0b1d4980-bae4-11ea-8ce0-e207be02c4eb.png)
![Screenshot from 2020-06-30 15-09-05](https://user-images.githubusercontent.com/47241114/86130150-0bb5e000-bae4-11ea-9362-1eba9458b419.png)
![Screenshot from 2020-06-30 15-08-29](https://user-images.githubusercontent.com/47241114/86130151-0c4e7680-bae4-11ea-9006-adbd95d8c9ba.png)

There's no Appendices College Algebra.

PR #2384, opened 2020-05-29
style chapter intro prealgebra

ajluk:
Issue: #2326

PR #2414, opened 2020-06-10
add styles to appendices prealgebra

ajluk:
Issue: #2413


styles/designs/corn/parts/_appendix-shapes.scss
```diff
@@ -0,0 +1,15 @@
+
```
wilkus27, line 1: Is it working without import `titles-components.scss` file? :thinking: Also I don't see `appendix-components.scss` file. I know it will contain only empty container wrapper but I think we should keep it for the consistent. On the other hand, maybe we should discuss if in this theme separate files for Preface and Appendix are necessary, because they don't have their own components. We style there only titles. Maybe we should create one big shape in `titles-shapes.scss` and there define titles from those two parts of book. Let me know what you all think :slightly_smiling_face: cc @Stackblocks @brittany-johnson

styles/designs/corn/parts/_titles-components.scss
```diff
@@ -8,10 +9,65 @@ $H1: (
...
+
+$Title__PartText: (
+  _name: "TitlePartText",
+  _subselector: '> .os-part-text',
+  _properties: (
+    display: none,
+  )
+);
+
+$Title__Number: (
```
wilkus27, line 28: I think we should keep the `H1` name here. What do you think about such convention:
```
$H1--Numbered__Container
$H1--Numbered__PartText
$H1--Numbered__Number
```
etc. Or maybe better will be `WithNumber` instead of `Numbered` :thinking: Let me know your thoughts.

Also I think we should keep in the selectors headers tags, so in `$Title__Container` component it should be `h1[data-type="document-title"]`

styles/designs/corn/parts/_titles-components.scss
```diff
@@ -8,10 +9,65 @@ $H1: (
...
+  _subselector: '> .os-number',
+  _properties: (
+    font-family: enum('ValueSet:::REQUIRED'),
+    font-size: enum('ValueSet:::REQUIRED'), 
+    line-height: enum('ValueSet:::REQUIRED'),
+    color: enum('ValueSet:::REQUIRED'),
+    border-style: solid,
+    border-width: enum('ValueSet:::REQUIRED'),
+    border-color: enum('ValueSet:::REQUIRED'),
+    font-weight: enum('ValueSet:::REQUIRED'),
```
wilkus27, line 39: What is the reason that properties like `font-size` etc. are REQUIRED and not hard coded? Are we using this component in a few places where different values are needed?

styles/templates/template2/parts/_appendix-settings.scss
```diff
@@ -0,0 +1,42 @@
+@import 'templates/template2/colors';
+@import 'templates/template2/typography';
```
wilkus27, line 2: We don't need those imports. Those files are imported in book.scss file.

styles/designs/corn/parts/_appendix-shapes.scss
```diff
@@ -0,0 +1,15 @@
...
+@include create_shape('corn:::AppendixShape', (
+    _components: (
+      map-merge($H1, (
+        _components: (
+          $Title__PartText,
+        //   $Title__Divider,
+          $Title__Text,
+          $Title__Number
```
wilkus27, line 9: In `titles-components.scss` there is created `Title__Container` component which is not added to this shape. In this case, is it necessary?

styles/designs/corn/parts/_titles-components.scss
ajluk: @wilkus They will be used in Chapter Intro Title.

styles/designs/corn/parts/_appendix-shapes.scss
ajluk: It will be used also in Chapter Intro title. I just keep those for me to decide about what values I should use, but I I have forgot to mention it. ;( But we can keep H1 for Chapter titile as well if we can accept a lot of required/optional values? I guess it will come up after merging all the shapes that use those titles.

styles/designs/corn/parts/_appendix-shapes.scss
ajluk: I haven't check if it is working without titles-compoents.scss file. I can check. I didn't use empty container, because I have notice if we use the settings file like in this case it is not needed, the selector comes from this file. For creating one big shape for titles in appendix and preface I'm not sure if it will be possible or confusing, because they are using different settings fir the same compoents, but maybe moving this shapes to titles-shapes.scss file can ba a good idea?

styles/designs/corn/parts/_titles-components.scss
ajluk: Yes, it is a good idea, it will be less confusing why the componantes are used and where.

styles/designs/corn/parts/_appendix-shapes.scss
ajluk: It is working without this import. I have added it so it is more clear where is they are come from.

ajluk:
@brittany-johnson, @Stackblocks I have added you tho this review, so you could tell what do you think about Natalia's suggestions about moving appendices and preface to titles shapes and removing shapes files for them.

styles/designs/corn/parts/_appendix-shapes.scss
wilkus27: I was thinking about adding additional components to the current shape from `titles-shapes.scss`. Something like that:
```
@include create_shape('corn:::TitlesShape', (
    _components: (
        map-merge($Preface__Container, (
            _components: (
                $H1
                $H2,
                $H3,
                $H4,
                $H5,
            )
        )),
        map-merge($Module__Container, (
            _components: (
                $H2,
                $H3,
                $H4,
                $H5,
            )
        )),
        map-merge($Appendix__Container, (
            _components: (
                $H1--Numbered__PartText
             etc.
            )
        )),
    )
));
```

brittany-johnson: This snippet looks good! 
And this would be in favor of having separate shape files for the preface, module, and appendix right?

styles/designs/corn/parts/_titles-components.scss
brittany-johnson: I like the re-naming suggestion. `$H1--WithNumber__Container` is good with me. I also like the idea of including the header tag with the selectors, which makes it even clearer.

ajluk:
@wilkus27 @brittany-johnson @Stackblocks I have created one big shape for Preface, Modules and Appendices titles. I have one concern about Chapter Titles. In this Chapter Intro PR I have created one shape for chapter and module titles(https://github.com/openstax/cnx-recipes/pull/2384/files): 
image. During the merging I will be not able to concatenate Chapter and Appendices components within the same shape because they use different settings and there is no possibility to nest components in the settings file or use the same component twice with different settings, it throws an error. I see two solutions: create a separate shape for Chapter title and use components used in Appendices or add these components to Chapter Intro shape. What solution is better in your opinion?

Issue #2416, opened 2020-06-12
Prealgebra 2e: Style Section Exercises


ajluk:
From #2059: "There is a section at the end of each module (3.4 Exercises) that requires an icon next to the header icon - structure is similar to (#2164) class="section-exercises""
Related issue with specs:  #2164

ajluk:
@lwtchu I have a question about the column layout in these exercises I have found place there where  some exercises are apearing in two columns. (actually found one but there can be more). Also the images there appearing differently than in the old pdf. I was trying to force three column layout behavior there, but without success. Is this look acceptable? (I'm asking here, because there is no separate issue with documentation for them).
![image](https://user-images.githubusercontent.com/55203598/84682428-e9459380-af35-11ea-88f4-eabdbc142bb3.png)
I have one more question: should these exercises start at new page?

lwtchu:
Talked via slack: but need to investigate if exercises can be displayed in order from left to right instead of up down then left to right. (screenshot below is from existing pdf)
<img width="751" alt="Screen Shot 2020-06-15 at 11 40 27 AM" src="https://user-images.githubusercontent.com/11299125/84694960-b4264a80-af0f-11ea-9199-658769680950.png">

ajluk:
Hi @lwtchu I have managed to style the 3 column exercises to display them fromm left to right!: 
![image](https://user-images.githubusercontent.com/55203598/84918486-f04cdd00-b0c0-11ea-9f57-34075d1451af.png)

But these exercises on the screenshot below, they were styled in 2 columns in the old pdf, so I tried to do this the same way as I did for three columns, but form some reason I can't force them to display form left to right in the pdf (just in html) also they weren't looking good in 3 columns. Is this acceptable just for these to stay this way?: 
![image](https://user-images.githubusercontent.com/55203598/84920253-34d97800-b0c3-11ea-9a15-16c7936abe59.png)

ajluk:
Hi! @lwtchu I have finally manage to style the two column exercises with help from @brittany-johnson! Here is how it looks now: 
![image](https://user-images.githubusercontent.com/55203598/85162348-7f90f680-b261-11ea-8bfd-0223bf752c49.png)
![image](https://user-images.githubusercontent.com/55203598/85162545-d8608f00-b261-11ea-87fc-c5846da371ce.png)

ajluk:
Specs added:

Main Title Text
display: inline-block,
margin-left: v-spacing(1),

TitleIcon
[...]

PR #2417, opened 2020-06-12
Section exercises prealgebra

ajluk:
Issue: #2416


ajluk:
I have tried to use flexbox to style two column exercises (Everyday Math, Writing Exercises) and I have stacked with something like this (the columns should be one by the other): 
image

ajluk:
Closed. The second solution added within this PR (https://github.com/openstax/cnx-recipes/pull/2434) will be used.

Issue #2433, opened 2020-06-18
Prealgebra 2e: style Key Concepts


wilkus27:
Style **Key Concepts** with specs from https://github.com/openstax/cnx-recipes/issues/2322

MartaSztuka:
Tested 
![Screenshot from 2020-06-26 14-32-50](https://user-images.githubusercontent.com/47241114/85857355-fd04bb80-b7b9-11ea-9554-63517bf99a58.png)
![Screenshot from 2020-06-26 14-32-33](https://user-images.githubusercontent.com/47241114/85857358-fd9d5200-b7b9-11ea-9691-304cc69db985.png)

PR #2435, opened 2020-06-19
Corn key concepts

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2433


wilkus27:
### Recipe changes

- create two composite chapters in EOC

Final structure:
```
Chapter Review
   Key Terms
   Key Concepts
Exercises
   Review Exercises
   Practice Test
```

styles/designs/corn/parts/_titles-shapes.scss
```diff
@@ -1,13 +1,26 @@
 @import 'titles-components';
 
-@include create_shape('corn:::ModuleTitlesShape', (
+@include create_shape('corn:::TitlesShape', (
     _components: (
-        map-merge($Module__Container, (
+        map-merge($Chapter__Container, (
```
ajluk, line 5: @wilkus27 I'm only concern about adding Chapter to this shape in connection with your comment here: https://github.com/openstax/cnx-recipes/pull/2414#discussion_r440011439. I have created general titles shape, but in this case we need to do a separate one for Appendices and Preface, because they use different selector `[data-type="page"]`. We already talked about this, but I wanted to link this PR's, when we decide about the solution. :)

wilkus27:
@ajluk I created separate shape for titles from composite chapters. It's the safest solution for now. After we make a decisions we will refactor code if needed.

PR #2437, opened 2020-06-22
Additional styles/recipe changes for AP Physics

brittany-johnson:
- added learning objectives to chapter outline
- added new selector to move the learning objectives to the chapter outline 
  - this is the "correct" version of the markup
  - in college physics, it uses the abstract data-type, this will be removed and updated in content at a later date
- add list styling to `FullWidthChapterIntroShape` 
- add page break rule learning objectives

The changes in this PR affect other books, but I did not put them behind a release flag because all of the changes are isolated to the chapter introduction section. I think we should test all of the books, chapter introduction section only. I've updated the testing checklist in the issue.


ajluk:
@brittany-johnson I have noticed that you changed the book and recipe name to `college-physics`, but this name wasn't changed in the release flags files I'm just bringing this up in case it has been missed.

brittweinstein:
@brittany-johnson can we merge this bad boy?

Issue #2450, opened 2020-06-25
College Algebra: style Example


wilkus27:
Style Example with specs from https://github.com/openstax/cnx-recipes/issues/2209

MartaSztuka:
Tested
![Screenshot from 2020-07-03 10-47-27](https://user-images.githubusercontent.com/47241114/86451060-a992f000-bd1a-11ea-886f-0273512c4bc4.png)
![Screenshot from 2020-07-03 10-47-03](https://user-images.githubusercontent.com/47241114/86451070-ab5cb380-bd1a-11ea-995c-84244ccb7453.png)

PR #2453, opened 2020-06-26
Style example in College Algebra

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2450


wilkus27:
### Recipe changes

- change markup of commentary title to be the same as for solution titile

Issue #2455, opened 2020-06-26
Prealgebra 2e: style Answer Key


ajluk:
Related issue with specs: #2460

ajluk:
Hi @lwtchu I have found out that images in in the Answer Key placed in the solution container has a new line element placed in paragraph before them, so they can't be styled exactly on top of the solution container as it was done in carnival theme: 
![image](https://user-images.githubusercontent.com/55203598/85869230-b3bd6780-b7cb-11ea-8a45-31f8081cc7fc.png)
In the pdf file they look like this: 
![image](https://user-images.githubusercontent.com/55203598/85869369-dd768e80-b7cb-11ea-8599-4a1ecb68e18c.png)
![image](https://user-images.githubusercontent.com/55203598/85869453-fb43f380-b7cb-11ea-8ef7-555e846b0750.png)
Is it acceptable or do you want to remove the lines from them?

ajluk:
Specs added: 
![image](https://user-images.githubusercontent.com/55203598/85870391-3abf0f80-b7cd-11ea-8701-95b35604b90c.png)
Solution Area
display: table;
margin-bottom: 1.4rem;
width: 100%;
Solution Container
![image](https://user-images.githubusercontent.com/55203598/85870526-6b9f4480-b7cd-11ea-822c-adfbbf3c5055.png)
display: inline-table;
overflow-wrap: break-word;
margin-bottom: 1.4rem;
vertical-align: top;
width: 33%;
padding-right: 0.7rem;
![image](https://user-images.githubusercontent.com/55203598/85870673-a4d7b480-b7cd-11ea-8b49-239a9ebc0d10.png)
Solution Number
display: table-cell;
text-decoration: none;
color: #000000;
font-weight: bold;
![image](https://user-images.githubusercontent.com/55203598/85870770-ccc71800-b7cd-11ea-964b-d1a6929b03ef.png)
Solution Container
    display: table-cell;
    width: 100%;
    padding-left: 0.7rem;

ajluk:
I've got one more question: do we want the lists with tokens also display inline if it's enough space?
![image](https://user-images.githubusercontent.com/55203598/85874777-7230ba80-b7d3-11ea-8f10-b184c1366a33.png)

lwtchu:
@ajluk yes please. if there's enough space, we should try to have them display inline. however, do we want to wait for the #2366 to be implemented before we proceed?

ajluk:
[...]

PR #2456, opened 2020-06-26
style answerkey exercises prealgebra

ajluk:
Issue: #2455


recipes/books/dev-math/_config.scss
```diff
@@ -176,6 +176,9 @@ $Config_UnnumberedElements: (
 
 $Config_FirstElements: (
   (selector: ' > [data-type="media"]:first-child'),
+  (selector: ' > .os-table:first-child'),
+  (selector: ' > ol:first-child > li:first-child > [data-type="media"]'),
```
ajluk, line 180: I have added `table` and  `[data-type="media"]` in ol lists to $Config_FirstElements, because there are appearing as first childern of the problem/solution container as well in some places in Prealgebra 2e.

styles/designs/common/pdf/_functions.scss
```diff
@@ -77,8 +77,9 @@
   $child: map-get($childSpecs, child);
   $name: map-get($childSpecs, name);
   $hasImageOnTop: map-get($childSpecs, hasImageOnTop);
-
-  @if $hasImageOnTop == true {
+  $hasTableOnTop:  map-get($childSpecs, hasTableOnTop);
+  
+  @if $hasImageOnTop == true or $hasTableOnTop == true {
```
ajluk, line 82: I have added this condition for components that have table on top of the problem/solution container.

styles/designs/common/pdf/_functions.scss
```diff
@@ -96,6 +98,8 @@
       _properties:(
           display: table-cell,
           width: 100%,
+          padding-left: enum('ValueSet:::OPTIONAL'), // used in Answer Key
+          vertical-align: enum('ValueSet:::OPTIONAL'), // used in Answer Key
```
wilkus27, line 102: I'm afraid it might causes problems. I created separate components for problems with and without image on top because of this property. When we add it to problems which starts with paragraph and this para contain math it displays wrong.

styles/designs/corn/parts/_answerkey-shapes.scss
```diff
@@ -0,0 +1,85 @@
...
+                    $UnorderedList--AnswerKey--Unstyled,
+                    map-merge($OrderedList--Exercises--Inline, (
+                      _components:(
+                          $OrderedList--Exercises__Item--Inline,
+                      ),
+                    )),
+                    $OrderedList--FirstLevel,
+                    map-merge($UnorderedList--AnswerKey--Inline, (
+                      _components:(
+                        $UnorderedList--AnswerKey__Item--Inline,
```
wilkus27, line 72: Based on our discusion about titles and lists I think this is a case where more optimal will be to create separate shapes for list in exercises/solutions in `list-shapes.scss` file.

styles/designs/corn/parts/_figureimage-components.scss
```diff
@@ -6,6 +6,14 @@ $Image: (
...
     )
 );
+// Used in AnswerKey
+// $Image--FullWidth: (
+//     _name: "Image",
+//     _subselector: " img",
+//     _properties:(
+//         width: 100%,
+//     )
+// );
```
wilkus27, line 16: Should it be removed or uncommented? :slightly_smiling_face:

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -39,13 +39,15 @@ $OrderedList--SecondLevel: (
 //Section Exercises List
 $OrderedList--Exercises--Inline: (
     _name: "ExercisesInlineOrderedList",
-    _subselector: " > ol",
+    _subselector: " > ol.circled",
```
wilkus27, line 42: Circled list appear also in other elements, not only in exercises. Can we make general components for token lists?

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -73,6 +75,37 @@ $OrderedList--Exercises__Item--Block: (
...
 );
 
+//used in AnswerKey
+$UnorderedList--AnswerKey--Unstyled: (
+    _name: "AnswerKeyUnorderedListUnstyled",
+    _subselector: " > ul",
+    _properties:(
+        list-style-type: none,
+    )
+);
```
wilkus27, line 85: image

Here is general component for unstyled lists. Can we use it instead of this one?

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -73,6 +75,37 @@ $OrderedList--Exercises__Item--Block: (
...
+
+$UnorderedList--AnswerKey--Inline: (
+    _name: "AnswerKeyUnorderedListInline",
+    _subselector: " > ul.circled",
+    _properties:(
+        list-style-type: none,
+        display: inline,
+        padding-right: .5rem
+    )
+);
```
wilkus27, line 95: As I said in one of the previous comment, I think we should create general components for token lists.

styles/designs/corn/parts/_figureimage-components.scss
ajluk: Off course! :)

styles/designs/common/pdf/_functions.scss
ajluk: @wilkus I have looked through entire Answer Key and it doesn't cause any problems.

ajluk:
The problem which I couldn't fix occurs only in two places: 
image
image

styles/designs/corn/parts/_answerkey-components.scss
```diff
@@ -0,0 +1,47 @@
...
+    margin-bottom: v-spacing(2),
+  ),
+);
+
+
+$Exercise__Solution__Wrapper: (
+  _name: "SolutionWrapper",
+  _subselector: ' [data-type="solution"]',
+  _properties: (
+    display: inline-table,
```
ajluk, line 36: When I tried display: inline-block as it was done in the section exercises the columns looks like this: 
image

styles/designs/carnival/parts/_answerkey-shapes.scss
```diff
@@ -26,7 +26,7 @@
                     $AnswerKey__UnorderedList--MarginBottomLeft,
                   ),
                 )),
-                map-merge($Exercise__Solution__Container--WithImageOnTop, (
+                map-merge($Exercise__Solution__Container--WithImageTableOnTop, (
```
ajluk, line 29: I have change the component's name to match insides tables on top as well (the same properties are needed there, so there is no need to create new components)

styles/designs/carnival/parts/_exercises-components.scss
```diff
@@ -21,8 +21,8 @@ $Exercise__Solution__Wrapper: (
   )
 );
 
-$Exercise__Solution__Container: exerciseChildContainer((child: solution, name: "SolutionContainer", hasImageOnTop: false));
-$Exercise__Solution__Container--WithImageOnTop: exerciseChildContainer((child: solution, name: "SolutionContainerWithImageOnTop", hasImageOnTop: true));
+$Exercise__Solution__Container: exerciseChildContainer((child: solution, name: "SolutionContainer", hasImageTableOnTop: false));
+$Exercise__Solution__Container--WithImageTableOnTop: exerciseChildContainer((child: solution, name: "SolutionContainerWithImageTableOnTop", hasImageTableOnTop: true));
```
ajluk, line 25: Same here in the exercises. There also tables on top appear.

styles/designs/common/pdf/_functions.scss
```diff
@@ -76,9 +76,9 @@
 @function exerciseChildContainer($childSpecs) {
   $child: map-get($childSpecs, child);
   $name: map-get($childSpecs, name);
-  $hasImageOnTop: map-get($childSpecs, hasImageOnTop);
-
-  @if $hasImageOnTop == true {
+  $hasImageTableOnTop: map-get($childSpecs, hasImageTableOnTop);
+  
+  @if $hasImageTableOnTop == true {
     @return (
```
ajluk, line 82: I have change the variable and condition names to match tables as well.

styles/designs/common/pdf/_functions.scss
ajluk: @wilkus27 I didn't create separate component for the Answer Key as we discussed, because regular solution containers cannot have both `vertical-align: bottom` and `padding-top: v-spacing(1)`.

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -1,8 +1,10 @@
 $ListContainer: empty_wrapper(Container, '');
 
-$ListWrapper--FirstLevel: empty_wrapper(FirstLevel, ' > section');
-$ListWrapper--SecondLevel: empty_wrapper(SecondLevel, ' > section > section');
-$ListWrapper--ThirdLevel: empty_wrapper(ThirdLevel, ' > section > section > section');
+$ListSectionWrapper--FirstLevel: empty_wrapper(SectionFirstLevel, ' > section');
+$ListSectionWrapper--SecondLevel: empty_wrapper(SectionSecondLevel, ' > section > section');
+$ListSectionWrapper--ThirdLevel: empty_wrapper(SectionThirdLevel, ' > section > section > section');
+$ListContainerWrapper--FirstLevel: empty_wrapper(ContainerFirstLevel, ' > div');
+$ListContainerWrapper--SecondLevel: empty_wrapper(ContainerSecondLevel, ' > div > div');
```
wilkus27, line 7: Similar change is already done. Branch should be updated with the newest changes.

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -36,40 +38,34 @@ $OrderedList--SecondLevel: (
...
-       display: inline-block,
+        padding-right: .5rem,
+        margin-left: enum('ValueSet:::OPTIONAL'), //used in Answer Key
     )
 );
 
-$OrderedList--Exercises--Block: (
-    _name: "ExercisesBlockOrderedList",
-    _subselector: " > ol",
+$UnorderedTokensList--Inline: (
```
wilkus27, line 53: I think there is no need to create two separate components for unordered and ordered lists with class `circled`. They always be styled the same.

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -82,10 +78,19 @@ $UnorderedList--LearningObjectives: (
...
     _name: "UnstyledUnorderedList",
     _subselector: " > ul",
     _properties:(
         list-style-type: none,
     )
 );
+
+$OrderedList--Unstyled: (
+    _name: "UnstyledOrderedList",
+    _subselector: " > ul",
```
wilkus27, line 92: It should be `ol`

styles/designs/corn/parts/_lists-shapes.scss
```diff
@@ -38,3 +38,58 @@
...
+                            )
+                        )),
+                    )
+                )),
+            )
+        )),
+    )
+));
+
+@include create_shape('corn:::SectionExercisesTokensListsShape',(
```
wilkus27, line 69: Those two shapes are identical. Maybe it would be better to create one shape for token lists.

styles/designs/corn/parts/_lists-components.scss
wilkus27: I noticed later that actually unordered list has one addition property. I added comment to code that we need to check if this property should be applied to ordered list as well.

styles/designs/corn/parts/_lists-components.scss
wilkus27: I noticed later that this component is not used anywhere so I removed it.

styles/designs/corn/parts/_lists-components.scss
```diff
@@ -40,53 +40,48 @@ $OrderedList--SecondLevel: (
...
-       padding-right: column-v-spacing(1),
-       display: inline-block,
     )
 );
 
-$OrderedList--Exercises--Block: (
-    _name: "ExercisesBlockOrderedList",
-    _subselector: " > ol",
+//Circled Lists
+$OrderedList--Circled: (
```
wilkus27, line 53: I changed names from `token` to `circled` because `token` is an element in stepwise lists so it  might be misleading if we use here `token` as a type of list.

styles/templates/template2/parts/_lists-settings.scss
```diff
@@ -3,6 +3,18 @@
         _selectors: ('[data-type="page"]'),
     ),
     CompositePageLists: (
-        _selectors: ('[data-type="composite-page"]'),
+        _selectors: ('.eoc[data-type="composite-page"]'),
```
wilkus27, line 6: I added class `eoc` to `CompositePageLists` to be sure that settings from this shape don't overwrite settings in composite pages from Answer Key

styles/designs/corn/parts/_lists-components.scss
wilkus27: After investigation I found out that this property was needed to overwrite settings for margin from `OrderedList--FirstLevel` component which was added in Answer Key shape. I removed it because there are no other lists then circled in solutions from Answer Key. And finally I create one component for lists with class `circled`.

wilkus27:
@Stackblocks @brittany-johnson I added some changes to Ania's code. I think it's ready for review. About problem which Ania mentioned here: https://github.com/openstax/cnx-recipes/pull/2456#issuecomment-653126693 I didn't find why it behaves in that way. I checked it in the browsers (Chrome and FireFox) and there it's working properly. Text from lists wrapped when width is decrising and there are always three columns. So I guess it is related with Prince :thinking: What do you think about creating separate issue for cases which Ania found?

brittany-johnson:
@wilkus27 Yes, I think that's a good idea to open an issue for the cases that @ajluk found since our deadline is approaching.

styles/designs/common/pdf/_functions.scss
brittany-johnson: @wilkus27 I want to make sure I understand this variable name. Do you know if `hasImageTableOnTop` means that and image OR a table can appear at the top? Or does this mean something else?

wilkus27: Yes, this is what Anna meant. She applied different settings for exercises which have image or table on top.

wilkus27:
### Recipes changes

- add class `first-element` to tables and media inside ol lists which are on the top of the solution container in Answer Key

Issue #2449, opened 2020-06-25
College Algebra: style Corequisite Skills


wilkus27:
Style corequisite skills with specs form https://github.com/openstax/cnx-recipes/issues/2222

MartaSztuka:
@wilkus27 Student Learning Objective isn't styled consistently, there are 3 types:

1. Student Learning Objective: with bullet points
![Screenshot from 2020-07-13 13-08-24.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0bce4107-f1af-49d9-8b0d-84706765efe1)

2. Student Learning Objectives with numbers
![Screenshot from 2020-07-13 13-08-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5954285f-d20e-471d-af4c-30c3f09f1c60)

3. Learning Objective: with yellow arrows
![Screenshot from 2020-07-13 13-04-23](https://user-images.githubusercontent.com/47241114/87297883-749c4f80-c509-11ea-84d6-6b8befe11fa2.png)

Is the difference in singular or plural form and missing colon intentional?

Elements within Corequisite Skills reset as supposed to.![Screenshot from 2020-07-13 13-24-26.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/eeb386d5-fe49-45bf-88e7-263c8c8f2736)![Screenshot from 2020-07-13 13-24-47.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3e6c92a2-da65-45ad-8e0c-9e68a68d4eab)![Screenshot from 2020-07-13 13-25-06.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/23fddc66-ab27-4b93-b157-71ef903bd46f)![Screenshot from 2020-07-13 13-25-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/b6556a50-2e38-472e-93b5-ffc5b73e0387)

wilkus27:
@lwtchu all problems which Marta mentioned are content related. Those learning objectives don't have class `learning-objectives`. What is more, used list should be unordered for the consistance with other Learning Objectives. Also titles are created in the content, not in recipes, so if we want them to be the same it should be changed in content as well.

lwtchu:
[...]

PR #2459, opened 2020-06-30
Coreq Skills

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2449


wilkus27:
### Recipe changes

- create `$Config_PartType_CoreqSkills` with keys: `sectionClassName`, `titleContent`, `counterName`, `exerciseTitleContent`, which are used for creating main Coreq Skills title and exercises from this section
- new mixin `count_countCoreqExercises` which counts exercises within coreq-skills section
- new mixin `number_numberCoreqExercises` which adds numbering to exercises
- new mixin `modify_tagCoreqExercises` which adds new class `os-coreq-exercises` to exercises from coreq-skills without class `unnumbered`
- new mixin `modify_CoreqSkillsTitle` which modifies main title of Coreq Skills section - create new title by recipe and move existing one below as a subtitle (Note: redundancy will be removed in the content)
- add to the existing mixin `number_numberContentExercises` exception `:not(.os-coreq-exercises)` - this changes output of each recipe but I didn't create release flag because class `os-coreq-exercises` is new, so it shouldn't have impact on other books

brittweinstein:
@wilkus27 I think we can merge!

PR #2464, opened 2020-07-01
supress coreq class in precalc and update readme

brittany-johnson:


brittweinstein:
@brittany-johnson we can merge!

Issue #2461, opened 2020-07-01
College Algebra Coreq: style notes


wilkus27:
Style notes in College Algebra Coreq:

- How To (https://github.com/openstax/cnx-recipes/issues/2208)
- Try It (https://github.com/openstax/cnx-recipes/issues/2179)
- Q&A (https://github.com/openstax/cnx-recipes/issues/2179)
- Media (https://github.com/openstax/cnx-recipes/issues/2206)
- No Class Note (https://github.com/openstax/cnx-recipes/issues/2191)

wilkus27:
@MartaSztuka Can you also test `Try It` , `Be Prepared` and `Media` note in Prealgebra 2e? I had to rework styling of icon condensed notes and I just want to be sure that they display properly in both books.

MartaSztuka:
**College Algebra Coreq**
@wilkus27 in How to note list is outside the box![Screenshot from 2020-07-09 12-18-57.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7b0b09f8-b48b-428f-95ca-964b79c76182)
Try it![Screenshot from 2020-07-09 12-24-41.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/155f49a1-e9ca-4117-b310-9d880704e2a5)
Q&A![Screenshot from 2020-07-09 12-26-52.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/07b01b80-9754-44e1-9456-6dd12e8e8a23)
Media![Screenshot from 2020-07-09 12-28-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ed969a37-834d-43be-86b2-c965ec9767e6)
No class note![Screenshot from 2020-07-09 12-32-01.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/71121ae0-2a4f-448e-8e50-5f3b602c18fa)

**Elementary Algebra**
Try it![Screenshot from 2020-07-09 12-33-49.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/d73a36e4-51da-4c53-a66d-65133fdab66f)
Be prepared![Screenshot from 2020-07-09 12-35-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fc7f6a40-60c9-4e0c-bf42-d7fc04642bd7)
[...]

PR #2472, opened 2020-07-02
Style notes in coreq book

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2461


styles/books/dev-math/book.scss
```diff
@@ -28,41 +28,39 @@ $bandColor: #093D4C;
...
   BePrepared: (
     _selectors: (".be-prepared"),
-    'Container:::display': flex,
-    'Container:::flex-direction': row,
-    'Title:::flex-shrink': 0,
-    'Title:::margin-right': h-spacing(3),
+    'Container:::display': table,
+    'Title:::display': table-cell,
+    'Title:::padding-right': h-spacing(3),
+    'Title:::white-space': nowrap,
```
wilkus27, line 34: I had to rework styling for Icon Condensed notes. In Precalc (coreq) trio there are complicated equations in such notes and there was a problem with displaying them when `display: flex` was used. I change it to `dispaly: table`. This is how it looks now:

image

brittany-johnson: I just noticed that the screenshot is for the "Try It" note. I'm assuming the complicated equations appear in the "Try It" and "Be Prepared" ?

wilkus27: Actually I didn't notice such cases in `Be Prepared`, but I decided to style all icon condensed notes with body on the side in one way just in case they contain equations in later chapters.

Issue #2468, opened 2020-07-01
Implement Footnotes in Corn Theme


brittweinstein:
Documentation: https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/2465

Books that Need This: 
- [x] Prealgebra 2e
- [x] Intermediate Algebra 2e
- [x] Elementary Algebra 2e
- [x] College Algebra with CoReq
- [x] Precalculus
- [x] Algebra & Trigonometry

brittany-johnson:
![Screen Shot 2020-07-07 at 7 09 59 PM](https://user-images.githubusercontent.com/24909007/86858121-a171e200-c085-11ea-9e99-8d189a6cb311.png)

MartaSztuka:
Prealgebra 2e![Screenshot from 2020-07-10 12-21-41.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/35bc58b9-6591-44ab-85a0-e11baba1e298)
Intermediate Algebra 2e![Screenshot from 2020-07-10 12-16-16.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/2bf175e5-6438-4ddf-8caa-391ec27e1d3f)
Elementary Algebra 2e![Screenshot from 2020-07-10 12-24-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/423e10f0-b42b-419f-b4b8-3bbefa1d09b1)
College Algebra with CoReq![Screenshot from 2020-07-10 13-10-43.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/9b7ff548-17e9-4db7-baca-c41c1331c1ce)
Precalculus![Screenshot from 2020-07-10 13-00-48.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ded56150-9970-413a-b732-f9dbc6fac295)
Algebra & Trigonometry![Screenshot from 2020-07-10 12-50-43.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/b6d09d0d-48d0-43dc-872a-c0a1c60ac655)

PR #2482, opened 2020-07-07
Footnotes for corn theme

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2468

- Footnotes will mostly be the same in all books, the only differences being fonts 
- because of this, I moved them to the common pdf folder


styles/output/precalculus-coreq-pdf.css
```diff
@@ -1,6 +1,8 @@
 @charset "UTF-8";
 @import url("https://fonts.googleapis.com/css2?family=Roboto+Slab:wght@400;700&display=swap&subset=latin,latin-ext");
 @import url("https://fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&display=swap&subset=latin,latin-ext");
+@import url("https://fonts.googleapis.com/css2?family=Roboto+Slab:wght@400;700&display=swap&subset=latin,latin-ext");
```
Stackblocks, line 4: Is there a reason that the font calls on lines 2 & 3 are getting imported again on 3&4? See also the `/* stylelint-disable-next-line meowtec/no-px */` comment on lines 604&605. (Sorry if i'm misreading this!)

styles/templates/template2/parts/_footnotes-settings.scss
```diff
@@ -0,0 +1,44 @@
+@import 'templates/template2/colors';
+@import 'templates/template2/typography';
```
wilkus27, line 2: I think we don't need those imports in `settings.scss` file. Those files are imported in `book.scss` files.

brittany-johnson: Ahh yes! I don't know why I put that there to be honest LOL! Thank you for catching that

PR #2490, opened 2020-07-09
Style lists in notes and exercises

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2474

PR #2492, opened 2020-07-10
tokens and tokens in lists

brittany-johnson:


wilkus27:
@brittany-johnson I have already made similar change within my work on lists in notes https://github.com/openstax/cnx-recipes/pull/2490. I noticed that actually even tokens appears in notes (and examples) they are inside `[data-type="problem"]`, so I created one shape for lists in exercises and I applied it for problems and solutions. This way tokens are styled in notes, examples, section exercises, eoc exercises and in answer key.

PR #2496, opened 2020-07-10
Fix wrapping of try it note

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2461

Fix problem noticed by Marta during testing:

image


brittweinstein:
@wilkus27 I think we can merge!

PR #2502, opened 2020-07-13
Coreq tokens list

brittany-johnson:


lwtchu:
@wilkus27 @brittany-johnson that area might be a content issue. we haven't gone through the PDF to check every instance  yet so we can probably do a content change there

Issue #2163, opened 2020-04-06
Create Polish Psychology Recipe


openstaxalina:
# Scenario

Natalia and/or Anna will create the Polish version of the Psychology recipe, to support publishing the Polish translation ~August 2020.

# Acceptance Criteria (required)

# Test Plan

wilkus27:
Informations for tester
-------
Known issues related with current content (https://katalyst01.cnx.org/contents/84JIUkA_@1.103:rs_ilJom@4/1-1-Wprowadzenie) not recipes.

### I. Missing elements

#### 1. Preface
#### 2. Learning Objectives

### II. Modules with wrong display

#### 1. Introduction page

All introduction pages are numbered for now:

![image](https://user-images.githubusercontent.com/40228854/88153990-78fdf200-cc06-11ea-92d7-f37012d1f965.png)

It is caused because they are missing `introduction` class.

#### 2. Indexes

For now, all terms are inside one module `Skorowidz rzeczowy`. When proper attributes be added to content, there will be three indexes: `Skorowidz rzeczowy`, `Skorowidz nazwisk`, `Skorowidz terminów obcojęzycznych`.

### III. Elements with wrong display

#### 1. Notes

Notes contain additional para with english name of note, and para with subtitle:

![image](https://user-images.githubusercontent.com/40228854/88155376-4e149d80-cc08-11ea-9cc3-748d6bbf7130.png)

First para will be removed, and the second one will be changed to div with `[data-type="title"]`.

#### 2. EOC exercises

Actually it's not content related, but I put this here for information. For now solutions to exercises from EOC are placed below them:

![image](https://user-images.githubusercontent.com/40228854/88156035-283bc880-cc09-11ea-8673-25fef0335d2e.png)

We haven't made a decision yet what to do with them. They will be moved to Answer Key or suppressed.

MartaSztuka:
## Tests results
### 1. Full stop is missing and there is random 'Psychologia' in italic.![Screenshot from 2020-08-26 14-51-40.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/1b7fd51c-35ef-4d2b-82ab-6aee5dbde3a2)
### 2. This Figure is removed in english version which affects further numbering![Screenshot from 2020-08-26 15-34-51.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/bf4e8bd4-26a8-42e8-9d81-32b999d303a7)
[...]

PR #2510, opened 2020-07-21
Polish psychology recipe

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2163


wilkus27:
### Recipe changes

- create new recipe for Polish Psychology: `pl-psychology`
   - create styleguides
   - create `config.scss` file
   - create `book.scss` file with imports of all needed files
- enable all release flags for `pl-psychology`
- move `$Config_FirstElements` from `pl-u-physics` config file to pl `common-stuff` file
- modify tables
   - create mixin `modify_tagTopCaptionedTables` to add class to tables with caption on the top
   - modify mixins `number_setCaptions` and `setCaptionsHelper`

recipes/books/pl-u-physics/_config.scss
```diff
@@ -159,12 +159,6 @@ $Config_UnnumberedElements: (
   (contextSelector: '.check-understanding', childSelector: '[data-type="exercise"]'),
 );
 
-$Config_FirstElements: (
-  (selector: ' > .os-figure:first-child'),
-  (selector: ' > div:first-child > p:first-child img:first-child'),
-  (selector: ' > [data-type="media"]:first-child'),
-);
-
```
wilkus27, line 167: This config is used in one of the release flags. I decided to move it to `common-stuff.scss` to not define it in each `config.scss` separately. I think it's better to have more then less.

recipes/mixins/_modify.scss
```diff
@@ -797,6 +797,12 @@
   }
 }
 
+@mixin modify_tagTopCaptionedTables() {
+  table {
+    class: "top-captioned";
+  }
+}
+
```
wilkus27, line 805: I created mixin which added class to tables which have caption on top. We don't want to support in Adaptarr `top-titled` tables, which contain fake title created in first row of `<thead>`:

image

In the future I think it would be better to add this class to the content, not create it by recipe. Especially when book contains a few different types of tables.

recipes/mixins/_number.scss
```diff
@@ -429,6 +444,15 @@
...
     }
+    @if ($hasTopCaption) {
+      &.top-captioned {
+        &::after {
+          class: "os-top-caption";
+          content: string(TopCaptionContent);
+          move-to: tableTopCaption;
+        }
+      }
+    }
```
wilkus27, line 455: I wasn't sure what class should have new div with title. For now I created the new one, but if you think it would be better to use the same class (`os-table-title`) as top-titled tables, please let me know and I will change it.

paulinaszczucinska:

recipes/mixins/_modify.scss
ajluk: @wilkus27 as we have discussed my only concern is that this class is added to every table in the polish Psychology book. We checked together that this particular book doesn't have any other types of tables. But maybe a better solution will be to add the class in the content instead to avoid future problems with other books as you suggested in the comment above. @brittany-johnson what do you think about this temporary solution Natalia has proposed? Can we accept it for now?

recipes/mixins/_number.scss
ajluk: Yes, I think different class is a better solution, but I'm not sure here either. @brittany-johnson I guess we need your advise here as well. :relaxed:

recipes/mixins/_modify.scss
brittany-johnson: @wilkus27 @ajluk 
I just realized that this would tag all tables in all books, not just Polish since this is in the `_modify` file. 
I saw that the class is only being utilized when the variable `$hasTopCaption` is used but I do have some concerns about tagging all tables. 
Have you tried building all recipes on this PR?

wilkus27: @brittany-johnson Yes, all recipes were built. There were no changes in other files. But I decided to remove this mixin. Class will be added in the content. :slightly_smiling_face:

brittany-johnson: Ahhhh, okay good to know. Thank you!

brittweinstein:
@wilkus27 we can merge!

Issue #2382, opened 2020-05-29
EoC Titles


MartaSztuka:
Titles `Pytania`, `Zadania` etc. are missing in EoC.
![Screenshot from 2020-05-29 12-46-59](https://user-images.githubusercontent.com/47241114/83251830-866ca200-a1aa-11ea-9bb6-8f55636e4b17.png)
![Screenshot from 2020-05-29 12-45-12](https://user-images.githubusercontent.com/47241114/83251837-87053880-a1aa-11ea-89ff-2d0971e1c24a.png)

brittweinstein:
On the 6/2 call, we decided we could address this when there is a small break in coding in the Corn theme and that the Katalyst team will address. I will put this in the appropriate milestone.

MartaSztuka:
# PASSED
## Titles in EoC (Kluczowe Pojęcia, Pytania, Zadania)
### Fizyka 1![Screenshot from 2020-08-13 15-59-50.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6a9a7840-857d-42fa-9147-e4aa9a963e78)
### Fizyka 2![Screenshot from 2020-08-13 14-53-46.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/053e1fc3-4f13-4516-ae10-33af5a7f9e72)![Screenshot from 2020-08-13 14-54-10.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/a8f6709a-ca84-44ae-8eb7-64c65ebd9e4e)
### Fizyka 3![Screenshot from 2020-08-13 14-55-22.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8a68ae57-489d-4fa3-95bd-cfd6fa9b310c)

PR #2525, opened 2020-07-22
Fix EOC titles in pl-u-physics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2382


styles/designs/carnival/parts/_eoc-shapes.scss
```diff
@@ -29,9 +29,21 @@
...
+        map-merge($Eoc__Container,(
+            _components:(
+                $H2,
+                map-merge($H3--WithBottomBorder,(
                     _components:(
-                        $H3--BottomBorder, // used only in pl-u-physics
+                        $H3--WithBottomBorder__Border,
                     ),
                 )),
                 $H4--Link,
```
wilkus27, line 49: I created separate shape for EOC titles with bottom border, where I'm using new component `$H3--WithBottomBorder` instead of regular `$H3`. I had to do it, because `$H3` has a few selectors and when sudo selector `:after` from child component is added then it stick to the last selector of parent component:

image

paulinaszczucinska:

Issue #2265, opened 2020-04-29
Symbols in Index


MartaSztuka:
Words that start with polish letter in Index shouldn't be under `Symbols` header at the beginning, but segregated by polish alphabet.
![Screenshot from 2020-04-29 16-43-29](https://user-images.githubusercontent.com/47241114/80609623-9b81d400-8a38-11ea-904e-f41ddeb35024.png)

wilkus27:
This problem was investigated here: https://github.com/openstax/cnx-recipes/issues/1042. It still appears. What is more, previously issue was resolved locally. It didn't work on webview. At this moment it doesn't work locally as well.

@brittany-johnson do you think it might be related with issues: https://app.zenhub.com/workspaces/content-engineering-tech-team-5af1f4cc12da5e6d74331b60/issues/openstax/cnx/924 and https://github.com/openstax/cnx/wiki/Character-encoding-bug-(Polish-books)?

Should we wait, and test it after those changes will be done?

brittany-johnson:
@ckline-tryptic Hello! Can you confirm if this issue is related to the work being done on these two issues: 
https://app.zenhub.com/workspaces/content-engineering-tech-team-5af1f4cc12da5e6d74331b60/issues/openstax/cnx/924 

https://github.com/openstax/cnx/wiki/Character-encoding-bug-(Polish-books)

Thank you!

MartaSztuka:
# PASSED
## Diacritics in Index
### Fizyka 1![Screenshot from 2020-08-05 17-42-46.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0bc9cf80-4163-499f-b4df-438c9ff542ec)![Screenshot from 2020-08-05 17-42-35.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/384ed70b-ab58-4b3b-9e96-9f02d4e23c3a)
### Fizyka 2![Screenshot from 2020-08-05 17-49-24.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/330ac9a9-d765-4cbd-8001-8e1ccc00d9a5)![Screenshot from 2020-08-05 17-49-11.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0f492a0a-3a27-497e-8592-004f4bebd0b6)
[...]

PR #2526, opened 2020-07-22
Polish characters in Index

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2265


paulinaszczucinska:

Issue #2555, opened 2020-07-29
Calculus vol1,2,3: Style Numbered Unnumbered Equations


ajluk:
Issues with specs: 
Unnumbered Equations: #2447
Numbered Equations:  #2210

ajluk:
@MartaSztuka >>> Information for testing purposes: In Calculus both numbered and unnumbered equations are present.

MartaSztuka:
# PASSED
## Unnumbered Equations
### Calculus 1![Screenshot from 2020-08-05 08-58-51.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/56d4bba4-a7de-4668-8cff-89da83f6b4f1)
### Calculus 2![Screenshot from 2020-08-05 08-59-22.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/46bf6ea4-2b82-474d-93ce-90e658038ebb)
### Calculus 3![Screenshot from 2020-08-05 09-03-15.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/422cd14f-9d35-4175-a75f-f2c30738d3e2)
## Numbered Equations
### Calculus 1![Screenshot from 2020-08-05 08-56-57.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/02259635-1363-476b-85a6-7ad74f78bf09)
### Calculus 2![Screenshot from 2020-08-05 08-59-48.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/eaeb7d23-6288-437b-81d1-66acb2ce9fec)
### Calculus 3![Screenshot from 2020-08-05 09-02-24.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3c9aff37-29e5-4693-9e35-72da29e7229e)

PR #2556, opened 2020-07-29
Style unnumbered numbered equations calculus

ajluk:
Issue: #2555


ajluk:
I have also added styles for unnumbered equations within this PR, because they can be found in the Calculus book as well.

brittweinstein:
@ajluk we can merge this!

Issue #2365, opened 2020-05-27
Implement condensed icon with no title note


brittweinstein:
Documentation: https://github.com/openstax/cnx-recipes/issues/2179
Books: 
- [x] Prealgebra 2e 
- [x] Int Alg 2e 
- [x] Elem Alg 2e 
- [x] College Alg w/ CoReq
- [x] Precalculus 
- [x] Alg and Trig
- [x] Calculus 
- [ ] Statistics

Answer collation styling: 
- [ ] Prealgebra 2e

brittweinstein:
We still need a PR for the Precalc, A&T, CA titles.

brittany-johnson:
![Screen Shot 2020-07-02 at 11 49 34 AM](https://user-images.githubusercontent.com/24909007/86388383-55c0d200-bc5a-11ea-82f7-b46e799aeed7.png)
![Screen Shot 2020-07-02 at 11 49 24 AM](https://user-images.githubusercontent.com/24909007/86388427-67a27500-bc5a-11ea-9133-97de7ed61647.png)

wilkus27:
@MartaSztuka It can be tested now in **Precalculus**, **College Algebra** and **Algebra & Trigonometry**

MartaSztuka:
Algebra & Trigonometry![Screenshot from 2020-07-28 12-53-12.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4eddda00-830a-4af9-9174-78cebc740a8b)
College Algebra![Screenshot from 2020-07-28 12-55-54.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c2840073-b3b0-4a4d-a492-b10cb935ed8c)
Precalculus![Screenshot from 2020-07-28 12-59-14.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c447c48a-1f35-4b92-9e5b-e8ede9b5f60e)

MartaSztuka:
Not tested on Calculus and Statistics yet.

wilkus27:
Can be tested in Calculus books. For Statistic will be created separate PR.

MartaSztuka:
# FAILED
## Condensed icon
@wilkus27 Media note is not styled
[...]

PR #2558, opened 2020-07-30
Style Checkpoint note in Calculus

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2365


styles/designs/corn/parts/notes/_notes-components.scss
```diff
@@ -36,6 +36,14 @@ $Notes--IconCondensed__Title__Label: (
...
 );
 
+$Notes--IconCondensed__Title__Label--Hidden: (
+    _name: "TitleLabelHidden",
+    _subselector: ' > .os-title-label',
+    _properties: (
+        display: none,
+    )
+);
+
```
wilkus27, line 46: I hid label in styles, not removed it in recipes because it's used in Answer Key.

styles/designs/corn/parts/notes/_notes-iconCondensed-shapes.scss
```diff
@@ -18,6 +18,23 @@
...
+                        $Notes--IconCondensed__Title__Label--Hidden,
+                        $Notes--IconCondensed__Title__Number,
+                    )
+                )),
+                $Notes--IconCondensed__Body,
+            )
+        )),
+    )
+));
+
```
wilkus27, line 37: This shape differs from original one with just one component. I wasn't sure what would be better, to create separate shape or to change all properties from `$Notes--IconCondensed__Title__Label` to OPTIONAL. I choose the first approach because I think it has less repetition. But I'm courious about your thoughts. Let me know what do you think :slightly_smiling_face:

ajluk: Yes, I think it's more clear then.

brittweinstein:
@wilkus27 we can merge this!

Issue #2559, opened 2020-07-30
Calculus: Style Theorem Note


ajluk:
Issue with specs: #2256

MartaSztuka:
# PASSED
## Theorem Notes
### Calculus 1![Screenshot from 2020-08-05 08-47-12.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4c9f0643-bfb2-49da-bb5e-e534fc7ad480)
### Calculus 2![Screenshot from 2020-08-05 08-51-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3386a470-8290-46fb-ae9c-1b1f2a4940b2)
### Calculus 3![Screenshot from 2020-08-05 08-52-58.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/358c1091-d3d1-4b18-bb81-d83e7aa97d59)

PR #2560, opened 2020-07-30
style theorem note

ajluk:
Issue: #2559


styles/designs/corn/parts/notes/_notes-components.scss
```diff
@@ -172,7 +172,7 @@ $Notes--Boxed__Container: (
 
 $Notes--Boxed__Title: (
     _name: "Title",
-    _subselector: ' > h3.os-title',
```
ajluk, line 175: I have removed `h3` from the subselector here to make this components match the markup title in Theorem note:
image
Let me know if should rather create a new component instead.

brittweinstein:
@ajluk we can merge this!

Issue #2561, opened 2020-07-31
Calculus 1,2,3: Style Problem-Solving Strategy note


ajluk:
Issue with specs:
 #2256

MartaSztuka:
# PASSED
## Problem-Solving Strategy
### Calculus 1![Screenshot from 2020-08-07 15-55-23.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0a706987-e3cc-48f4-9ad0-4a9c7acdf5e9)
### Calculus 2![Screenshot from 2020-08-07 15-54-08.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c57a4ce9-bdd9-4156-928f-412f864367cd)
### Calculus 3![Screenshot from 2020-08-07 15-54-53.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/2caacd95-c02d-4a29-9251-b9e07c8db4f9)

PR #2562, opened 2020-07-31
style problem solving note calculus

ajluk:
Issue:  #2561


recipes/books/calculus/_config.scss
```diff
@@ -119,6 +119,7 @@ $Config_Notes: (
   (className: 'checkpoint', moveSolutionTo: $AREA_EOB, counterName: checkpoint, titleContent: $_checkpointTitleContent, hasSolutions: true, solutionTitleContent: $_checkpointSolutionTitleContent),
   (className: 'theorem', counterName: theorem, titleContent: $_theoremTitleContent),
   (className: 'project', labelText: "Student Project"),
+  (className: 'problem-solving', labelText: "Problem-Solving Strategy"),
```
ajluk, line 122: I have added the title label for this note to split the main title from the subtitle to standardize it with the other similar notes, because they appeared as one title in the content. 
image
After these changes the note looks like this and I have checked it with Larissa.The `Problem-Solving Strategy` part of the subtitle will be removed in the content:
image

styles/templates/template2/_colors.scss
```diff
@@ -4,7 +4,7 @@
     B1:    #093D4C,
     B2:    #1E7E91,
     B3:    #34BDD8,
-    B4:    #D2E5E9,
```
ajluk, line 7: I have changed this color value in the scheme based on the information from Larissa.

Issue #2511, opened 2020-07-21
corn theme: stepwise lists need to be supported


lwtchu:
- [x] College Algebra w/ Coreq
- [x] Precalculus
- [x] Alg and Trig

![Screen Shot 2020-07-21 at 10 10 56 AM](https://user-images.githubusercontent.com/11299125/88072639-e7757e00-cb3a-11ea-949f-d1d010e7ad98.png)
Screenshot from Ch 4 Linear Functions / Sec 2 Linear Functions
These have the class `stepwise` for the list and it would be nice if they looked like the how to ones with the words "Step 1. Step 2. etc"

What it looks like now:
![Screen Shot 2020-07-21 at 10 16 35 AM](https://user-images.githubusercontent.com/11299125/88073164-800bfe00-cb3b-11ea-9e52-591edbdc4df5.png)
Screenshot from Ch 4 Linear Functions / Sec 2 Linear Functions

What it should look like:
![Screen Shot 2020-07-21 at 10 18 08 AM](https://user-images.githubusercontent.com/11299125/88073186-8601df00-cb3b-11ea-8e80-b588455a1394.png)

MartaSztuka:
@wilkus27 in 'How to' is missing, also Step lists aren't supported.
(Example from Precalculus) ![Screenshot from 2020-07-27 15-53-25.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/004ccffd-cc2d-47e9-9aaa-61de9930dcda)

College Algebra coreq! Here it is ok. [Screenshot from 2020-07-27 15-45-44.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4d994754-1af8-4a1c-abbf-30fe6c50c446)
Algebra&Trigonometry![Screenshot from 2020-07-27 15-47-12.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5cdf5b44-6dd4-4f14-a03d-ffdd756e22bd)
College Algebra![Screenshot from 2020-07-27 15-49-25.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/1bd46f7e-c633-4b65-8b90-7a833cce37db)

MartaSztuka:
# FAILED
## Stepwise lists
[...]

Issue #2474, opened 2020-07-02
Implement Lists Within Notes for Dev Math and Co Req


brittweinstein:
Documentation: https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/2207

- [x] Prealgebra 2e 
- [x] Int Alg 2e 
- [x] Elem Alg 2e 
- [x] College Alg w/ CoReq
- [x] Precalculus 
- [x] Alg and Trig

brittany-johnson:
Link to a screenshot of where this issue occurs: https://github.com/openstax/cnx-recipes/issues/2368#issuecomment-655827026

brittany-johnson:
![Screen Shot 2020-07-10 at 2 55 32 PM](https://user-images.githubusercontent.com/24909007/87193048-740f7900-c2bd-11ea-8718-6947ad5db81f.png)

MartaSztuka:
@brittany-johnson in Precalculus and  Alg and Trig all notes are unstyled
Precalculus ![Screenshot from 2020-07-13 14-09-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ea91ae3e-61b3-4f1f-9c2b-749f17903470)
Alegbra&Trigonometry![Screenshot from 2020-07-13 14-11-01.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ad837fb0-0d35-4792-8aae-3762b8122c69)
 Prealgebra 2e![Screenshot from 2020-07-13 14-05-52.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/81001015-da58-4389-ab9e-5f3997a1abb4)
 Int Alg 2e![Screenshot from 2020-07-13 14-06-31.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7d7f194f-1937-4ec1-8ac4-658179d658f0)
 Elem Alg 2e![Screenshot from 2020-07-13 14-07-33.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/25689277-1f5d-401c-b8e1-da4b593d2e5e)
 College Alg CoReq![Screenshot from 2020-07-13 14-07-53.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/eb180f57-7112-44a5-bf81-d5e93096dbf0)

[...]

PR #2565, opened 2020-07-31
Style special cases of lists

wilkus27:
Issues: 
- https://github.com/openstax/cnx-recipes/issues/2474
- https://github.com/openstax/cnx-recipes/issues/2511

and generally I added also styles to special cases of lists which I noticed that they are still unstyled.


brittweinstein:
@wilkus27 we can merge this!

Issue #2585, opened 2020-08-05
Calculus vol1, 2, 3: Style Answer Key


ajluk:
Related issue with specs: #2460

MartaSztuka:
# FAILED
## Answer Key
@lwtchu Answer Key should be in three column display, but two column appears randomly. Same issue as in this one: https://github.com/openstax/cnx-recipes/issues/2486 
### Calculus 1https://github.com/openstax/cnx-recipes/issues/2486
### Calculus 2![Screenshot from 2020-08-07 15-51-11.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/07c6015e-fcc4-46c3-8322-ccbd3db4d14f)
### Calculus 3![Screenshot from 2020-08-07 15-30-40.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/76bc9059-5e5c-4323-afe3-0eae1fe73673)
cc @ajluk

lwtchu:
@MartaSztuka i think in terms of answer key testing, we can close. and keep the spontaneous two column issue to #2486

PR #2586, opened 2020-08-05
Answer key calculus

ajluk:
Issues:
#2585
#2587


styles/templates/template2/parts/_answerkey-settings.scss
```diff
@@ -2,7 +2,7 @@
     AnswerKey: (
         //There is a .os-solutions-container and a .os-solution-container
         //This should be cleaned up so that we are using one not both in the recipe
-        _selectors: ('[data-type="composite-chapter"].os-eob.os-solutions-container, [data-type="composite-chapter"].os-eob.os-solution-container'),
+        _selectors: ('[data-type="composite-chapter"].os-eob.os-solutions-container'),
         H1Title: (
             color: (_ref: 'colorMap:::pageTitleColor'),
             font-family: (_ref: 'typography:::titleFont'),
```
ajluk, line 8: After investigate I have found that additional sectors has been added in the `answer-key-setting.scss` file for both shapes `AnswerKey` and `AnswerKeyNestedSolutionAreas`:
image
image which caused a regression in books using AnswerKey shape (f.ex Prealgebra 2e). The cause of this issue is that the right selector for these books should be `'[data-type="composite-chapter"].os-eob.os-solutions-container'` which was added as the first one after the changes. Then the child component `h1` has been given only the second selector as a parent in the output : `'[data-type="composite-chapter"].os-eob.os-solution-container' > h1`. The letter s is the only difference in the name of the component: should be solutions not solution. I have removed the additional selector and checked the books

styles/templates/template2/parts/_answerkey-settings.scss
```diff
@@ -39,7 +39,7 @@
     AnswerKeyNestedSolutionAreas: (
         //There is a .os-solutions-container and a .os-solution-container
         //This should be cleaned up so that we are using one not both in the recipe
-        _selectors: ('[data-type="composite-chapter"].os-eob.os-solutions-container, [data-type="composite-chapter"].os-eob.os-solution-container'),
+        _selectors: ('[data-type="composite-chapter"].os-eob.os-solution-container'),
```
ajluk, line 42: See my comment above. I have removed the additional selector for this shape as well. Here the right selector should be the second one: `'[data-type="composite-chapter"].os-eob.os-solution-container'` (that's why there was no rgression). I have also checked College Algebra Coreq and nothing changed.

Issue #2591, opened 2020-08-06
Statistics: Style Tables and Figures


ajluk:
Related issue with specs: 
Tables:  #2204
Figures and Captions: #2185

MartaSztuka:
# PASSED
## Tables
### Business Statistics![Screenshot from 2020-08-11 13-37-05.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/1d8c14c5-6467-4d7e-9d59-5cfd6a061e97)
### HS Statistics![Screenshot from 2020-08-11 13-36-06.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/018ceba1-3049-4acf-99c8-58c42263aa58)
### Introductory Statistics![Screenshot from 2020-08-11 13-33-56.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4228a7c9-741c-4995-b649-d45d419ce8c9)

## Figures
### Business Statistics![Screenshot from 2020-08-11 13-37-27.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/41807225-f5cb-4cd9-b7ec-dd61aa6ba6b2)
### HS Statistics![Screenshot from 2020-08-11 13-35-30.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8bf5b89d-f9d8-4de5-a948-09400b0efbfd)
### Introductory Statistics![Screenshot from 2020-08-11 13-39-01.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3215b2b3-8443-428b-8871-2944afe7d100)

PR #2593, opened 2020-08-06
style figures tables in statistics

ajluk:
issue: #2591


brittweinstein:
@ajluk  we can merge!

Issue #2588, opened 2020-08-06
Implement Introduction to Statistics


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/2314

wilkus27:
@MartaSztuka Here are more informations https://github.com/openstax/cnx-recipes/issues/2314#issuecomment-669942177

MartaSztuka:
# PASSED
## Introductions
### HS Statistics![Screenshot from 2020-08-12 10-17-32.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7123927a-e7dc-46eb-b1a8-9add043800d2)
### Intro Statistics![Screenshot from 2020-08-12 10-18-47.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0da0fef4-8da8-4e63-a109-189bd49429b3)
### Business Statistics![Screenshot from 2020-08-12 10-20-45.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/d16e8402-c7c1-4908-82fd-898470b935fb)

PR #2594, opened 2020-08-06
Style introduction page in Statistics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2588


styles/designs/corn/parts/_chapterintro-components.scss
```diff
@@ -1,13 +1,6 @@
 $ChapterIntro__Container: empty_wrapper(ChapterIntroContainer, ' > .intro-body');
 
-$Introduction__Container: (
-    _name: "IntroductionContainer",
-    _subselector: '',
-    _properties: (
-        position: relative,
-        top: -6rem,
-    )
-);
```
wilkus27, line 10: These properties was added because there was blank space between splash image and folio band. Unfortunatelly it caused such problem when introduction text jumped to the next page:

image

That's why I removed this code. I found out that blank space occured because chapter title had property `position: relative`, so it still took up space where it originally was. Instead I added `position: relative` to chapter and applied `position: absolute` to chapter title.

styles/designs/corn/parts/_chapterintro-shapes.scss
```diff
@@ -64,3 +64,49 @@
     
     )
 ));
+
+@include create_shape('corn:::ChapterIntroWithoutChapterOutlineShape', (
```
wilkus27, line 68: I created new shape because Statistics doesn't have Chapter Outline on Introduction page. Instead it has Chapter Objectives which should be styled exactly as Learning Objectives. That's why I created separate shape for this element in `learning-shapes.scss` file. What do you think about moving also Chapter Outline part to this file? In corn theme chapter outline is independent on introduction page, not like it was in carnival theme. Also Chapter Outline and Learning Objectives are styled similarly here, so some of the components could be reused. If we do this, then there will be just one shape for Introduction page. Let me know your thoughts :wink:

ajluk: I think is much cleaner this way to keep separate these shapes, but second proposition is also acceptable. So I guess it's up to you Natalia.

Issue #2377, opened 2020-05-27
Implement stats lab feature (full page section note feature


brittweinstein:
Documentation: https://app.zenhub.com/workspaces/textbook-developers-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/2273

Books: 

- [x] Statistics

ajluk:
The bleeds in the top bar and bottom line has been checked by Larissa:
![image](https://user-images.githubusercontent.com/55203598/89825696-47b57980-db55-11ea-9928-81176f364a30.png)
![image](https://user-images.githubusercontent.com/55203598/89825730-5439d200-db55-11ea-8d4d-e3da7a1f7848.png)

MartaSztuka:
# PASSED
## Stats lab
### Intro Statistics![Screenshot from 2020-08-12 16-26-03.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/16f5cc7a-ba9b-4996-a3e1-932c664c0edc)

ajluk:
I have reopened this issue, because this note is also present in HS Statistics and needs to be tested there as well.

MartaSztuka:
# PASSED
## Stats Lab
### HS Statistics
![Screenshot from 2020-08-12 16-48-12](https://user-images.githubusercontent.com/47241114/90029864-a432a900-dcbb-11ea-9495-7873e9f00efe.png)

PR #2599, opened 2020-08-10
style Stats Lab feature statistics

ajluk:
Issue:  #2377


recipes/books/statistics/_config.scss
```diff
@@ -104,7 +104,7 @@ $Config_BookCompositePages: (
 
 $_exampleTitleContent        : (os-title-label: "Example ", os-number: counter(chapter) "." counter(example), os-divider: " ");
 $_exampleSolutionTitleContent: (os-title-label: "Solution ", os-number: counter(chapter) "." counter(example), os-divider: " ");
-$_labTitleContent           : (os-title-label: "Stats Lab ", os-number: counter(chapter) "." counter(lab));
+$_labTitleContent           : (os-title-label: "Stats Lab ");
```
ajluk, line 107: I have removed the numbering suggested by Larissa.

styles/designs/corn/parts/notes/_notes-iconFullPage-shapes.scss
```diff
@@ -0,0 +1,32 @@
+@import 'notes-components';
+@import '../lists-components';
+
+
+@include create_shape('corn:::IconFullPageNoteShape', (
```
ajluk, line 5: I have created new shape for this note, because in this case I had to add/change a lot of properties values. The design is different than in other Icon Condensed notes. So I thought it will be cleaner that way. But let me know if should rather use existing shape and make more properties REQUIRED/OPTIONAL.

styles/designs/corn/parts/notes/_notes-iconFullPage-shapes.scss
```diff
@@ -0,0 +1,44 @@
...
+    _components: (
+        map-merge($Notes--IconFullPage__Container, (
+            _components: (
+                map-merge($Notes--IconFullPage__Title, (
+                    _components: (
+                        $Notes--IconFullPage__Title__Icon,
+                        $Notes--IconFullPage__Title__Label,
+                    )
+                )),
+                map-merge($Notes--IconFullPage__Body, (
```
ajluk, line 15: There are lists nested in the note body in this note. Let me know if I should rather create a separate shape for them. It will be used only in this note, so wasn't sure if this is a good idea.

styles/designs/corn/parts/notes/_notes-iconFullPage-shapes.scss
wilkus27: I think it's a good idea :relaxed: It's more clear now.

styles/designs/corn/parts/notes/_notes-iconFullPage-shapes.scss
wilkus27: Hmm there is a lot of lists components here, what makes note shape complex. In this case I think it will be better to create separate shape for lists in notes, but maybe not in this PR. Maybe we can merge this for now as is and create separate issue "Rework lists in notes" and take care of this later. What do you think? :slightly_smiling_face:

Issue #2606, opened 2020-08-12
Statistics: style Key Terms


ajluk:
Issue with specs: #2321

ajluk:
Link to comment about content issues in Statistics books in Key Terms: https://github.com/openstax/cnx-recipes/issues/2321#issuecomment-672804331

MartaSztuka:
# PASSED
## Key Terms
### HS Statistics![Screenshot from 2020-08-18 13-48-22.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ec4db147-1832-4287-8cfe-a3611fbbf5f3)
### Intro Statistics![Screenshot from 2020-08-18 13-50-24.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f6b5f1a6-dedb-491b-aaaf-3f371734b245)
### Business Statistics![Screenshot from 2020-08-18 13-50-59.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/17129d1c-28fd-4ab3-832c-8bc0ec1b08fb)

PR #2607, opened 2020-08-12
style key terms statistics

ajluk:
Issue:  #2606


styles/books/statistics/book.scss
```diff
@@ -92,6 +92,12 @@ $bandColor: #093D4C;
   ),
 ));
 
+@include add_settings((
+    KeyTerm: (
+    _selectors: ('[data-type = "chapter"] > [data-type="composite-page"].os-glossary-container')
+  )
+));
+
```
wilkus27, line 100: Is it necessery? Maybe we could change selector in `eoc-settings.scss` file just to `os-glossary-conatiner`. Then it will be common for all Key Terms, those in composite chapters and composite pages.

wilkus27: Also selector on the top of this file for eoc bands should be changed:

image

In Statistics there is no composite chapter. EOC starts with Key Terms.

ajluk: @wilkus27 I haven't put attention to these band selectors, because I haven't changed them within this PR, but thanks for noticing it!

Issue #2609, opened 2020-08-13
Statistics: Style Chapter Review


ajluk:

ajluk:
Hi @lwtchu! Here how the Chpter Review and Formula Review look now in Statistics (only titles and lists has been styled):

**Chapter Review:**

![image](https://user-images.githubusercontent.com/55203598/90122449-933a7400-dd5d-11ea-954f-b5b5fb03f1ab.png)
![image](https://user-images.githubusercontent.com/55203598/90122490-a51c1700-dd5d-11ea-9437-717e9b0e8f1a.png)

**Formula Review:**
![image](https://user-images.githubusercontent.com/55203598/90122591-cf6dd480-dd5d-11ea-9d27-71b2836442c5.png)
![image](https://user-images.githubusercontent.com/55203598/90122651-e4e2fe80-dd5d-11ea-9a20-71cca7d7038d.png)

Do you want them to stay in one column as they are now?
Also could I ask you to open a separate documentation cards for them? I will re paste the screenshots and put specs there?

lwtchu:
For chapter review can we follow the same specs as the key concepts? #2322 where in this case, in addition to lists there's paragraphs? 

Here's the new formula review card #2614

lwtchu:
Here's the new Chapter Review documentatio card #2615

MartaSztuka:
# PASSED
## Chapter Review
### HS Statistics![Screenshot from 2020-08-18 13-54-59.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8f74b8cb-ab3b-4c55-abfd-a3c38d37e7a2)![Screenshot from 2020-08-18 13-54-28.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6ce17881-7502-4443-bed8-b02eaeef18ad)
### Intro Statistics![Screenshot from 2020-08-18 13-54-03.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/887b58e4-8576-47a0-8da3-f38231214731)
### Business Statistics![Screenshot from 2020-08-18 13-53-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/620716b5-7455-404a-be3d-e67ab0768d54)

PR #2610, opened 2020-08-13
Chapter review statistics

ajluk:
Issue: #2609


styles/templates/template2/parts/_titles-settings.scss
```diff
@@ -146,6 +148,7 @@
         ),
         H3LinkTitle: (
             font-family: (_ref: 'typography:::titleFont'),
+            color: (_ref: "scheme:::B5"),
```
ajluk, line 151: I have noticed that color of the link titles in the EoC  has been overwritten by other title components, probably because the selectors became more general in the corn theme. That is the reason I have added link color to all existing link tittles to prevent that behavior in the future. Otherwise we need to add a :not(a) selector to every title selector, where there is also a link title with corresponding tag (h3, h4, h5). Let me know what do you think about this solution.

Issue #2617, opened 2020-08-14
Implement Formula Review to Statistics


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/2614

MartaSztuka:
## PASSED
### Formula review
#### HS Statistics![Screenshot from 2020-08-28 11-17-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/389d2e7d-df0a-4f75-a7e4-e2209b4aaf5e)
#### Intro Statistics![Screenshot from 2020-08-28 11-19-50.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fbd3809a-8141-4bcd-813a-6f5041f394fe)
#### Business Statistics![Screenshot from 2020-08-28 11-20-37.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/1ffb141c-b717-43a7-8f74-e669a1f7ecd0)

PR #2626, opened 2020-08-17
Formula review statistics

ajluk:
Issue: #2617


styles/designs/corn/parts/_caption-components.scss
```diff
@@ -9,6 +9,7 @@ $Caption__Container: (
         margin-left:  enum('ValueSet:::OPTIONAL'),  //used in Figures
         margin-right:  enum('ValueSet:::OPTIONAL'), //used in Figures
         margin-bottom: enum('ValueSet:::OPTIONAL'), //used in Figures
+        break-inside: enum('ValueSet:::OPTIONAL'), //used in Tables in Formula Review
```
ajluk, line 12: I have added this property, because I have found spot where long table caption jumped to the next column: 
image
Here is how it looks after: 
image

wilkus27: I think tables should be centered. They are left aligned at this moment, based on the screenshots.

styles/designs/corn/parts/_eoc-shapes.scss
```diff
@@ -13,4 +15,24 @@
             ),
         )),
     )
+));
+
+@include create_shape('corn:::FormulaReviewShape',(
```
wilkus27, line 20: I think it would be better to add name to this shape related with its specific features. Otherwise we can't reuse it for other modules which should be displayed as Formula Review. So in two columns layout with text going down not to the right. I don't know, maybe `TwoColumnsVerticalEocShape`? @Stackblocks do you have other ideas?

wilkus27 (changes requested):

styles/designs/corn/parts/_eoc-shapes.scss
Stackblocks: I agree that a functionally descriptive name with a comment like `//used for Formula Review` is the way to go.

`TwoColumnVerticalEocShape` seems like a good option to me!

styles/designs/corn/parts/_eoc-components.scss
```diff
@@ -36,4 +36,6 @@ $Term--DL:(
         text-indent: h-spacing(-2),
         padding-left: h-spacing(2),
     ),
-);
\ No newline at end of file
+);
+
+$EoCSection__Container: empty_wrapper(EoCSectionContainer, ' > section');
```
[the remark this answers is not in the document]
ajluk: Yes, good point! Thank you Natalia!

styles/designs/corn/parts/_eoc-shapes.scss
ajluk: Thank you both! I agree the name should be more descriptive.

wilkus27 (approved):

Issue #2638, opened 2020-08-18
coreq math: unstyled elements nested in example solutions


lwtchu:
for some reason `class=unnumbered unstyled` tables  and `<span class="tokens">` are not being styled if they're nested in solutions
![Screen Shot 2020-08-18 at 9 17 41 AM](https://user-images.githubusercontent.com/11299125/90524788-1845d480-e134-11ea-892b-d5e1163a1e29.png)

this is in chapter 2

ajluk:
@lwtchu The issue with tables is a content issue. The class `unstyled` is connected with `os-coreq-element`. These classes should be separated: 
![image](https://user-images.githubusercontent.com/55203598/90608204-9375bc00-e202-11ea-9bc9-e31897cf69c8.png)

lwtchu:
@ajluk in the CNXML, we didn't add the os-coreq-element to the class  - could it be injected at some point??

i removed the solution word in teh content, please ignore the double solution, it was highlighted so i can find them.

ajluk:
@lwtchu Yes, it was. I wasn't aware this class was added via recipes. We need to do some changes there in this case.

ajluk:
@lwtchu here is how this example looks after spliting the classes via recipe: 
![image](https://user-images.githubusercontent.com/55203598/90753921-7bbd3700-e2d9-11ea-82a8-9a5168958a7b.png)

lwtchu:
@ajluk 😻  yay!! it looks great thank you

MartaSztuka:
# PASSED
## Example solutions
### College Algebra coreq![Screenshot from 2020-08-24 10-17-03.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7cbedd1e-f6f9-4812-a735-3d0aa8a6d0a8)

PR #2645, opened 2020-08-20
add space coreq class

ajluk:
Issue: #2638

### DIFF
--- a/recipes/mixins/_modify.scss
+++ b/recipes/mixins/_modify.scss
@@ -864,7 +864,7 @@
 @mixin modify_tagCoreqElement($selector) {
   [data-type="page"] section.coreq-skills {
     #{$selector} {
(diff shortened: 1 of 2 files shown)


recipes/mixins/_modify.scss
```diff
@@ -864,7 +864,7 @@
 @mixin modify_tagCoreqElement($selector) {
   [data-type="page"] section.coreq-skills {
     #{$selector} {
-      class: attr(class) "os-coreq-element" ;
+      class: attr(class) " os-coreq-element" ;
```
ajluk, line 867: I have added the additional space to the `os-coreq-element` class, because when it was added to element with existing class, the classes connected into one string and the tagged elements were loosing their styles.
image
Here is how the example looks with splitted classes: 
image

Issue #2630, opened 2020-08-17
circled list: remove indentation for first line if items go full width


lwtchu:
When the `class="circled"` list can't be aligned next to each toher, it weirdly indents the first line of text. If possible can we remove the indentation if the items don't fall next to each other?

This screenshot is from College Algebra with Coreq - can probably be found in any of the books that use tokens though.

![Screen Shot 2020-08-17 at 12 40 24 PM](https://user-images.githubusercontent.com/11299125/90426660-0c500900-e087-11ea-85f9-2eaf834efab5.png)

ajluk:
Hi @lwtchu! Could you provide the chapter/exercise, where you have found this issue if it's possible? I couldn't found this particular place. However I have checked these kind of lists when they are not next to each other and the look fine in College Algebra Coreq in the Answer Key: 
![image](https://user-images.githubusercontent.com/55203598/90513115-763be180-e15f-11ea-8fc1-b8ca564292d1.png)
![image](https://user-images.githubusercontent.com/55203598/90513305-b4d19c00-e15f-11ea-8de0-21a336f68269.png)
I have checked also Prealgebra 2e in Section Exercises and here it looks ok:
![image](https://user-images.githubusercontent.com/55203598/90513384-ddf22c80-e15f-11ea-9e66-909c0fd82017.png)
but I spotted different but I guess related issue:
![image](https://user-images.githubusercontent.com/55203598/90515278-b18bdf80-e162-11ea-9342-7c7be410b9fa.png)

ajluk:
@lwtchu I have tried to style those lists as we did to images and tables in exercises when they are a first child of the problem container and they look like this then: 
![image](https://user-images.githubusercontent.com/55203598/90525647-c3c04a80-e16f-11ea-9401-0cd4f66cf615.png)
![image](https://user-images.githubusercontent.com/55203598/90525708-d63a8400-e16f-11ea-8047-8f76466e04d3.png)

or you rather prefer them to be in one line with the problem number:
![image](https://user-images.githubusercontent.com/55203598/90526014-2c0f2c00-e170-11ea-9861-81e93b976d6a.png)
I will take more time to do this but is also achievable. Let me know what are your thoughts about this?
Is that look acceptable? They are a little lower than the problem number.

lwtchu:
this content should be in 6.1 Exponential Functions - i just migrated it. i'm trying to migrate the rest of chapter 6 but the server is being slow....but exponential functions should be in there :)

[...]

PR #2646, opened 2020-08-20
Circled lists issue

ajluk:
Issue: #2630


recipes/books/calculus/_config.scss
```diff
@@ -165,6 +165,7 @@ $Config_FirstElements: (
   (selector: ' > [data-type="media"]:first-child'),
   (selector: ' > .os-table:first-child'),
   (selector: ' > ol:first-child > li:first-child > [data-type="media"]'),
+  (selector: ' > ol.circled:first-child'),
```
ajluk, line 168: I have added this class to `$Config_FirstElements` to cover first circled lists children of problem containers. When the basic specs for problem containers are added these lists broken to another line line look like this (they jump over the problem number):
image
I have also couldn't use the specs for Images and Tables, because then the lists are little lower than the problem number: 
image
Only one property can be used there `vertical align: bottom` without top padding. I couldn't also applied this property to "regular" problem containers because then the math in para wouldn't be in one line with the problem number. It only woks in the Answer Key (I'm not sure why exactly).

recipes/mixins/_modify.scss
```diff
@@ -965,24 +965,36 @@
...
 
       @if $type == "Answer Key" { // see config.scss file for u-physics for details
         .os-eob [data-type="solution"] .os-solution-container {
-          #{$element} {
+          @if $element ==  " > ol.circled:first-child" {
+            #{$element} {
+              class: attr(class) " first-circled-list-element";
+            }
+          } @else { 
+            #{$element} {
```
ajluk, line 989: I have modified the existing mixin for adding first element class instead of creating a new one. Let me know what do you think about it?

recipes/mixins/_modify.scss
```diff
@@ -994,6 +1006,12 @@
...
       }
+      .first-circled-list-element {
+        string-set: hasFirstElement " has-first-circled-list-element";
+      }
+      &:deferred {
+        class: attr(class) string(hasFirstElement);
+      }
     }
 
     @each $page in $Config_BookCompositePages {
```
ajluk, line 1017: I have also modified this mixin, but it has s defect that this change affects output css files for all books. I think it can't done any harm, because they don't have the `first-circled-list-element` class util it is added in the `config.scss file`. But let me know what are your thought about that solution.

recipes/mixins/_modify.scss
```diff
@@ -1008,6 +1026,12 @@
           &:deferred {
             class: attr(class) string(hasFirstElement);
           }
+          .first-circled-list-element {
+            string-set: hasFirstElement " has-first-circled-list-element";
+          }
+          &:deferred {
+            class: attr(class) string(hasFirstElement);
+          }
```
ajluk, line 1034: See my comment above.

styles/templates/template2/parts/_answerkey-settings.scss
```diff
@@ -23,11 +23,13 @@
         ),
         SolutionContainer: (
             padding-left: v-spacing(1),
-            vertical-align: bottom,
```
ajluk, line 26: I have removes this property, because it wasn't needed any longer.

recipes/output/calculus.css
```diff
@@ -2552,6 +2552,12 @@
 :pass(12) .os-eob [data-type="solution"] .os-solution-container > ol:first-child > li:first-child > [data-type="media"] {
   class: attr(class) " first-element"; }
 
+:pass(12) [data-type="exercise"]:not(.unnumbered) .os-problem-container > ol.circled:first-child {
+  class: attr(class) " first-circled-list-element"; }
+
+:pass(12) .os-eob [data-type="solution"] .os-solution-container > ol.circled:first-child {
+  class: attr(class) " first-circled-list-element"; }
+
```
ajluk, line 2560: In the corn theme books this lines are added as well. They are missing in the rest of the books output css files.

recipes/mixins/_modify.scss
brittany-johnson: Hello! @wilkus27 I'm going to summarize how I understand your suggestion above to make sure that I understand you correctly. Please let me know if I got this right: 

- Add an optional key to the `$Config_FirstElements` map. 
- In this optional key, we will specify the element type. 
- We will then use the element type specified in the map to add the class instead of using `ol.circled:first-child`

If this is correct, then I agree. I think its always a good idea to make solutions that we can reuse!

brittany-johnson: @ajluk If you go with this solution, please re-assign me for review once it's implemented. Thank you!

styles/designs/corn/parts/_exercises-shapes.scss
```diff
@@ -155,6 +187,38 @@
...
                                 )),
+                                map-merge($Exercise__Problem__Container--WithCircledListOnTop, (
+                                    _components: (
+                                        $Exercise__ProblemSolution__Paragraph,
+                                        $Image,
+                                        map-merge($Table--Exercise__Container, (
+                                            _components:(
+                                                map-merge($Table, (
+                                                    _components:(
+                                                        map-merge($Table__Head, (
```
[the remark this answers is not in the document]
ajluk: Yes, definitely!

recipes/mixins/_modify.scss
wilkus27: @brittany-johnson Yes, this is what I meant. :slightly_smiling_face:

recipes/books/_common-stuff.scss
```diff
@@ -71,6 +71,7 @@ $Config_TargetLabels: (
 $Config_Notes: ();
 $Config_UnnumberedElements: ();
 $Config_FirstElements: ();
+$ConfighasFirstInlineListElement: false;
```
ajluk, line 74: I have created this config conditional to use it in the if statement, where I couldn't use `elementType` condition.

recipes/mixins/_modify.scss
```diff
@@ -994,6 +1006,14 @@
       &:deferred {
         class: attr(class) string(hasFirstElement);
       }
+        @if $ConfighasFirstInlineListElement == true {
```
ajluk, line 1009: The config I have created at the beginning is used in this mixin where I couldn't use `$type == inlineList`. I've got errors that it is undefined for this mixin. Let me know if this solution is acceptable.

recipes/mixins/_modify.scss
ajluk: @wilkus27 thank you for all your suggestions! :heart:  I have applied the requested changes. @brittany-johnson @Stackblocks  Let me know what do you think as well.

recipes/books/calculus/_config.scss
```diff
@@ -165,8 +165,11 @@ $Config_FirstElements: (
   (selector: ' > [data-type="media"]:first-child'),
   (selector: ' > .os-table:first-child'),
   (selector: ' > ol:first-child > li:first-child > [data-type="media"]'),
+  (selector: ' > .circled:first-child', elementType: inlineList),
```
ajluk, line 168: I have also changed this selector to cover both `ol` and `ul` circled lists.

styles/designs/corn/parts/_exercises-shapes.scss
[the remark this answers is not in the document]
brittany-johnson: I was also thinking that when reviewing, very large shape 😭

recipes/mixins/_modify.scss
```diff
@@ -994,6 +1005,14 @@
...
         class: attr(class) string(hasFirstElement);
       }
+        @if $Config_hasFirstInlineListElement == true {
+          .first-inline-list-element {
+            string-set: hasFirstElement " has-first-inline-list-element";
+          }
+          &:deferred {
+            class: attr(class) string(hasFirstElement);
+          }
+      }
```
wilkus27, line 1015: I'm not sure if it's necessary to create new config. I think it would be better to use one condition everywhere. Either `$Config_hasFirstInlineListElement` or `$elementType`. Both telling us that there is inline list at the beginning.

wilkus27 (changes requested):

recipes/mixins/_modify.scss
wilkus27: As we discussed in this case it's hard to use `$elementType` key because config `$Config_FirstElements` has nested maps. To extract it we need to use loop `each` what causes a lot of repetitions. So let's keep both condition. :sunglasses:

wilkus27 (approved):

Issue #2445, opened 2020-06-24
corn theme: Titles shapes convention


ajluk:
Styling Titles in Corn Theme needs decision what approach we should take/be the best solution.
Related document with some propositions: https://docs.google.com/document/d/1RS7ostFOCoF378p_Gi-ganVkEEKlimGW3zXFo6HEPOU/edit#heading=h.6vastgn7j4uy

ajluk:
@MartaSztuka Information for testing purposes:
This is a clean up issue. I have cleaned the titles in the Answer Key (moved to the title shape), so this title needs to be checked in at least in one book from each dev-math, calculus,  preacalculus or precalculus-coreq and nothing should be changed there.
Screenshot for reference (Prealgebra 2e):
![image](https://user-images.githubusercontent.com/55203598/93208833-7ed30800-f75d-11ea-94cb-d3855f424a7f.png)

The paragraph titles has been also cleaned and need to be checked if there are no differences in the books from above:
Screenshot for reference (Prealgebra 2e):
![image](https://user-images.githubusercontent.com/55203598/93208958-aaee8900-f75d-11ea-9bf1-90f788dce3e6.png)

 
Within this PR I have also added page breaks to paragraph titles in the Statistics book (this is new), so they need to be tested here as well in one from the statistics books (they shouldn't be orphaned or broken inside into another page): 
Screenshots for reference:
![image](https://user-images.githubusercontent.com/55203598/93208599-18e68080-f75d-11ea-84bf-f4a5e7ae50a3.png)
![image](https://user-images.githubusercontent.com/55203598/93208408-cf963100-f75c-11ea-8d0b-bc001f86e3a9.png)
![image](https://user-images.githubusercontent.com/55203598/93208439-dae95c80-f75c-11ea-8d95-24045d0c77f8.png)

MartaSztuka:
# PASSED
## Titles
### Precalculus
![Screenshot from 2020-09-16 13-04-41](https://user-images.githubusercontent.com/47241114/93329319-501c6680-f81d-11ea-943d-e90487cef07b.png)
![Screenshot from 2020-09-16 13-04-16](https://user-images.githubusercontent.com/47241114/93329320-514d9380-f81d-11ea-9b20-7304e301b6fa.png)
### Prealgebra
![Screenshot from 2020-09-16 13-07-44](https://user-images.githubusercontent.com/47241114/93329549-a4274b00-f81d-11ea-850a-c4a0193ca995.png)
[...]

PR #2667, opened 2020-08-24
Clean up titles corn

ajluk:
Issue: #2445


styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -120,6 +120,26 @@ div.appendix {
...
+.os-eoc.os-practice-test-container > section.practice-test > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p > strong
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > p > strong,
+.os-eoc.os-review-exercises-container > section.review-exercises > p,
+.os-eos.os-section-exercises-container > section.section-exercises > p > strong,
+.os-eos.os-section-exercises-container > section.section-exercises > p {
+  break-after: avoid;
+  break-inside: avoid;
+}
```
ajluk, line 140: I have moved those to `page-breaks.scss file`, but these are specific only for corn theme. Maybe a better solution could be to create a separate `page-break.scss` file for corn and carnival or leave these properties in the components? Let me know what do you think.

styles/designs/corn/parts/_titles-shapes.scss
```diff
@@ -78,6 +78,7 @@
     _components: (
         map-merge($Chapter--Composite__Container, (
             _components: (
+                $H1,
```
ajluk, line 81: I have added the H1 title here to make this shape cover the Answer Key as well.

styles/designs/corn/parts/_answerkey-shapes.scss
```diff
@@ -8,11 +8,8 @@
   _components: (
     map-merge($AnswerKey__Container, (
       _components: (
-        $H1,
-        $H3,
         map-merge($AnswerKey__CompositePage, (
           _components: (
-            $H2,
```
ajluk, line 15: I have removed H1, H2, H3 titles from this shape, because title shape can cover them now.

styles/designs/common/pdf/_page-breaks.scss
```diff
@@ -120,6 +120,26 @@ div.appendix {
...
+.os-eoc.os-practice-test-container > section.practice-test > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p > [data-effect="bold"]
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > p > [data-effect="bold"],
+.os-eoc.os-review-exercises-container > section.review-exercises > p,
+.os-eos.os-section-exercises-container > section.section-exercises > p > [data-effect="bold"],
+.os-eos.os-section-exercises-container > section.section-exercises > p {
+  break-after: avoid;
+  break-inside: avoid;
+}
```
ajluk, line 140: I have moved those to page-breaks.scss file, but these are specific only for corn theme. Maybe a better solution could be to create a separate page-break.scss file for corn and carnival or leave these properties in the components? Let me know what do you think.

brittany-johnson: I think we should keep the `common` page-break file for the library-wide elements, but for elements like the ones above, we could have a page-break file for corn that imports the common page break file. What do you think? 

The reason why we opted to have a file just for page breaks is that in the past with oer.exports, the page breaks were scattered all over the repo and it was difficult to debug.

wilkus27: I think it's a very good idea. Everything will be clear and such solution will not have an impact on Carnival theme. 

@ajluk I have one more question. Is it necessary to use specific classes? 
For example here:
image

Doesn't first selector cover two others marked by arrow?

wilkus27 (changes requested):

styles/designs/corn/pdf/_page-breaks.scss
```diff
@@ -0,0 +1,13 @@
...
+.os-eos.os-section-exercises-container > section.section-exercises > section > p,
+.os-eos.os-section-exercises-container > section.section-exercises > section.writing > p,
+.os-eos.os-section-exercises-container > section.section-exercises > section.everyday > p,
+.os-eoc.os-practice-test-container > section.practice-test > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > p,
+.os-eos.os-section-exercises-container > section.section-exercises > p {
+  break-after: avoid;
+  break-inside: avoid;
+}
```
wilkus27, line 13: @ajluk In my last comment I meant classes next to `section`:

image

My question was, is it necesarry to add classes if in the first line it is applied to all sections. It should cover also those with `.writing` and `.everyday`

wilkus27 (changes requested):

styles/designs/corn/pdf/_page-breaks.scss
```diff
@@ -0,0 +1,15 @@
...
+// Paragraph titles in Section Exercises and EoC exercises shouldn't be orphaned or broken inside
+.os-eos.os-section-exercises-container > section.section-exercises > section > p,
+.os-eoc.os-practice-test-container > section.practice-test > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > section > p,
+.os-eoc.os-review-exercises-container > section.review-exercises > p,
+.os-eos.os-section-exercises-container > section.section-exercises > p,
+.os-eoc.os-practice-container > section.practice > p,
+.os-eoc.os-free-response-container > section.free-response > p,
+.os-eoc.os-bring-together-homework-container > section.bring-together-homework > p,
+.os-eoc.os-bring-together-exercises-container > section.bring-together-exercises > p {
```
ajluk, line 12: I have also added selectors for paragraphs present in Statistics books. These aren't styled differently as in the other books from corn, but also needed these page break properties:
image
image

styles/designs/corn/pdf/_page-breaks.scss
ajluk: @wilkus27 as we discussed I didn't understand your suggestion properly, but thank to this was able to notice I can make these selectors a little more general. :grinning:

styles/designs/corn/pdf/_page-breaks.scss
brittany-johnson: Just curious if we need to have the class-specific selector for each one of these? For example: 
Instead of `.os-eos.os-section-exercises-container > section.section-exercises > p,` 
Will `.os-eos.os-section-exercises-container > section > p,` work?

brittany-johnson (approved):
Just had one question out of curiosity

styles/designs/corn/pdf/_page-breaks.scss
ajluk: @brittany-johnson I was thinking about that as well, but it wouldn't reduce the amount of selectors, because we still need these sections specific selectors, so I have left them: 
image
But as you suggested it works without them (I have checked), so after your comment I have removed them.

Issue #2657, opened 2020-08-21
Business stats - Tables not styled


meeragovindaraju:
Business stats - COPS Production

**PDF:**
![Screen Shot 2020-08-21 at 11.11.47 AM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/08f94806-2b15-4432-a70c-c727ee5da2d9)

**Webview**
![Screen Shot 2020-08-21 at 11.33.26 AM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/50fd48f7-8d16-4b88-84eb-04c39862b372)# What happened?

meeragovindaraju:
Looks good on suspicious.sting branch

![Screen Shot 2020-08-26 at 4 26 24 PM](https://user-images.githubusercontent.com/33066187/91358713-e7f7d900-e7b8-11ea-9a30-61f02c147848.png)

PR #2682, opened 2020-08-24
Style one column tables

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2657


styles/designs/corn/parts/_table-components.scss
```diff
@@ -60,7 +60,7 @@ $Table__Row: empty_wrapper(TableRow, ' > tr');
 
 $Table__Data: (
     _name: "TableData",
-    _subselector: " > td:not(:only-of-type)",
+    _subselector: " > td",
```
wilkus27, line 63: I checked and removing this doesn't have impact on multicolumns tables:

image

image

ajluk (approved):

Issue #2661, opened 2020-08-21
Note box not uniform


meeragovindaraju:
Chapter 9 Homework section

COPS production PDF

![Screen Shot 2020-08-21 at 2.26.04 PM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/d666fe87-d2cb-454e-9997-8b28e8c8e23f)

![Screen Shot 2020-08-21 at 2.28.51 PM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/cc3d5116-5163-446f-9333-c92d9f630ccf)

Stackblocks:
![image](https://user-images.githubusercontent.com/12678235/91356130-c72d8480-e7b4-11ea-8c7f-fe89ab2856da.png)

meeragovindaraju:
Looks good in the Homework section

![Screen Shot 2020-08-28 at 2 29 01 PM](https://user-images.githubusercontent.com/33066187/91608813-d7bc3700-e93b-11ea-9e5b-c1f43f61dcd8.png)

![Screen Shot 2020-08-28 at 2 27 33 PM](https://user-images.githubusercontent.com/33066187/91608816-da1e9100-e93b-11ea-91c9-f3eb04ad0a74.png)

![Screen Shot 2020-08-28 at 2 27 23 PM](https://user-images.githubusercontent.com/33066187/91608830-e1459f00-e93b-11ea-920a-322c07db828f.png)

meeragovindaraju:
This is how it looks in the chapter section 
section 10.4 example 10.12

![Screen Shot 2020-08-28 at 2 28 23 PM](https://user-images.githubusercontent.com/33066187/91608931-1225d400-e93c-11ea-9ff3-0466fe76541b.png)

lwtchu:
@meeragovindaraju can you check again? the "note" title isn't enclosed in the box in the first few screenshots

Stackblocks:
Ok, it should be working now 🤞

meeragovindaraju:
It looks great now!

![Screen Shot 2020-08-31 at 9 12 44 AM](https://user-images.githubusercontent.com/33066187/91729459-349d3480-eb6a-11ea-8064-483394d8511c.png)

![Screen Shot 2020-08-31 at 9 12 35 AM](https://user-images.githubusercontent.com/33066187/91729461-3535cb00-eb6a-11ea-80fa-52aa5311eeaf.png)

[...]

PR #2722, opened 2020-08-26
Fix NoClassNote title gap in Stats EOC

Stackblocks:
FIX: Remove space between NoClassNote body and title in EOC

    - Update title component to address unintended inheritance from
    generic `h3[data-type="title"]` targeted components
    - Set `margin-bottom: 0` for Statistics to remove gap
    - Regenerate styles

    Refs: #2661

---

image


styles/designs/corn/parts/notes/_notes-components.scss
```diff
@@ -203,6 +203,15 @@ $Notes--Boxed__Title: (
     )
 );
 
+// Used to prevent unwanted inheritance from generic h3 styling without using !important
+$Notes--Boxed__Title--Strong: (
```
brittany-johnson, line 207: I'm inclined to try to use the `$Notes--Boxed__Title` that we already have. 
Have you tired adding an additional selector to the `$Notes--Boxed__Title`? Something like this: 
`_subselector: ' > .os-title,  > h3[data-type="title"].os-title',`

Then you could add `margin-bottom: enum('ValueSet:::OPTIONAL'),` to `$Notes--Boxed__Title` and set the margin to `0` in the `DefaultNoClassNote` settings block in `books.scss`.

Stackblocks: I'll give that a shot! I tried the ` > .os-title, > h3[data-type="title"]` compound selector like that before and couldn't get it to work but it would make sense that since this is a new selector it might. 🤞

ajluk (approved):

Issue #2629, opened 2020-08-17
corn theme: aligning exercises with tall math in multi column layout


lwtchu:
![Screen Shot 2020-08-13 at 11 10 52 AM](https://user-images.githubusercontent.com/11299125/90423990-02c4a200-e083-11ea-896b-5a5250bfca73.png)

We increased the math size. Can we please have the numbers for each exercise align to the middle of the line height set by the new increase math size if there's a fraction? 

Let me know if there's any questions. This content is from College Algebra/College Algebra with Coreq. However, you can probably find instances in the dev math or precal trio.

wilkus27:
@lwtchu I looked into it and I'm afraid it's impossible to align exercises in this way. Problem is that exercises are independent of each other, so height of the one exercise don't have an impact on height of another exercise.


@brittany-johnson @Stackblocks I have one idea how it could be resolved. Maybe we can put each three exercises in one container div or section with some class, something like that:

```
<section class="some-class">
  <div data-type="exercise"></div>
  <div data-type="exercise"></div>
  <div data-type="exercise"></div>
</section>
<section class="some-class">
  <div data-type="exercise"></div>
  <div data-type="exercise"></div>
  <div data-type="exercise"></div>
</section>
```
and then set `display: flex` or `display: table` to section , not sure what would be better. Generally it should be easier to align these three siblings in one line.

But, I see a few problems:

1. Where to create sections/divs? In the content or by recipes?
Probably it would be safer to do it by recipe in the last pass, but I'm not sure if we can do it using sass. 

2. We need such markup to style PDFs. It might have huge impact on webview, where exercises are not in columns.

3. I'm not 100% sure that putting three exercises in one parent container will help. Exercises start with paragraphs, equations, tables, figures, paragraphs with equations inside. It might be hard to fix all instances.

What do you recommend? Do you have any other ideas how we can handle this?

lwtchu:
@wilkus27 hmm let's park this tomorrow for the call. i can join

wilkus27:
@meeragovindaraju It should work now. Applied changes could have a huge impact on all types of exercises, that's why we decided it should be tested on feature branch `exercises-alignment`. 

Please check:
- three columns exercises
- two columns exercises
- one column exercises
- answer key
- solutions (EOC module from Statistics)

meeragovindaraju:
- College-algebra

Three column exercise

[...]

PR #2723, opened 2020-08-26
[WIP] issue-2629

brittany-johnson:
https://github.com/openstax/cnx-recipes/issues/2629


styles/designs/common/pdf/_functions.scss
```diff
@@ -95,7 +95,6 @@
       _name: #{$name},
       _subselector: ' .os-#{$child}-container',
       _properties:(
-          display: table-cell,
```
brittany-johnson, line 98: Because removing this generated changes in all the books I believe we have two options: 
1. change the display to `optional` and add `display: table-cell` only where it is needed
2. visually check each place that is affected to ensure that there are no negative effects

styles/designs/corn/parts/_exercises-components.scss
```diff
@@ -30,7 +30,7 @@ $ExercisesModule--Multicolumn__Container: (
   _properties: (
       margin-bottom: v-spacing(2),
       width: 100%,
-      display: enum('ValueSet:::REQUIRED'),
```
brittany-johnson, line 33: We'll have to check all of the places where the multi-column container is used in the corn theme to confirm that there are no negative effects.

styles/designs/corn/parts/_exercises-components.scss
```diff
@@ -112,7 +112,7 @@ $Exercise__ProblemSolution__Number: (
   _name: "ProblemSolutionNumber",
   _subselector: ' > .os-number',
   _properties: (
-    display: table-cell,
+    display: enum('ValueSet:::OPTIONAL'),
```
brittany-johnson, line 115: If `table-cell` is required in some places we can add it via settings

styles/designs/common/pdf/_functions.scss
wilkus27: This function was created for Carnival theme, where text in columns goes from the top to the bottom. Maybe we shouldn't use it in Corn theme, instead we can create components with needed properties in corn files. What do you think?

styles/designs/corn/parts/_exercises-components.scss
```diff
@@ -30,7 +30,7 @@ $ExercisesModule--Multicolumn__Container: (
   _properties: (
       margin-bottom: v-spacing(2),
       width: 100%,
-      display: enum('ValueSet:::REQUIRED'),
+      align-items: baseline,
```
wilkus27, line 33: I think we can remove this. I believe this property works only when component has also property `display: flex`.

styles/designs/corn/parts/_exercises-components.scss
```diff
@@ -71,7 +71,7 @@ $Exercise__Problem__Wrapper: (
     _name: "ProblemWrapper",
     _subselector: " [data-type='problem']",
     _properties:(
-        display: table,
+        display: flex,
```
wilkus27, line 74: I'm wondering if it's necessery to use flexbox here. I think we can obtain the same effect when we just add `vertical-align: bottom` to `.os-problem-container`. At this moment we add this property only to problem containers with image/table on top. We didn't add this property everywhere because then exercises which start with paragraph containing math displayed in a wrong way. I checked and this problem also appears with changes from this PR:

image

(word "Solve" is not in line with exercise number). But if it's acceptable we can create just one component for problem container. It will have the same properties as `$Exercise__Problem__Container--WithImageTableOnTop`.

wilkus27 (changes requested):
@brittany-johnson I added some comments. Please let me know what do you think. If you think it is worth to try those ideas I can take care of it. Maybe in separate PR, in case something wouldn't work.

styles/designs/corn/parts/_exercises-components.scss
wilkus27: I checked and `vertical-align: bottom` doesn't work as I expected. So it's better to stay with `display: flex` :wink:

wilkus27:
@brittany-johnson I was trying to finish that and apply this solution also to Answer Key. Unfortunatelly, we can't do this :disappointed: The problem is in markup. In Eoc exercises we have:

```
section
  [data-type="exercise"]
    [data-type="problem"]
```
and `display: flex` is applied to exercise container. In Answer Key there is just:

```
section
  [data-type="solution"]
```
We can't apply flexbox to the whole section because Prince have problems with big elements with this property.

I tried to find another solution. Here is my PR https://github.com/openstax/cnx-recipes/pull/2751.

brittany-johnson:
> @brittany-johnson I was trying to finish that and apply this solution also to Answer Key. Unfortunatelly, we can't do this 😞 The problem is in markup. In Eoc exercises we have:
> 
> ```
> section
>   [data-type="exercise"]
>     [data-type="problem"]
> ```
> 
> and `display: flex` is applied to exercise container. In Answer Key there is just:
> 
> ```
> section
>   [data-type="solution"]
> ```
> 
> We can't apply flexbox to the whole section because Prince have problems with big elements with this property.
> 
> I tried to find another solution. Here is my PR #2751.

Okay, this is good to know. Thank you for testing this so thoroughly! I will close this PR now and review your new one!

Issue #2702, opened 2020-08-25
Elementary Algebra -  Wild dot above 'In the following exercise (...)`


MartaSztuka:
### There is singe dot above 'In the following exercise (...)` which is unnecessary ![Screenshot from 2020-08-25 13-22-50.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/291d1102-0c5c-4cad-9f79-df96803f81b1)

lwtchu:
I think this is related to #2688. All of that is supposed to be on paragraph.

ajluk:
Screenshot for reference: 
![image](https://user-images.githubusercontent.com/55203598/91824443-13584900-ec3b-11ea-9091-768ec39e6fc5.png)

meeragovindaraju:
Looks good!

![Screen Shot 2020-08-31 at 9 55 10 AM](https://user-images.githubusercontent.com/33066187/91865323-e3a74200-ec36-11ea-8877-4f629c3ec788.png)

meeragovindaraju:
Looks good!

![Screen Shot 2020-09-01 at 9 39 47 AM](https://user-images.githubusercontent.com/33066187/91865534-1bae8500-ec37-11ea-8c03-2446d63c2392.png)

PR #2725, opened 2020-08-27
remove unused property

ajluk:
Issue: #2702


styles/designs/corn/parts/_exercises-components.scss
```diff
@@ -44,7 +44,6 @@ $Exercises_Paragraph__Title: (
         line-height: line-height(0),
         margin-top: v-spacing(1),
         font-style: normal,
-        display: block,
```
ajluk, line 47: This property was causing the issue. I have investigate it and removing it doesn't change anything, because probably it was left after some attempts to create column layout. I have checked few different books from corn theme. 
image

wilkus27 (approved):

Issue #2721, opened 2020-08-26
college algebra/college algebra coreq - nested lists in how to not indenting properly


lwtchu:
![Screen Shot 2020-08-26 at 1 04 50 PM](https://user-images.githubusercontent.com/11299125/91340081-05b74500-e79d-11ea-8669-b1a16db57f3b.png)

i looked in a few places but not sure if this was addressed with any existing pr.

the a/b list is nested within the numeric list. can we make sure that nested lists get proper indentation in features?

This is in Chapter 6.4 Graphs of Logarithmic Functions; handful of other instances peppered throughout chapter 6

lwtchu:
![Screen Shot 2020-08-26 at 1 29 26 PM](https://user-images.githubusercontent.com/11299125/91342139-38af0800-e7a0-11ea-8b1a-252488917f37.png)
another example from 8.1

wilkus27:
@lwtchu This is how it looks now:

![image](https://user-images.githubusercontent.com/40228854/91548374-915bdd80-e925-11ea-8a4d-78ce38e4fa81.png)

![image](https://user-images.githubusercontent.com/40228854/91548499-be0ff500-e925-11ea-89e0-1d93e3ce1d0e.png)

Is it ok?

lwtchu:
@wilkus27 can the a start under "vertices" - the bullets start right under the text so the alpha lists should too

wilkus27:
@lwtchu To align "a" with the text from above I had to use hard coded value `2rem` for margin:

![image](https://user-images.githubusercontent.com/40228854/91835519-cb8cee00-ec49-11ea-98cc-a29ca83c8888.png)

As @Stackblocks suggested in my PR it would be better to use created function `h-spacing`, but then "a" is moved more on the right:

![image](https://user-images.githubusercontent.com/40228854/91835694-042cc780-ec4a-11ea-89fb-a681401b42f8.png)

Are you ok with that? Or we should stay with `2rem`?

lwtchu:
@wilkus27 can we go with the h-spacing thanks.

brittany-johnson:
Verified on development branch:
![Screen Shot 2020-09-01 at 2 41 21 PM](https://user-images.githubusercontent.com/24909007/91898607-98565900-ec61-11ea-9653-4f8d0ab0e097.png)

meeragovindaraju:
Looks good on the feature branch College Algebra with coreq

[...]

Issue #2608, opened 2020-08-13
Rework lists in notes from Corn Theme


wilkus27:
Lists components should be moved from notes shapes in Corn Theme. We need to create separate shape for them in `lists-shapes.scss` file. 

Link to PR with conversation about it: https://github.com/openstax/cnx-recipes/pull/2599/files#r468555490

wilkus27:
Will be resolved within https://github.com/openstax/cnx-recipes/pull/2726

wilkus27:
Changes for this issue were merged directly to master and there it should be tested. I think the best way to do that is to check lists in different notes in a few books from Carnival Theme.

@brittweinstein Can you recommend which books should be tested?

lwtchu:
@wilkus27 can you give me a little more detail about the change? what is the scope? does it impact all lists in all books? and what is the expected change (or is it not visible to the human eyes :joy:?)

lwtchu:
actually alan explained it to me (thank you @Stackblocks )

can we do concentrated testing on lists within notes for the followign titles

statistics (high school)
calculus volume 2
college algebra with coreq
interediate algebra

maybe only like...1 chapter each and only concentrating on notes.

wilkus27:
@lwtchu Sorry, I should add more details in the description :grin:  I will put it here.

Previously if someone noticed some unstyled list in note, we styled this case only in shape of this specific note, e.g iconBoxed.
Now all of those cases are moved to the separate shape for lists in notes, which is applied to all notes except three:

- with class `media-2`
- with class `media-notitle`
- with class `chapter-objectives`

First two contain unordered lists for which we remove bullets:

![image](https://user-images.githubusercontent.com/40228854/92010642-5228f480-ed4a-11ea-92bd-3678fc72ee1b.png)

Third one is styled like Learning Objectives in Statistics:

![image](https://user-images.githubusercontent.com/40228854/92010772-7c7ab200-ed4a-11ea-9e85-edd54e97e5df.png)

(NOTE: headers of these notes are fixed in another issue).

So there might be some visible changes, because notes get additional settings for some lists.

meeragovindaraju:
@wilkus27 I have tried to locally generate the PDFs from **master** branch but the books are unstyled. I have tried hs-stats and calculus vol2. Please advise on how to proceed.

meeragovindaraju:
Clac-vol2 looks good

[...]

PR #2726, opened 2020-08-28
Create shape for lists in notes

wilkus27:
Issues: https://github.com/openstax/cnx-recipes/issues/2721 and https://github.com/openstax/cnx-recipes/issues/2608


styles/designs/corn/parts/_lists-components.scss
```diff
@@ -40,7 +40,7 @@ $OrderedList--SecondLevel: (
     _name: "SecondLevelOrderedList",
     _subselector: " > li  > ol:not(.circled):not(.os-stepwise)",
     _properties:(
-        margin-left: h-spacing(2),
+        margin-left: 2rem,
```
wilkus27, line 43: I changed it because with `h-spacing(2)` it didn't look good:

image

This is how it looks now:

image

For unordered lists `h-spacing(2)` is ok. They look good with such margin:

image

styles/templates/template2/parts/_lists-settings.scss
```diff
@@ -28,3 +28,9 @@
         ),
     ),
 ));
+
+@include add_settings((
+    NotesLists: (
+        _selectors: ('[data-type="note"]:not(.media-2):not(.media-notitle):not(.chapter-objectives) > .os-note-body'),
+    ),
+));
```
wilkus27, line 36: I put in the `:not` pseudo selector all notes from Corn theme where lists are styled differently. Please let me know what do you think about it? Maybe it will be better to define selector for `NotesLists` in book.scss files. Then selector will contain only those classes which appear in the book. E.g `.chapter-objectives` exists only in Statistics.

brittany-johnson:
@wilkus27 Hi! I just changed the base branch to master. I did this because we have to do another deploy to cops for coreq and for us to do that, it needs to be merged into master. If you have any questions please let me know!

styles/designs/corn/parts/_lists-components.scss
Stackblocks: Would `h-spacing(3)` be too much? I think if possible we should stick with the scale so if we make changes the changes are applied uniformly. If not, then this'll do as a band-aid.

styles/templates/template2/parts/_lists-settings.scss
ajluk: I think it's clear. Maybe a comment about the other notes, where lists are styled differently is the reason why they can't use this one in the code would be a useful?

ajluk (approved):

styles/designs/corn/parts/_lists-components.scss
wilkus27: @Stackblocks Hmm :thinking: Actually it also looks good:

image

It is moved just a little bit. Let me ask Larissa if she is ok with that. I also think it's better to use created functions for consistency.

wilkus27:
Changes from this PR were moved to https://github.com/openstax/cnx-recipes/pull/2734. This can be closed now.

PR #2751, opened 2020-09-07
Exercises alignment

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2629


wilkus27:
@brittany-johnson @Stackblocks I think I have found solution for exercises alignment. Here is example from the issue:

image

But, there is a lot of different instances: exercises in one, two, three columns; which starts with images, tables, circled lists; solutions in Answer Key;  exercises with special classes. I was trying to cover everything but I could miss something. Do you think we should test it on feature branch?

brittany-johnson:
> @brittany-johnson @Stackblocks I think I have found solution for exercises alignment. Here is example from the issue:
> 
> image
> 
> But, there is a lot of different instances: exercises in one, two, three columns; which starts with images, tables, circled lists; solutions in Answer Key; exercises with special classes. I was trying to cover everything but I could miss something. Do you think we should test it on feature branch?

Yes, I think that would be best. All of the EOC, EOS, and answer key content can be tested just to make sure everything is covered. 
Thanks for working on this!

brittany-johnson (approved):
The code looks good to me in this current state but if you have to add more changes after testing on the feature branch feel free to add me again to review!

Stackblocks (approved):
Looks good! Thanks so much for looking into this.

PR #2763, opened 2020-09-10
Update BUG TEMPLATE

lwtchu:
Can you guys review to see if this has everything you'll need? Updating the bug template so people will be able to provide us with better information.


meeragovindaraju:
@lwtchu One more thing that will be useful for testing is knowing how to go about producing the PDF, Local PDF using CLI or COPS or Generating Testable Data for “Recipes” (Structural) Changes in Cnx-Recipes or Generating Testable Data for “Styles” (Surface-Level) Changes in Cnx-Recipes or any other way that is appropriate.

lwtchu:
@meeragovindaraju the issue with that is that testing locations vary depending on where the code is. i think a better way to tell is to see where to test is to see where the PR is on or where the bug was located. we also have people outside of our team opening issues and they won't know the details of where to generate the pdf.

i can also add if the person reporting suspects if it's a recipe or style change. however, i'll add the caveat that sometimes it's both or it's not clear in the initial bug. the information that you are seeking are usually answered _after_ the investigation or the fix has been made.

PR #2764, opened 2020-09-11
trash elements in first pass

brittany-johnson:
The coreq skills elements were being numbered and then trashed, causing a numbering discrepancy. I think in all cases we should trash `$Config_TrashSelectedElements` in the first pass so we can avoid issues like this. Please let me know if you have any concerns.

### DIFF
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -65,7 +65,7 @@
 :pass(0) {
   //putting all unnecessary elements in the trash and emptying the trash
   //first step so that these elements are not counted
-  @if $Config_TrashSelectedElements {
+  @if $Config_TrashSelectedElements != () {
     @each $trash in $Config_TrashSelectedElements {
       @include modify_trash($trash);
     }
@@ -616,6 +616,12 @@
 
   @if release-flag-enabled(custom-lists) {
     @include modify_lists($Config_CustomList);
+  }
+
+  @if $Config_TrashSelectedElements != () {
+    @each $trash in $Config_TrashSelectedElements {
+      @include modify_trash($trash);
+    }
   }
 }
 :pass(11) {
--- a/recipes/books/_generator.scss
+++ b/recipes/books/_generator.scss
@@ -63,6 +63,15 @@
 
 
 :pass(0) {
+  //putting all unnecessary elements in the trash and emptying the trash
+  //first step so that these elements are not counted
+  @if $Config_TrashSelectedElements {
+    @each $trash in $Config_TrashSelectedElements {
+      @include modify_trash($trash);
+    }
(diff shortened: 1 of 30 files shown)


recipes/books/_generator.scss
```diff
@@ -63,6 +63,15 @@
...
 :pass(0) {
+  //putting all unnecessary elements in the trash and emptying the trash
+  //first step so that these elements are not counted
+  @if $Config_TrashSelectedElements {
+    @each $trash in $Config_TrashSelectedElements {
+      @include modify_trash($trash);
+    }
+    @include utils_clearTrash();
+  }
+
```
wilkus27, line 74: I have a question here :) What is the reason that `utils_clearTrash` is applied here? We include it also in pass 100. I'm just curious :grin: I'm also wondering if we can change condition to `@if $Config_TrashSelectedElements != ()`. Then output will change only for books which have any elements to trash.

wilkus27 (changes requested):
I added some thoughts :slightly_smiling_face:

recipes/books/_generator.scss
brittany-johnson: I first tried this solution without clearing the trash but for some reason the coreq skills section was still being counted 🤔 so that’s why I’m emptying it twice 😂 

Ooooo and yes I didn’t think about that! Good idea

recipes/output/dev-math.css
```diff
@@ -415,6 +415,12 @@
 @namespace xhtml "http://www.w3.org/1999/xhtml";
 @namespace cmlnle "http://katalysteducation.org/cmlnle/1.0";
 @namespace cxlxt "http://katalysteducation.org/cxlxt/1.0";
+:pass(0) [data-type="example"] .os-solution-container > [data-type="title"] {
+  move-to: trash; }
```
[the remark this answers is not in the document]
brittany-johnson: Ooooo I missed this one, thank you catching this! 
My next thought is to trash at pass 0 like I added in this PR and then keep the previous “trashing” that was on pass 10. What do you think about that? @wilkus27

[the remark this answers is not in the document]
wilkus27: Hmm my first idea was to change selector. I checked and in assembled.xhtml file (with markup from pass 0), title is direct child of `[data-type="solution"]`:

image

But you are right, it's safer to keep "trashing" in two passes, just in case :relaxed: Some of the elements could be used before they are trashed.

recipes/books/_generator.scss
wilkus27: Thank you for clarifying! :relaxed:

wilkus27 (approved):
Everything looks good! :)

Issue #2766, opened 2020-09-14
Images on top in solutions in the EoC


ajluk:
In HS Statistics there are images placed on top of the solution containers but mixin for such containers doesn't cover them (they are present in the EoC sections not EoB of the book): 
![image](https://user-images.githubusercontent.com/55203598/93096966-4b34a700-f6a5-11ea-8502-fb72070d26b6.png)

MartaSztuka:
## PASSED
### HS Statistics![Screenshot from 2020-09-23 16-23-42.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/d5388a60-d083-48e3-8a05-2bd7d3db35fe)

PR #2767, opened 2020-09-14
eoc solutions first elements

ajluk:
Issue: #2766


Stackblocks (approved):

recipes/mixins/_modify.scss
```diff
@@ -992,6 +992,22 @@
...
+                  class: attr(class) " first-inline-list-element";
+                } 
+              } @else { 
+                #{$element} {
+                class: attr(class) " first-element";
+              }
+            }
+          }
+        }
+      }
```
wilkus27, line 1010: Everything looks good :relaxed: I have one concern. This code is almost the same as part from the above. It differs only in a few places:

image

Is it possible to reduce these parts to avoid repetition?

recipes/mixins/_modify.scss
```diff
@@ -1038,6 +1054,29 @@
...
+              .first-inline-list-element {
+                string-set: hasFirstElement " has-first-inline-list-element";
+              }
+              &:deferred {
+                class: attr(class) string(hasFirstElement);
+              }
+            }
+          }
+      }
+    }
```
wilkus27, line 1079: Same for this part :slightly_smiling_face:

wilkus27 (changes requested):

ajluk:
For the problem with code refactoring in this PR will be opened a new issue card #2784

wilkus27 (approved):

Issue #2736, opened 2020-09-01
coreq math - token and list classes are not working in try it features


lwtchu:
related to #2732 

the classes are not working within the try its
![image](https://user-images.githubusercontent.com/11299125/91875624-59b0a680-ec41-11ea-9e0a-f1c01efe97f7.png)

made a content change on production to accomodate for publication but  this needs to fixed for other titles within the theme and for this title later

screenshot from 7.6 in college algebra coreq

lwtchu:
![Screen Shot 2020-09-01 at 2 45 14 PM](https://user-images.githubusercontent.com/11299125/91898721-c5a30700-ec61-11ea-9b2d-88ead7dd7f79.png)
still persists in pdf 389 - can do a lil content stuff to make it less noticeable. but should be fixed before we take this to print

ajluk:
@lwtchu I was trying to find the note from the screenshot you posted and couldn't do that. Natalia told me we don't have updated the content in chapter 7 in College Algebra Coreq on katalyst01. Only chapter 1 and 4 were updated. Could it be updated so we can checkk it out?

brittweinstein:
@lwtchu is there any update on this?

lwtchu:
@brittweinstein @ajluk oops i thought i updated the card here. but i migrated all the coreq content over last week.

ajluk:
@lwtchu Oh, Thank you!

ajluk:
@lwtchu This is how the lists look now: 
![image](https://user-images.githubusercontent.com/55203598/93339620-156dfa80-f82c-11ea-956c-9084613bad07.png)

lwtchu:
@ajluk yes this si fine thanks!

ajluk:
Hi @lwtchu Natalia has noticed that Try It notes have exercises inside of them and this one from above doesn't have it in the content, so the markup is different there. I didn't know about this, so I have added the styles for lists in the notes, but I think in this case a content change is needed because if the markup will be correct the lists will be styled without my changes using shape for exercises. 
The markup in the note from above: 
![image](https://user-images.githubusercontent.com/55203598/93461123-95f03200-f8e4-11ea-9de8-a63959b4ed38.png)
The correct markup from another Try It note:
![image](https://user-images.githubusercontent.com/55203598/93461350-eff0f780-f8e4-11ea-850b-84e99b2494ba.png)
Invalid markup in the assembled file of the note:
[...]

PR #2772, opened 2020-09-16
Style circled lists in Try it note

ajluk:
Issue: #2736


Stackblocks (approved):

wilkus27 (approved):
@ajluk everything looks fine :slightly_smiling_face: I'm just wondering if note from the issue has correct markup :thinking: In Prealgebra, for example, there is `[data-type="exercise"]` inside `os-note-body`. I think we should make sure if exercise tag is necessery here as well.

ajluk:
The issue linked with this PR has been resolved in content by Larissa.

Issue #2784, opened 2020-09-22
Mixins modify_addFirstElementClass() and modify_tagExerciseWithFirstElementClass() have some code repetitions


ajluk:
During working on issue  #2766 a problem with code refactoring has occurred. The goal we wanted to achieve was to make less code duplicates in the mixins(`modify_addFirstElementClass()`, `modify_tagExerciseWithFirstElementClass()`)It was described by Natalia in this comment:
https://github.com/openstax/cnx-recipes/pull/2767#discussion_r489411557
The main obstacles in doing this were the nested maps: `$Config_BookCompositePages`, `$Config_ChapterCompositePages` when we would like to add some conditionals and variables based on internal the map key `type`.

brittweinstein:
@brittany-johnson will review (15-30 min) this to assess when this may need to get done.

brittany-johnson:
@ajluk Hello! From a quick review, of the issue, I don't think that the time or effort needed for this issue is worth the payoff since we are prioritizing the recipe rewrite this upcoming quarter. 
BUT this is just my opinion. @ajluk If you feel differently please let me know!

ajluk:
@brittany-johnson I have missed your comment somehow in the inbox. I'm sorry for the late response. Yes, you're right. In the case of prioritizing the recipe rewriting I think it's not worth the effort.

PR #2787, opened 2020-09-22
Refactor mixins first elements

ajluk:
Issue: #2784

### DIFF
--- a/recipes/output/anatomy.css
+++ b/recipes/output/anatomy.css
@@ -2232,7 +2232,13 @@
 :pass(12) [data-type="exercise"]:not(.unnumbered) .os-problem-container > .os-figure:first-child {
   class: attr(class) " first-element"; }
 
+:pass(12) [data-type="solution"] .os-solution-container > .os-figure:first-child {
+  class: attr(class) " first-element"; }
+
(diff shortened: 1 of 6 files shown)


recipes/output/anatomy.css
```diff
@@ -2232,9 +2232,15 @@
...
 
+:pass(12) [data-type="solution"] .os-solution-container > .os-figure:first-child {
+  class: attr(class) " first-element"; }
+
 :pass(12) [data-type="exercise"]:not(.unnumbered) .os-problem-container > div:first-child > p:first-child img:first-child {
   class: attr(class) " first-element"; }
 
+:pass(12) [data-type="solution"] .os-solution-container > div:first-child > p:first-child img:first-child {
+  class: attr(class) " first-element"; }
+
```
ajluk, line 2243: I have tried to remove some duplicated code by using a `type` key form maps  `$Config_ChapterCompositePages` and 
`$Config_BookCompositePages` but in case of the second mixin `modify_tagExerciseWithFirstElementClass()` the code has been also applied to the books that don't have Answer Key like Anatomy as you can see above and AP Biology, Biology, HS Physics  and Polish Physics as you can see in the output css files below.

PR #2788, opened 2020-09-22
Refactor mixins first elements second attempt

ajluk:
Issue: #2784


ajluk:
When trying to implement a function solution I've got these kind of errors: 
image
which means I couldn't get the key within the function I have created.

PR #2792, opened 2020-09-22
Alpha-sort the lists in books.txt

Stackblocks:
Books.txt has gotten messy and while you _can_ find things via search it's also useful to be able to navigate by landmark. 

This also clears the path for us to sort future entries to prevent messiness down the road 👍


brittweinstein:
@Stackblocks please put points on this when you can!

ajluk (approved):

brittany-johnson (approved):

Issue #2486, opened 2020-07-09
Two columns in three columns layout


wilkus27:
In exercises modules (Section Exercises, EOC Exercises, Answer Key) with three columns layout sometimes appear only two columns:

![image](https://user-images.githubusercontent.com/40228854/87045146-d4d47e00-c1f7-11ea-93b7-c390f33721e0.png)

Screenshot is from **Prealgebra 2e**

brittweinstein:
This occurs in two spots. @lwtchu could you please assign priority to this?

brittweinstein:
Another thought that Alan had is maybe this could be something wonky within the content since it is so limited?

lwtchu:
![Screen Shot 2020-07-09 at 9 47 30 AM](https://user-images.githubusercontent.com/11299125/87054862-3af5dc80-c1c9-11ea-8739-4768821d2a50.png)
In the past the lists have been able to slim down to fix - is this a limitation with the new pipeline? or new designs?

lwtchu:
Update: Talked to BrittanyJ. She'll take a look. Might be something in the code

MartaSztuka:
@lwtchu Two columns in Answer Key occurs in other books too.
 
### Calcusus
![Screenshot from 2020-08-07 17-15-51](https://user-images.githubusercontent.com/47241114/89660468-aa96de80-d8d1-11ea-9b28-45b608cfb828.png)
### Precalculus
![Screenshot from 2020-08-07 17-17-07](https://user-images.githubusercontent.com/47241114/89660593-d87c2300-d8d1-11ea-941b-951a0d121c9e.png)
### College Algebra coreq
![Screenshot from 2020-08-07 17-17-39](https://user-images.githubusercontent.com/47241114/89660649-ec278980-d8d1-11ea-877a-5bccf599b4b8.png)

lwtchu:
is this because of spacing issues? I'm not sure how far you got with the investigation on this @brittany-johnson :eyes:

brittany-johnson:
Haven't looked at this in a few weeks ☠️
Once the fractions issue is resolved I can take a look again - hopefully sometime next week

lwtchu:
@brittany-johnson okay not super high priority - not a show stopper at this point for the coreqs

ajluk:
@lwtchu in Prealgebra 2e lists are placed next to the exercise number as they should be now. The new line has been removed from there in the content: 
[...]

PR #2794, opened 2020-09-23
Decrease table cell padding in the Answer Key

ajluk:
Issue: #2486
This PR only resolves the problem big tables inside forcing two columns layout.


Stackblocks (approved):

wilkus27 (approved):
I add approved but I think there is too many repetition. Maybe it will be better to create separate shape for tables in three columns layout. What do you think? We can create separate issue for that and discuss it in the future.

brittany-johnson:
I agree with Natalia, I think it would be nice to create a separate shape for this in the table shapes file. But this is something that can be done now or later.

brittany-johnson (approved):

Issue #2817, opened 2020-09-30
Business Ethics: style Learning Objectives


ajluk:
Documentation issue: #2711

MartaSztuka:
## PASSED
![Screenshot from 2020-10-05 13-04-32.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8ecce0dc-d0e0-4b7b-9557-fffb3473c679)

lwtchu:
@MartaSztuka can you see if the "learning objectives" is bolded? or is that how the font renders?

MartaSztuka:
@lwtchu Yes, it is bolded.
![Screenshot from 2020-10-05 16-01-05](https://user-images.githubusercontent.com/47241114/95089165-19eb4c00-0724-11eb-875b-0de9199281ed.png)

lwtchu:
There's some recent chatter on #2711 to add a font-weight: lighter on the text so it doesn't render bolded in the pdf. I'm going to reopen to make sure that gets captured.

brittweinstein:
@ajluk please see Larissa's note above.

MartaSztuka:
#### PASSED![Screenshot from 2020-10-21 17-37-56.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6f598c4c-40c1-40ee-93c8-f9cc0820dc9f)

PR #2818, opened 2020-09-30
style learning objectives

ajluk:
Issue:  #2817


styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -0,0 +1,7 @@
+$UnorderedList--FirstLevel: (
+    _name: "FirstLevelUnorderedList",
+    _subselector: ' > ul:not(.circled):not(.os-stepwise):not([data-labeled-item="true"])',
```
ajluk, line 3: I have left the :not() selectors to which were added in the corn theme to prevent this regular lists overwrite other types of lists in the books, but let me know if it's a good idea at the starting level.

Stackblocks (approved):

styles/designs/cardboard/parts/_lists-components.scss
wilkus27: For `[data-labeled-item="true"]` it's ok, because there are some settings for this list in `common-style` file. For circled and stepwise lists we should ask Larissa if they exists in books from these theme. But I think we can do it later, during work on lists.

wilkus27 (approved):

Issue #2822, opened 2020-10-01
Business Ethics: style blockquotes


ajluk:
Issue with documentation:  #2749

MartaSztuka:
###PASSED![Screenshot from 2020-10-06 12-43-28.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fdfa3c52-cd2c-4fd4-97ed-2e2411ebf75e)

PR #2823, opened 2020-10-01
style blockquotes business ethics

ajluk:
Issue:  #2822


styles/books/business-ethics/book.scss
```diff
@@ -4,6 +4,7 @@ $PagesWithBands: (
   (pageName: appendix, firstSelector: '.appendix', generalSelector: '.appendix'),
 );
 
+
```
Stackblocks, line 7: Any reason for the added line here?

styles/templates/template3/_design.scss
```diff
@@ -10,3 +10,6 @@
 @import 'designs/common/pdf/page';
 @import 'designs/common/reset';
 @import 'designs/common/pdf/footnotes-shapes.scss';
+
```
Stackblocks, line 13: If this is a gap for the purposes of separation could you include a comment about what differentiates the groups?

styles/templates/template3/_settings.scss
```diff
@@ -1,6 +1,13 @@
+//global template settings
 @import "colors";
 @import "typography";
 
+
+//template parts
+@import 'templates/template3/parts/link-settings';
+
```
Stackblocks, line 8: Extra spacing here as well

Stackblocks (changes requested):
Looks good aside from some spare newlines here and there :+1:

styles/output/business-ethics-pdf.css
```diff
@@ -688,4 +688,11 @@ sub {
   prince-image-resolution: auto, 200dpi;
   prince-background-image-resolution: 200dpi; }
 
+:root {
+  margin-left: 0.25inch;
+  margin-right: 0.25inch; }
```
wilkus27, line 693: Somethings went wrong :thinking: Settings are added to the root not to blockquote.

wilkus27 (changes requested):

styles/books/business-ethics/book.scss
ajluk: I think I have pasted the settings first here by accident and left the line empty after removing it. Thanks!

styles/templates/template3/_design.scss
ajluk: Yes, I just wanted to separate the shapes that are inside the `parts` folder from the others,  common/pdf shapes as we did for the previous themes.

styles/output/business-ethics-pdf.css
ajluk: @wilkus Thank you for catching it up!

Stackblocks (approved):

wilkus27 (approved):

Issue #2824, opened 2020-10-01
Business Ethics: style footnotes


ajluk:
Issue with documentation: #2747

ajluk:
@MartaSztuka >> Information for testing purposes: The footnotes can be easily found in Appendices.

MartaSztuka:
### PASSED
![Screenshot from 2020-10-12 13-55-34](https://user-images.githubusercontent.com/47241114/95743855-b2d51680-0c92-11eb-9773-ffcef96f1764.png)

PR #2830, opened 2020-10-02
style footnotes business ethics

ajluk:
Issue: #2824

- create mixin `modify_customNumberFootnotes` used to get the counter type set in `$Config_CustomNumberFootnotes` in the the config.scss file for books where a different counter type is needed 

- create a mixin `custom_numberFootnotes`  for custom numbering footnotes using counter type


ajluk:
The link color will be added within this PR: https://github.com/openstax/cnx-recipes/pull/2823. I will update my branch then.

recipes/books/_generator.scss
```diff
@@ -192,9 +192,13 @@
...
-    @include count_footnotes(footnotes);
-    @include count_footnoteLinks(footnoteLinks);
-    @include number_footnotes(footnotes, footnoteLinks);
+    @if $Config_CustomNumberFootnotes !=() {
+      @include modify_customNumberFootnotes($Config_CustomNumberFootnotes);
+    } @else {
+      @include count_footnotes(footnotes);
+      @include count_footnoteLinks(footnoteLinks);
+      @include number_footnotes(footnotes, footnoteLinks);
+    }
```
wilkus27, line 201: I would recommend to enable `move-footnotes` flag for Business Ethics. Then this part of code shouldn't be changed.

recipes/mixins/_count.scss
```diff
@@ -371,7 +371,7 @@
   @include _count(
     $name,
     'aside[role="doc-footnote"]',
-    '[data-type="chapter"], .appendix'
+    '[data-type="chapter"], .appendix',
```
wilkus27, line 374: Is this change necessary?

recipes/mixins/_number.scss
```diff
@@ -647,6 +647,17 @@
...
+@mixin custom_numberFootnotes($footnoteCounter, $footnoteLinkCounter, $counterType) {
+  a[role="doc-noteref"] {
+    content: counter($footnoteLinkCounter, $counterType);
+  }
+  aside[role="doc-footnote"]::before {
+    data-type: "footnote-number";
+    content: counter($footnoteCounter, $counterType);
+  }
+}
+
```
wilkus27, line 660: I think we should have just one config for custom footnotes. It was already created for footnotes in precalc, but I just noticed that we didn't modify it after tags for footnotes changed, so it doesn't work now and footnotes in precalc are not numbered properly :disappointed: 

In this situation, I think we should keep all mixins created by you in this PR and I will create separate issue for footnotes in Precalc. Then some of us will modify your mixins to add additional option needed in this book and remove old mixins. What do you think?

wilkus27 (changes requested):

recipes/mixins/_count.scss
ajluk: No, I have just left it accidentally. Already removed! :)

recipes/mixins/_number.scss
ajluk: Yes, it's a good idea. These changes could be rather time consuming.

wilkus27 (approved):

Stackblocks (approved):

Issue #2834, opened 2020-10-05
Business Ethics: style What Would note


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/2737 Option 3

MartaSztuka:
#### PASSED
![Screenshot from 2020-10-07 14-50-19.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5365bade-af87-4004-8169-735fada8f1e7)

PR #2835, opened 2020-10-05
Style What Would note

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2834


Stackblocks (approved):

styles/templates/template3/_colors.scss
```diff
@@ -36,6 +36,9 @@
     option2HeaderBackgroundColor: (_ref: "scheme:::B4"),
     option2BodyBackgroundColor: (_ref: "scheme:::B6"),
     option2HeaderColor: (_ref: "scheme:::white"),
+    option3HeaderBackgroundColor: (_ref: "scheme:::O2"),
+    option3BodyBackgroundColor: (_ref: "scheme:::B5"),
+    option3HeaderColor: (_ref: "scheme:::black"),
```
ajluk, line 41: @wilkus27 I guess a more suggestive name for the colors used by notes could be better: f.ex. `noteOption3Color`, what do you think?

ajluk (commented):
Everything looks good! I have only left a suggestion about color names.

styles/templates/template3/_colors.scss
wilkus27: @ajluk Yes, I was also thinking about it, but coudn't decide what be better :smile: Because then names are very long, eg. `noteOption3HeaderBackgroundColor`. On the other hand name with `note` inside is more descriptive. Let me change it :relaxed:

Issue #2851, opened 2020-10-07
Business Ethics: Style Introduction Page


ajluk:
Issues with documentation: #2642, #2655

lwtchu:
### BLEEDS PASSED
![Screen Shot 2020-10-16 at 12 31 07 PM](https://user-images.githubusercontent.com/11299125/96289966-7dd70580-0fab-11eb-936f-aff34ee638cd.png)

MartaSztuka:
## PASSED
![Screenshot from 2020-10-27 14-32-16.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/9af17983-a3ab-47fb-aaa6-565947ddb6f6)

PR #2859, opened 2020-10-11
Chapter intro business ethics

ajluk:
Issue: #2851


Stackblocks (approved):

styles/designs/cardboard/parts/_titles-components.scss
```diff
@@ -28,22 +38,40 @@ $H1--Numbered: (
...
+    flex-direction: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    vertical-align: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    z-index: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    margin-right: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    padding-right: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    margin-left: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    padding-left: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    width: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    left: enum('ValueSet:::OPTIONAL'), //used in Chapters
+    margin-top: enum('ValueSet:::OPTIONAL'), //used in Chapters
```
wilkus27, line 62: There is so many OPTIONALs here. Based on the our work on previous themes I have a feeling it's very confusing. When we want to reuse some component later it's very hard to find out which property should be used in our case. What do you think to create separate components if we need to change a lot of properties? In this case I would create `H1--ChapterIntro`, `H1--ChapterIntro__Text` etc. and apply as much hard coded values as possible. It's just my point of view. :wink:   I'm really curious what is your opinion. Maybe you like more to have less amount of components which contain REQUIRED and OPTIONAL properties. Let me know what do you think :slightly_smiling_face:

styles/designs/cardboard/parts/_chapterintro-components.scss
```diff
@@ -0,0 +1,150 @@
...
+    )
+);
+
+$ChapterIntro__Content: empty_wrapper(IntroContent, ' > .intro-text');
+
+$ChapterIntro__Content__Header: (
+    _name: "IntroContentHeader",
+    _subselector: ' > h2[data-type="document-title"]:first-of-type', 
+    _properties: (
+        font-size: enum('ValueSet:::REQUIRED'),
```
wilkus27, line 61: Can we hard code it?

wilkus27 (changes requested):

styles/designs/cardboard/parts/_titles-components.scss
ajluk: Yes, it is a good idea to create new component for such cases. :slightly_smiling_face:

wilkus27 (changes requested):

wilkus27 (approved):
Everything looks good! :relaxed: A lot of work is done here. Good job! :+1:

Issue #2890, opened 2020-10-23
Business Law: style no class tables


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/2741

MartaSztuka:
### PASSED
![Screenshot from 2020-11-02 09-38-57.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c7f4056f-f48c-40a1-8e0c-b9b68371aeb9)

Issue #2741, opened 2020-09-02
cardboard theme: tables


lwtchu:
In the original issue (https://github.com/openstax/oer.exports/issues/3399) it goes in depth into how the `top-titled` class works styling. I believe most of the work is done on the recipe side (#1241) and we just need to style it (hopefully my understanding is correct!). 

There are two types: regular and `text-heavy`. Either can support the `top-titled` class where it takes the first header row and creates no borders/no boxes, centers the text and makes it bold.

# Type 1 Regular Table
![Screen Shot 2020-09-02 at 5 49 46 PM](https://user-images.githubusercontent.com/11299125/92044854-bea0f500-ed44-11ea-8e26-4640829145b0.png)
![Screen Shot 2020-09-02 at 5 49 40 PM](https://user-images.githubusercontent.com/11299125/92044856-bfd22200-ed44-11ea-84cb-52d83d03d689.png)

## First header row 
cell color: B3: `#0DC0DC`
font: noto sans, base font size
font treatment: bold
font color: B2: `#344456`
alignment: content driven

## Second header row (if present)
cell color: B4: `#24739E`
font: noto sans, base font size
font color: white `#FFFFFF`
alignment: content driven

## Non header rows
font: noto sans, base font size
font color: black `#000000`
alignment: content driven

## ALL Cell borders
Cell border color: G2: `#F2F2F2`
Cell border width: 1px 
Applies to border-left, right, top, bottom

## Cell corners
please make them all right angles contrary to what exists in the pdf

# Type 2 `text-heavy` Table
![image](https://user-images.githubusercontent.com/11299125/92045018-25bea980-ed45-11ea-9456-e5ec5dcdd968.png)

## all header rows 
cell color: G2: `#F2F2F2`
font style: noto sans
font treatment: bold
font color: B2: `#344456`
cell alignment: content driven

## Non header rows
font: noto sans, base font size
font color: black `#000000`
alignment: content driven

## Cell border
BOTTOM ONLY FOR HEADER
color: B5: `#D6D9DD`
border width: 1px

## Cell corners
please make them all right angles contrary to what exists in the pdf

# Alternate color schema for table 1 (default table)
For _Astronomy_
![Screen Shot 2020-10-26 at 12 43 18 PM](https://user-images.githubusercontent.com/11299125/97208232-e59a0700-1788-11eb-804b-0c1b2ada1ada.png)

Table body and cell borders should be the same. Only difference is the colors in the header row

## Header row
[...]

PR #2891, opened 2020-10-23
style no class tables business law

ajluk:
Issue: #2890


styles/templates/template3/parts/_table-settings.scss
```diff
@@ -37,5 +37,44 @@
...
+            margin-right: h-spacing(1),
+        ),
+        CaptionTitle: (
+            font-family: (_ref: "typography:::baseFont"),
+            color: (_ref: "colorMap:::captionTextColor"),
+        ),
+        CaptionText:(
+            font-family: (_ref: "typography:::baseFont"),
+            color: (_ref: "colorMap:::captionTextColor"),
+        ),
```
ajluk, line 78: I have just added different selector and settings for these tables and they will be using the same shapoe as Fake Title Tables, because the only difference is that they don't have the title present in the FakeTitleTable shape. But let me know if these tables should rather have separate shape, because the markup is a little different.

wilkus27: Hmm :thinking: Based on the specs from https://github.com/openstax/cnx-recipes/issues/2741 there are actually two types of tables: with and without borders. So maybe we can name shapes based on this property? For example `TablesWithBordersShape` and `TablesWithoutBordersShape`. Then `top-titled` tables and `no class` tables will use the first one and `text-heavy` tables the second one. What do you think? cc @Stackblocks

wilkus27 (changes requested):

styles/templates/template3/parts/_table-settings.scss
Stackblocks: I like that. It's descriptive of the specific use-cases and doesn't lose too much versatility since all of the tables fall into one of those two groups anyways. ✅

Stackblocks (approved):

styles/templates/template3/parts/_table-settings.scss
ajluk: Yes. That's a good idea! :smile:

wilkus27 (approved):

brittweinstein:
@ajluk we can merge

Issue #2910, opened 2020-10-27
Introduction to Business: implement Splash image, Chapter Outline, Introduction


ajluk:
Issues with documentation: https://github.com/openstax/cnx-recipes/issues/2655 , https://github.com/openstax/cnx-recipes/issues/2642

ajluk:
Screenshot for refenrence: 
![image](https://user-images.githubusercontent.com/55203598/97451210-d0151080-1933-11eb-93c1-d0adef404b5e.png)
Learning outcomes will be styled within another PR.

MartaSztuka:
### PASSES![Screenshot from 2020-11-02 11-23-09.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/6feb4234-3832-4dda-9283-520659947ca2)

PR #2914, opened 2020-10-28
Implement splash figure chapter outline

ajluk:
Issue:  #2910


ajluk:
@wilkus27 I have based my PR on yours with styling figures. 
Within this PR I needed to change the settings for figure caption in Chapter Intro, because I have noticed that if the caption text is short (there are short captions in Intro to Business) it is moved to the right to much (because of centering), so I have made the centering propertis OPTIONAL for the container as we previously did in corn and carnival themes. 
image
image

Stackblocks (approved):

wilkus27 (approved):

brittweinstein:
@ajluk we can merge

PR #2918, opened 2020-10-28
Fix AddMan Exercise Selectors

Stackblocks:
I missed these when initially setting up AddMan, so the build-styles script currently breaks when run. This fix should resolve that.


ajluk (changes requested):
@Stackblocks The styles should be regenerated as well, because the selectors weren't added in the `pdf-css` file before. Now the PR have no conflicts, so you should do it before merging.

Stackblocks:
@ajluk I ran `./script/run ./script/build-styles` and there were no changes as a result. Am I missing another component of this construction outside of the `_exercises-settings.scss` file?

brittany-johnson:
Try running `PLATFORM="pdf" node ./styles/build/build.js  styles/books/additive-manufacturing/book.scss ./styles/output/additive-manufacturing-pdf.css`. That generated an output for me

brittany-johnson (changes requested):
left comment

Stackblocks:
Using the alternative script worked! I've pushed a second commit containing the updated style files.

ajluk:
@Stackblocks I'm not sure why the general script wouldn't  work for this book on your machine locally. I have never met such errors, but  I'm glad the other script for one book worked! Everything looks good!

ajluk (approved):

brittany-johnson (approved):

PR #3049, opened 2020-11-12
add principles management to template

ajluk:
This Pr only adds Principles of management to template.


styles/books/pl-u-physics/book.scss
```diff
@@ -9,9 +9,9 @@ $PagesWithBands: (
   $ChapterIntroType: fullWidth;
   
   @import 'framework/framework';
-  @import 'templates/template1/design';
-  @import '../../designs/carnival/pdf/locale/pl/folio';
-  @import 'templates/template1/settings';
+  @import 'design-settings/carnival/design';
+  @import 'design-settings/carnival/settings';
+  @import '../../designs/carnival/pdf/locale/pl/folio';
```
ajluk, line 15: @brittany-johnson I have updated the branch with `new-x.x` just to check if it resolve the problem with not passing checks and there were some paths that needed to be fixed after your changes, so I have done it within this PR.

Stackblocks (approved):

styles/output/intr-business-pdf.css
```diff
@@ -0,0 +1,1971 @@
+@charset "UTF-8";
```
brittany-johnson, line 1: I noticed that two of the files for intro-business are spelled `intr-business`

brittany-johnson (changes requested):
Just have one comment, other than that this is a great solution, good job!

styles/output/intr-business-pdf.css
ajluk: Thank you for noticing! I missed it somehow.

brittany-johnson (approved):
Yay!!

Stackblocks (approved):

Issue #2943, opened 2020-11-05
Accounting: Concepts In Practice


brittweinstein:
`<note class="concepts-practice">` - uses https://github.com/openstax/cnx-recipes/issues/2737 with color option 2

kswanson33:
https://github.com/openstax/cnx-recipes/pull/3060

kswanson33:
This content can be fetched from katalyst01.cnx.org, with col25448. Specs: [Color option 2](https://github.com/openstax/cnx-recipes/issues/2737). 

Screenshot of applied changes:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/01ff75ad-6604-4669-a4b6-bf2f114093fb)

meeragovindaraju:
Concepts in Practice looks good 

![Screen Shot 2020-11-18 at 12 15 23 PM](https://user-images.githubusercontent.com/33066187/99573511-cfecbb80-299b-11eb-8f53-cccdb6a6fbf8.png)

kswanson33:
https://github.com/openstax/cnx-recipes/issues/2944 Thread where we discuss the letter spacing issue in the title

PR #3060, opened 2020-11-16
implement styling for concepts-practice elements

kswanson33:
Issue #2943 

image


styles/designs/cardboard/parts/notes/_notes-components.scss
```diff
@@ -15,6 +15,7 @@ $Notes--Boxed__Title: (
         font-family: enum('ValueSet:::REQUIRED'),
         font-size: font-scale(1),
         line-height: line-height(1),
+        letter-spacing: 3px,
```
kswanson33, line 18: I should point out that I used a static value here, rather than something like letter-spacing(1). Is there a better way of doing this?

kswanson33: I think this is causing the build to fail as well!

brittany-johnson: Yes, it looks like the build is failing due because px is being used instead of rem. 

I forgot to mention before that we have a `_grid.scss` file where we keep functions used for spacing. The most frequently used ones are probably `v-spacing` for vertical spacing and `h-spacing` for horizontal. 
We try to use the functions for spacing whenever possible. If a value needs to be super specific, like -5.654rem (or something like that), then its fine to go off the grid since the current functions won't return that value

cc: @LostMojo

styles/books/accounting/book.scss
```diff
@@ -0,0 +1,43 @@
...
+@import '../../design-settings/cardboard/_design.scss';
+@import '../../design-settings/cardboard/_settings.scss';
+@import '../../designs/cardboard/pdf/folio';
+
+@include add_settings((
+    // BoxedNoteWithSubtitleShape
+    ConceptsPractice: (
+        _selectors: ('.concepts-practice'),
+        // Notes--Boxed__Title
+        Title: (
```
brittany-johnson, line 13: When writing settings in the `book.scss` file we historically opt for the 'one-line' style to save space and make it easier to read. For example, `Title` would be written like this: 

`'Title:::color': (_ref: "colorMap:::noteOption2HeaderColor"),`

Also, be sure to use the color map for colors. 

cc: @LostMojo

brittany-johnson (approved):
Left a couple of comments, otherwise, looking great!

brittany-johnson (changes requested):
Oops didn't mean to approve lol

brittany-johnson (approved):
Woo!!!

Stackblocks (approved):

Issue #2892, opened 2020-10-23
accounting - red text


lwtchu:
![Screen Shot 2020-10-23 at 11 20 55 AM](https://user-images.githubusercontent.com/11299125/97028498-d6fde680-1521-11eb-8631-155af6aa54eb.png)

- targeted by `<emphasis class="colored-text">`
- apply the color: R1: `#CC0000`
- remove the bold that is default to `<emphasis>`

kswanson33:
PR: https://github.com/openstax/cnx-recipes/pull/3072

Affects accounting; found on katalyst01 with col25448.

All text targeted by `.colored-text` should be colored with `#CC0000`, and have `font-weight: normal`.

Example of applied changes:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/ea2766dd-c5e7-48a8-91fe-9cfaa9005522)

meeragovindaraju:
**Test Passed**

<img width="821" alt="Screen Shot 2020-12-02 at 8 51 06 AM" src="https://user-images.githubusercontent.com/33066187/100888439-bd827f80-347b-11eb-94a3-43fca2b9b61b.png">


<img width="776" alt="Screen Shot 2020-12-02 at 8 52 09 AM" src="https://user-images.githubusercontent.com/33066187/100888459-c2473380-347b-11eb-97a7-9b2eab9167ce.png">

PR #3072, opened 2020-11-19
Implement colored text emphasis in R1 for accounting

kswanson33:
#2892 
image


styles/design-settings/cardboard/_colors.scss
```diff
@@ -77,5 +78,6 @@
     indexGroupLabelColor: (_ref: "scheme:::B2"),
     tableHeaderColor: (_ref: "scheme:::white"),
     tableHeaderBgColor: (_ref: "scheme:::B4"),
+    emTextColor: (_ref: "scheme:::R1"),
```
kswanson33, line 81: Is this a good name for this color?

styles/designs/cardboard/parts/_emtext-components.scss
```diff
@@ -0,0 +1,8 @@
+$ColoredText__Container: (
```
kswanson33, line 1: Is it appropriate to create a new component & shape for emphasizing text, or is there an existing shape & component file that I could add these changes to instead?

styles/design-settings/cardboard/_colors.scss
LostMojo: It seems to match the naming conventions in the other styles-- Carnival's reds are also named in the R1, R2 pattern.

LostMojo: Oh!! But should we spell out Emphasis? The map names seem to weigh clarity over briefness, so it might be worth it to spell things out.

styles/books/accounting/book.scss
```diff
@@ -19,4 +19,12 @@ $PagesWithBands: ();
     )
 ));
 
-@include use('ConceptsPractice', 'cardboard:::BoxedNoteWithSubtitleShape');
\ No newline at end of file
+@include add_settings((
+    EmText: (
```
LostMojo, line 23: I think EmText could be a little more clear about its purpose. It might be worth refactoring it to something like "EmphasizedText" or even "EmphasizedColoredText"?

LostMojo (changes requested):
It looks good to me! However, I do think it would be worth it to refactor EmText to something a little more descriptive & specific. "Em" isn't immediately apparent, and the shape is being applied to text that's emphasized _and_ colored.

styles/designs/cardboard/parts/_emtext-components.scss
brittany-johnson: Good question! I think the best place to put it now would be in `common-components` and include it in the book root shape since this `colored-text` class could be anywhere from my understanding

styles/books/accounting/book.scss
brittany-johnson: Yes I agree! EmphasizedColoredText would be a good name

brittany-johnson (changes requested):
Left a comment ~

styles/books/accounting/book.scss
kswanson33: Resolved!

styles/designs/cardboard/parts/_emtext-components.scss
kswanson33: Resolved!

LostMojo (approved):

brittany-johnson (approved):
Looks good!

Issue #2934, opened 2020-11-05
Accounting: Table of Contents


brittweinstein:
Should be in line with: https://github.com/openstax/cnx-recipes/issues/1740

kswanson33:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/59f4c5f5-4454-4c68-8ba1-59325a2c5e8a)

After applying all styles, the TOC looks like this. I have checked intro-business, which has a TOC that's been approved, and determined that intro-business is inheriting some styles from the root which are essential to making the TOC look appropriately styled. 

@lwtchu What styles should be applied at the root of accounting? Is there a card for this?

lwtchu:
@brittany-johnson can you advise on the root thing? i'm not sure how the other titles are being styled :eyes:

kswanson33:
@lwtchu @brittany-johnson I have now determined that a one-line addition of an import from common fixes this issue. It does also add additional styles throughout the book. I think this is probably what we want, but it seems to be outside the scope of this card! Thoughts?

lwtchu:
@kswanson33 i think it's up to you ... i don't have strong feelings about it but it might be helpful for new folks?? is this something that needs to happen for every new book..like there's theme level styles to import?

kswanson33:
@lwtchu I'll follow up with my team about this!

kswanson33:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/9d660f14-4077-4ec9-b1a1-5c2379910065)

@lwtchu Progress so far^ 

The only thing that doesn't match the specs at this point is that the chapter numbers (1.1, 2.3, etc) aren't bold (they are bold in the specs). I've looked into it, and previously approved ToCs ([intro-business](https://github.com/openstax/cnx-recipes/issues/2897) and [business-ethics](https://github.com/openstax/cnx-recipes/issues/2803)) are also missing that style. 

Option 1: I could match these books by skipping the bold
Option 2: I could add the bold styling to just the accounting book
Option 3: I could add the bold styling to the cardboard theme, which would affect all books.

What do you think is best?

brittany-johnson:
Hello! @kswanson33 I just re-opened and assigned you to https://github.com/openstax/cnx-recipes/issues/2897
While you're implementing the design for Accounting, can you make changes to the TOC components and shapes so that the comment I left in #2897 is addressed in both books? 

Thank you for pointing out the missing bold text!

[...]

Issue #2897, opened 2020-10-26
Introduction to Business: style Table of Contents


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/1740 - version blue

MartaSztuka:
### PASSED
![Screenshot from 2020-10-30 11-45-08.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/98fd140f-f04c-4309-9657-6940c2e5bd2b)

brittany-johnson:
The final design is missing some specs - the grey background color should be on odd number chapters, not even and the module numbers should be bolded. Reopening

kswanson33:
[Link to PR](https://github.com/openstax/cnx-recipes/pull/3075)

kswanson33:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/ad628869-28d1-43a8-a925-7d096c4c151b)

kswanson33:
[Pull request](https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/3075)

intro-business is on katalyst01 with col25734.

Applied changes affect only the table of contents:

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/924dc3ef-1fb4-460a-b9f6-342075dd5314)

meeragovindaraju:
**Test Passed**

![Screen Shot 2020-12-07 at 6 41 28 PM](https://user-images.githubusercontent.com/33066187/101422283-e95a9680-38bb-11eb-86c1-edfd7c148c4b.png)

PR #3075, opened 2020-11-20
Fixes the cardboard ToC by adding missing styles; implements ToC for Accounting

kswanson33:
#2934 #2897 
image


styles/designs/cardboard/parts/_toc-components.scss
```diff
@@ -163,7 +171,7 @@ $Toc__Link--MiddleLevel__Page: (
 
 $Toc__Chapter--OddItem: (
   _name: 'ChapterOddItem',
-  _subselector: ' > li.os-toc-chapter:nth-child(2n+1)',
+  _subselector: ' > li.os-toc-chapter:nth-child(2n)',
```
LostMojo, line 174: Does this still select the odd items? It looks like it's the even items that have backgrounds.

LostMojo (approved):
Looks good!

styles/designs/cardboard/parts/_toc-components.scss
kswanson33: Counterintuitively, yes -- since the first item in this list (in all books I looked at) is the preface, odd numbered chapters are actually even list items. 

But this is a good issue to raise: if a book doesn't have a preface, the above will stop working. Is this something to worry about?

brittany-johnson (changes requested):
I haven't looked at your code yet but I noticed in the screenshot that the page numbering is now off in the TOC. [Here is a PDF on production for reference ](https://d3bxy9euw4e147.cloudfront.net/oscms-prodcms/media/documents/FinancialAccounting-OP_YioY6nY.pdf)

The text is also very close to the right border of the darker gray background. 

Let me know if you have any questions!

kswanson33:
image

Addresses the padding issue so that padding resembles the PDF on prod. Specs added to Top Level List Item:
- `padding-top: v-spacing(2),`
- `padding-bottom: v-spacing(2),`
- `padding-right: h-spacing(5),`
- `padding-left: h-spacing(1),`

kswanson33:
image

For the page numbering to start at page 1 for the Preface, `class="preface"` must be added to the page containing the preface. (Sample image is from me making this change locally -- the problem still needs to be fixed for the baked XHTML.)

kswanson33:
image

ToC for intro-business

Stackblocks (approved):

Issue #2942, opened 2020-11-05
Accounting: Think It Through


brittweinstein:
`<note class="think-through">` - uses https://github.com/openstax/cnx-recipes/issues/2737 with color option 2

LostMojo:
40 Test Content Locations
collection ID: col25448
server: katalyst01.cnx.org
chapter/section: Sprinkled throughout

![Screen Shot 2020-12-01 at 11.42.08 AM.png](https://images.zenhubusercontent.com/5fa9a1a8a99686b64c824016/9013a79c-7985-419d-bd5b-3d0f035196c8)

meeragovindaraju:
**Test Passed** But the PDF had only 4 'Think it through' sections. The PDF did not have all the content. 

![Screen Shot 2020-12-03 at 4 53 25 PM](https://user-images.githubusercontent.com/33066187/101098363-6ae2b980-3588-11eb-8c67-336a7b5f3330.png)

[collection.pdf](https://github.com/openstax/cnx-recipes/files/5639461/collection.pdf)

PR #3112, opened 2020-12-01
adds Think It Through Styling

LostMojo:
#2942 
Screen Shot 2020-12-01 at 11 42 08 AM


styles/books/accounting/book.scss
```diff
@@ -27,6 +27,16 @@ $PagesWithBands: ();
...
+    ),
+    ThinkItThrough: (
+        _selectors: ('.think-through'),
+        "Title:::color": (_ref: "colorMap:::noteOption2HeaderColor"),
+        "Title:::font-family": ( _ref: "typography:::titleFont" ),
+        "Title:::background-color": ( _ref: "colorMap:::noteOption2HeaderBackgroundColor" ),
+        "Body:::font-family": ( _ref: "typography:::baseFont" ),
+        "Body:::background-color": ( _ref: "colorMap:::noteOption2BodyBackgroundColor" ),
+        "Subtitle:::color": ( _ref: "colorMap:::noteSubtitleColor" ),
+        "Subtitle:::font-family": ( _ref: "typography:::titleFont" ),
```
kswanson33, line 39: Same thought as in code review yesterday -- could the ThinkItThrough and ConceptsPractice become one settings block, with a comment to indicate what types of note they affect? Looks like they use the same settings.

styles/books/accounting/book.scss
```diff
@@ -45,3 +55,4 @@ $PagesWithBands: ();
 //Notes
 @include use('ConceptsPractice', 'cardboard:::BoxedNoteWithSubtitleShape');
 @include use('YourTurn', 'cardboard:::BoxedNoteWithSubtitleShape');
+@include use('ThinkItThrough', 'cardboard:::BoxedNoteWithSubtitleShape');
```
kswanson33, line 58: This isn't in the scope of this card, but I happened to notice yesterday in code review that other books import `@include use('NotesLists', 'cardboard:::NotesListsShape');`. This import seems to align list items correctly within the note box (there's a couple of Your Turn notes in chapter 3 with lists). Since I don't think there's specifically a card for that import, I thought this might be a good PR to sneak that in!

kswanson33 (approved):
Left a few thoughts, but looks good overall!

styles/books/accounting/book.scss
ajluk: @kswanson33 Yes, you're right. We have created separate shape for lists that are placed in notes, so it's a good practice to include this shape whenever there are lists present within a note that someone is working on and there some additional changes needed for the lists inside note we're adding it there as well.

styles/books/accounting/book.scss
ajluk: @kswanson33 It is a good thought, but the notes have different selectors and if we include two or more selectors together in one settings it cause problems sometimes with creating next selectors within a shape, that's why we're adding them separately.

ajluk (changes requested):
@LostMojo Everything looks good, but could you include the Note Lists Shape within this PR, because notes that have lists inside are already styled and I'm afraid it can be missed later?

ajluk (approved):

Issue #2948, opened 2020-11-05
Accounting: Continuing Application at Work


brittweinstein:
` <note class="continuing-application">`- uses https://github.com/openstax/cnx-recipes/issues/2737 with color option 2

kswanson33:
Specs #2737

Changes applied to Continuing Application at Work note in accounting

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/81fd7a25-47f6-4385-9e10-e5fcb520749f)

[Link to PR](https://github.com/openstax/cnx-recipes/pull/3119)

accounting is on katalyst01 with col25448

meeragovindaraju:
**Test Passed**

![Screen Shot 2020-12-09 at 5 39 18 PM](https://user-images.githubusercontent.com/33066187/101701946-85fc7000-3a45-11eb-911d-74d457554932.png)

Issue #2947, opened 2020-11-05
Accounting: Link to Learning


brittweinstein:
`<note class="link-to-learning"> `- uses #2737 with color option 1

kswanson33:
Specs #2737

Changes applied to Link to Learning note in accounting

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/8da09257-0b2b-41c4-a16a-7ea86a94c17e)

[Link to PR](https://github.com/openstax/cnx-recipes/pull/3119)

accounting is on katalyst01 with col25448

meeragovindaraju:
**Test Passed**

![Screen Shot 2020-12-09 at 5 40 30 PM](https://user-images.githubusercontent.com/33066187/101702026-af1d0080-3a45-11eb-8830-09731a9e09fd.png)

Issue #2946, opened 2020-11-05
Accounting: Ethical Considerations


brittweinstein:
`<note class="ethical-considerations"> `- uses #2737 with color option 2

kswanson33:
Specs #2737

Changes applied to Ethical Considerations note in accounting

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/8f3d1460-385f-4408-af06-1e700f9340f1)

[Link to PR](https://github.com/openstax/cnx-recipes/pull/3119)

accounting is on katalyst01 with col25448

meeragovindaraju:
**Test Passed**

![Screen Shot 2020-12-09 at 5 41 36 PM](https://user-images.githubusercontent.com/33066187/101702098-d4aa0a00-3a45-11eb-9d9e-c3d6eb69821b.png)

Issue #2945, opened 2020-11-05
Accounting: IFRS Connection


brittweinstein:
`<note class="ifrs-connection">` - uses https://github.com/openstax/cnx-recipes/issues/2737 with color option 2

kswanson33:
Specs #2737

Changes applied to IFRS Connection note

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/d2a6a62e-2834-4b17-9e6d-0476f0aeef9e)

[Link to PR](https://github.com/openstax/cnx-recipes/pull/3119)

accounting is on katalyst01 with col25448

meeragovindaraju:
**Test passed**

![Screen Shot 2020-12-09 at 5 44 41 PM](https://user-images.githubusercontent.com/33066187/101702332-42eecc80-3a46-11eb-8f66-e153759ef263.png)

PR #3119, opened 2020-12-03
style accounting notes

kswanson33:
#2946 #2948 #2947 #2945

image

image

image

image


styles/books/accounting/book.scss
```diff
@@ -8,7 +8,7 @@ $PagesWithBands: ();
 //Notes
 @include add_settings ((
     // BoxedNoteWithSubtitleShape
-    YourTurn: (
+    Opt3Note: (
```
ajluk, line 11: @kswanson33 The idea to use the number of styling option from the specs in the note name is really good! But if we would like to change the naming convention for the notes, all of the notes names should be changed as well for other books. There are also some notes with icon in this theme, so then the name could be `Opt3WithIcon` for example. @brittany-johnson @wilkus27 @Stackblocks @LostMojo what do you thin of this idea?

ajluk (commented):
Everything looks good! I have just left a comment about changing the note naming convention and I'm curious what the rest of the team think about it.

styles/books/accounting/book.scss
ajluk: I'm only concerned what the name should look like in the case when more that one note uses the same option styling number (like in Principles of Management). Maybe we could use both note and option name f.ex: `YourTurnOpt3`.

LostMojo: Well, if we're hoping to make this an overarching style goal, it would be good to have specific categories (like Opt3NoteWithIcon) but making the name more specific (like YourTurnOpt3) would undermine making the categories broadly applicable.

Stackblocks: The overlap of more than one note using the styling number can be helped out by having comments listing the notes that use the option above each one, like 

```
//Used in: Your Turn, Note B, Note C
NoteOption3Icon
...
```

Like Mary Jo said, we don't want to introduce new points that require unique code chunks.

This may be something to set out in a separate card like converting everything to BEM was, just so the fixes in this card aren't held up.

ajluk: @Stackblocks I meant that when there will be more than one note using the same styling option in the same book, the name used in the settings for these notes will be the same, but the notes have different selectors so the name must be different too for each one of them. It can be mixed with note title and styling option like I and MaryJo have written above f.ex.:`YourTurnOpt3`. If we decide the change the naming convention for the notes in cardboard I guess MaryJo can create a card to change it in every book in the future during the cleaning.

wilkus27: First I was very enthusiastic with conception that @kswanson33 applied. It avoids a lots of repetitions. But after discusion with @ajluk I'm not sure if it's a good idea. As Anna noticed, in some books we style notes with some option but they have additional icon, different for each note (each selector). In this case we have to create separate settings. Here is the example: https://github.com/openstax/cnx-recipes/issues/2737#issuecomment-730379256. In this case, in my opinion, it will be better to use old conception for consistency, even there are repetitions. It might be confusing in  the future why some notes use `Opt3Note` settings and another self named settings even they are styled like notes from Option 3.

wilkus27 (changes requested):

Stackblocks:
For the time being we should move forward with the old-style to get this card moving again, and we can include the ideas being discussed here in the to-do list for problems to work out as we continue improving our code base and pipelines.

kswanson33:
Sounds good. Thanks for all the input!

styles/books/accounting/book.scss
```diff
@@ -18,15 +18,55 @@ $PagesWithBands: ();
         'Subtitle:::color': (_ref: "colorMap:::noteSubtitleColor"),
         'Subtitle:::font-family': (_ref: "typography:::titleFont"),
     ),
+    // Option 2 Notes
     ConceptsPractice: (
-        _selectors: ('.concepts-practice'),
+        _selectors: ('.concepts-practice', '.ifrs-connection', '.continuing-application', '.ethical-considerations'),
```
wilkus27, line 23: There should be only selector for Concept Practice note. Same for notes from below :wink:

styles/books/accounting/book.scss
```diff
@@ -18,15 +18,55 @@ $PagesWithBands: ();
...
         "Title:::color": (_ref: "colorMap:::noteOption2HeaderColor"),
         "Title:::font-family": ( _ref: "typography:::titleFont" ),
         "Title:::background-color": ( _ref: "colorMap:::noteOption2HeaderBackgroundColor" ),
         "Body:::font-family": ( _ref: "typography:::baseFont" ),
         "Body:::background-color": ( _ref: "colorMap:::noteOption2BodyBackgroundColor" ),
         "Subtitle:::color": ( _ref: "colorMap:::noteSubtitleColor" ),
         "Subtitle:::font-family": ( _ref: "typography:::titleFont" ),
+    ),
+    IFRS: (
```
wilkus27, line 32: We are using Camel Case for naming. So it should be `Ifrs`. I would also reccomend to name it similar to note class, so maybe `IfrsConnection`.

wilkus27 (changes requested):

wilkus27 (approved):

Stackblocks (approved):

Issue #2893, opened 2020-10-23
accounting - eoc problem sets


lwtchu:
Accounting has 7 question sections in the EOC.  Not every chapter will have all of them.

### Multiple Choice `<section class="multiple-choice">`
![Screen Shot 2020-10-23 at 5 12 33 PM](https://user-images.githubusercontent.com/11299125/97058301-05e08080-1553-11eb-93be-9dfd6f8222f2.png)

- standard multiple choice exercise 
- exercise number is bolded with period
- the LO number and icon should come from #2889 
- spacing between each exercise is the same as business ethics: 0.7rem
- numbering is 1, 2, 3, etc and continuous for the chapter
- icon: https://drive.google.com/file/d/1Oz3Ej81o-JVLC75amloQ3m-YPDH-hoCN/view?usp=sharing

### Questions `<section class="questions">`
![Screen Shot 2020-10-23 at 5 21 08 PM](https://user-images.githubusercontent.com/11299125/97058735-3e348e80-1554-11eb-8817-d10ba7210218.png)
- icon: https://drive.google.com/file/d/1-IA4Gbm1OlPC2bg8lSedIUUIrjvZdBrB/view?usp=sharing
- exercise number is bolded with period
- LO icon ssame as bove
- spacing between each exercise is 0.7rem
- number is 1, 2, 3, etc and continuous for the chapter

### Exercise Set A / Exercise Set B `<section class="exercise-set-a">` `<section class="exercise-set-b">`
![Screen Shot 2020-10-23 at 5 36 41 PM](https://user-images.githubusercontent.com/11299125/97059553-560d1200-1556-11eb-9635-7c8a9d90f930.png)
- IMPORTANT: i think this is done in the recipe already but the numbering differs in that it starts with EA or EB. EA1, EA2, EA3... and EB1, EB2, etc 
- icon for set A https://drive.google.com/file/d/1LxWsRljZtkKxBDxQQOgFbrvtDRevOc2D/view?usp=sharing
- icon for set B https://drive.google.com/file/d/1-pbKFNu3ZWzyDYsEjfX0gMnUb_MeYWhh/view?usp=sharing
- exercise number is bolded with period
- LO icon ssame as bove
- spacing between each exercise is 0.7rem

### Problem Set A / Problem Set B `<section class="problem-set-a">` `<section class="problem-set-b">`
[...]

PR #3144, opened 2020-12-08
Style eoc exercises for accounting (no LO links)

kswanson33:
#2893

Resolving LO links in a separate PR.

1.
image

2.
image

3.
image

4.
image

5.
image

6.
image

7.
image


LostMojo (approved):

styles/designs/cardboard/parts/_exercises-components.scss
```diff
@@ -31,6 +31,17 @@ $Exercise__Solution__Wrapper: (
 $Exercise__Solution__Container: exerciseChildContainer((child: solution, name: "SolutionContainer", hasImageTableOnTop: false, hasInlineListOnTop: false));
 
 // General problem and solution components
+$Exercise__ProblemSolution__Text: (
+  _name: "ProblemSolutionNumber",
+  _subselector: ' > .os-text',
+  _properties: (
+    display: table-cell,
+    text-decoration: none,
+    color: enum('ValueSet:::REQUIRED'),
```
ajluk, line 40: @kswanson33 The name for the `$Exercise__ProblemSolution__Text` component should be changed to `ProblemSolutionNumberText` f.ex. Now it has the same name as the components below. Color property value should be set for it in the book.scss file as well. I have also noticed the you haven't regenerate all the styles for all books. Some other books are using this shape as well, so there the change should be also visible and color for this component should be set in their book.scss files as well. Maybe a better solution could be to create a separate shape for exercises for Accounting book with this additional component, what do you think? Let me know what is you opinion @Stackblocks @LostMojo ?

ajluk (changes requested):

styles/designs/cardboard/parts/_exercises-components.scss
Stackblocks: Yeah, the `_name` needs to be changed so there isn't a collision between `$Exercise__ProblemSolution__Text` and `$Exercise__ProblemSolution__Number`.

As for the colors, if there's going to be any variance when reusing a component we should use `enum'ValueSet:::REQUIRED'` in the shape if the setting has to be present, or `enum'ValueSet:::OPTIONAL'` if the setting can be present in some iterations but not all of them; and then put the actual specific setting in the most specific place it can be. We want to build as few new shapes as possible, but as many as is necessary.

Additionally, Anna is right about regenerating all of the styles before pushing your final changes to GitHub and it is a good habit to get into when working with shapes so you can get a flag if something you've changed has had unexpected impacts in other books. (It also keeps Travis from throwing a fit) The same is true of Recipes when you start working on those.

kswanson33: Thank you for catching this! To me, setting the enum value to OPTIONAL seems like the best solution here. Thoughts?

ajluk: Sure, if we prefer less shapes than less properties Alan proposition would be the best and fastest choice :)  
Although in this case not only the color doesn't appear in the feature but the whole component. Alan if we can accept that approach I'm ok with it as well.

ajluk (approved):

ajluk (commented):
I have just left a comment to confirm we're on the same page :relaxed:

Issue #3045, opened 2020-11-11
Business Ethics - EOC Endnotes numbered list not inline as it is in the old PDF


meeragovindaraju:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [ ] styles [ ] both [x] i don't know

COPS PDF

![Screen Shot 2020-11-11 at 1.44.23 PM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/32ed493a-9cbd-4bf5-8588-273c1cb6822a)

PDF on openstax.org

![Screen Shot 2020-11-11 at 1.49.42 PM.png](https://images.zenhubusercontent.com/5b9c2a1115912b2ee068c7e3/8f94dceb-f725-4b0f-843e-da5d1fc9750b)

# Acceptance Criteria

# Bug Location(s)
collection ID: col25722
server: COPS production
chapter/section: EOC all chapters
surrounding text: EOC Endnotes
PDF link: https://openstax-cops-artifacts.s3.amazonaws.com/col25722-latest-production-519.pdf

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
lowish prio, but can we have the second line line up with the start of the first line

ajluk:
Screenshot for reference: 
![image](https://user-images.githubusercontent.com/55203598/101772126-3ec8ba80-3aeb-11eb-8c44-16f09996cee1.png)

lwtchu:
@ajluk looks good thank you

ajluk:
@lwtchu I have added some specs while styling it: 
**EndnoteNoteSection** (whole section with class .references)
 margin: 0,
**EndnoteNote** ([data-type="note"])
padding: '0 0 0 2.2rem',
**EndnoteLink** (a)
        position: relative,
        left: -0.8rem,
        width: 0,
        display: inline-block,
       property removed in this component: margin-right: h-spacing(1),

MartaSztuka:
#### PASSED![Screenshot from 2020-12-15 13-49-49.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/b3ca0984-9eca-401d-adf5-90e568d71d79)

PR #3156, opened 2020-12-10
fix endnotes styling

ajluk:
Issue:  #3045


styles/designs/cardboard/parts/_eoc-components.scss
```diff
@@ -36,20 +36,33 @@ $Term--DL:(
 );
 
 //Endnotes
+$Endnote__Section: (
+    _name: "EndnoteNoteSection",
```
Stackblocks, line 40: This is the only one with `EndnoteNote...` is that intentional?

Stackblocks (changes requested):

styles/designs/cardboard/parts/_eoc-components.scss
ajluk: @Stackblocks It's a typo. Thank you for catching it up!

kswanson33 (approved):

Stackblocks (approved):

Issue #2888, opened 2020-10-22
accounting - chapter outline & introduction


lwtchu:
![Screen Shot 2020-10-21 at 3 46 20 PM](https://user-images.githubusercontent.com/11299125/96785607-a8aec880-13b4-11eb-9909-5aa40422e24e.png)

This is mostly the same as #2655 but notable difference is the absence of the grey bar and the new LO bullets.

### Chapter Outline text
font style: Roboto Condensed 
font size:  font-scale(3): 15.55pt, 20.4px, 1.7rem
color:  B2: `#344456` 

### LO number and bullet treatment
link color: standard link color
treatment: bold
font size: font-scale(0)
font style: noto sans
icon: https://drive.google.com/file/d/14Nu6KIQwuDQcbA-O9RJdIWnHR6WU4Hxp/view?usp=sharing

### Module links/text 
font style: noto sans
font size: font-scale(0)
module title style: normal
module title color: `#000000`

### Why It Matters
Icon: https://drive.google.com/file/d/16nxeqHi64Axd8mYzt-qCpXh2J1XUkaXL/view?usp=sharing
font style: Roboto Condensed 
font-scale(3): 15.55pt, 20.4px, 1.7rem
color:  B2: `#344456`

LostMojo:
Hey, @lwtchu , I just want to double check that the size of the LO buttons match what you picture (They say they're the same height as the text, but appear a little larger)

(This is from the pdf)
![Screen Shot 2020-12-08 at 2.10.39 PM.png](https://images.zenhubusercontent.com/5fa9a1a8a99686b64c824016/75d05ed9-2b7d-411d-be2d-5b5dfad89790)

lwtchu:
@LostMojo hmmmm can you explain the margin thing 😬  like within what margins (page? the list?) i just wanna be on the same page 😅

MartaSztuka:
Hi! @LostMojo Lo numbers colour is light blue, if this is ok, I'll close.![Screenshot from 2020-12-17 13-37-26.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c3158eaa-31ef-4153-8b2d-f187c9f53139)
##### PDF![Screenshot from 2020-12-17 13-41-15.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f7289213-f96e-4221-be93-05fb78aed3b4)

LostMojo:
[...]

Issue #2655, opened 2020-08-20
cardboard theme: chapter outline & introduction


lwtchu:
![Screen Shot 2020-08-20 at 3 57 49 PM](https://user-images.githubusercontent.com/11299125/90825013-ef713b00-e2fd-11ea-9e75-ddac63bf5c02.png)
![Screen Shot 2020-08-20 at 3 57 40 PM](https://user-images.githubusercontent.com/11299125/90825017-f0a26800-e2fd-11ea-9f8f-d5d2342749e0.png)

### Chapter Outline text
font style: Dosis for Astronomy, Roboto Condensed for everything else
font size:  font-scale(3): 15.55pt, 20.4px, 1.7rem
color: P2: `#772949` for Astronomy, B2: `#344456` for everything else

### Module links/text 
font style: noto sans
font size: font-scale(0)
module title style: normal
module title color: `#000000`
number style: bold, 
number color: default link color for entire library

### Bar next to chapter module titles
Line next to chapter contents
width: 0.2727rem (3px)
color: G2: `#F2F2F2`
padding between line and module: 0.5rem

^ Same as TOC

### Introduction 
Icon: Business - https://drive.google.com/file/d/16nxeqHi64Axd8mYzt-qCpXh2J1XUkaXL/view?usp=sharing;
Astronomy - https://drive.google.com/file/d/1C4fs6R-cP0UwIMnp7WrakCwlBjsWOc4d/view?usp=sharing
font style: Dosis for Astronomy, Roboto Condensed for everything else
font-scale(3): 15.55pt, 20.4px, 1.7rem
color: P2: `#772949` for Astronomy, B2: `#344456` for everything else

The icon is the same as corn theme - do we want to try??? for the svg icon #1272

ajluk:
@lwtchu In Introduction to Business and Principles to management there is no Chapter Outline but Learning Outcomes at the start of the Chapter. I have checked and the same content/recipes is on `katalyst01` 
![image](https://user-images.githubusercontent.com/55203598/99248597-0219e400-2809-11eb-9b75-bb4609023d84.png)
I assumed that is desirable, but asking just to be sure.

lwtchu:
@ajluk this is intended. on the card it says it'll be learning outcomes for these titles #2982

ajluk:
@lwtchu Could you add icon for Introduction for Astronomy?

lwtchu:
[...]

PR #3159, opened 2020-12-10
Issue 2888 chapter outline introduction

LostMojo:
Styles the Chapter Outline for #2888 with some overlap with #2655 . (If you bake it, make sure you're looking at the pdf, styling will not work in a browser)

Screen Shot 2020-12-10 at 10 19 40 AM


LostMojo:
The lack of border broke during merging, taking it back to fix

Screen Shot 2020-12-10 at 2 08 07 PM

LostMojo:
All fixed! Also moved the LO boxes out of the margins

ajluk:
@LostMojo I have noticed that the Travis builds failed due to changes added for Introduction page during regenerating recipes css files. If regenerate css files during resolving conflicts won't help there is a config variable I have created for Principles of Management (`$Config_hasGroupedIntroBody`), where was similar problem. You can try to use it in the accounting `_config.scss` file if the problem persists: 
image. It has been added in the `_generator.scss` file as a hack solution to this condition when the template condition is not working properly.

styles/books/accounting/book.scss
```diff
@@ -5,13 +5,19 @@ $PagesWithBands: ();
 @import '../../design-settings/cardboard/_settings.scss';
 @import '../../designs/cardboard/pdf/folio';
 
-// Introduction & splash image
-@include add_settings((
+
+//Settings
+
+@include add_settings ((
     ChapterIntro: (
```
ajluk, line 12: I would be more descriptive if you could change the name `ChapterIntro` to `ChapterIntroWithLO`.

books.txt
```diff
@@ -80,7 +80,7 @@ BOOK_CONFIGS=(
   "additive-manufacturing     additive-manufacturing             d7f23ac9-cf1c-498a-ae49-f34de2cb379f   col10000   content03.cnx.org     mixins/                   cardboard"
   "accounting-vol-2           accounting                         920d1c8a-606c-4888-bfd4-d1ee27ce1795   col67941   katalyst01.cnx.org    mixins                    cardboard"
   "algebra-trigonometry       precalculus                        13ac107a-f15f-49d2-97e8-60ab2e3b519c   col11758   katalyst01.cnx.org    mixins/                   corn"
-  "american-government-2e     american-government                9d8df601-4f12-4ac1-8224-b450bf739e5f   col26739   easyvm5.cnx.org       mixins/"
+  "american-government-2e     american-government                9d8df601-4f12-4ac1-8224-b450bf739e5gif   col26739   easyvm5.cnx.org       mixins/"
```
ajluk, line 83: @LostMojo I guess this uuid change for american government wasn't intentional. If yes it should be reverted.

yarn.lock
```diff
@@ -132,6 +132,18 @@
   resolved "https://registry.yarnpkg.com/@nodelib/fs.stat/-/fs.stat-1.1.3.tgz#2b5a3ab3f918cca48a8c754c08168e3f03eba61b"
   integrity sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw==
 
+"@openstax/bakery-cli@^4.0.0":
```
ajluk, line 135: @LostMojo I guess this change was intentional, but I'm asking just to be sure? :relaxed:

ajluk (changes requested):
Looks good! I have only left few comments with some suggestions and questions.

yarn.lock
LostMojo: It was not, thank you for catching these!

styles/books/accounting/book.scss
LostMojo: It looks like changing it will break some of the ChapterIntro infrastructure that ChapterIntroWithLO is borrowing.

ajluk: @LostMojo I guess the cause of this is that you need also add settings for the new shape in the `desing-setiings/cardboard/parts/_chapterintro-settings.scss` file like it was done for the ChapterIntro shape: 
image

LostMojo: Excellent! Thank you!!

yarn.lock
```diff
@@ -1,7019 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kswanson33, line 1: Have a feeling that this file should not be deleted (but I don't actually know)

styles/books/accounting/book.scss
```diff
@@ -251,10 +254,12 @@ $PagesWithBands: ();
 
 // Notes
 @include use('NotesLists', 'cardboard:::NotesListsShape');
+
```
kswanson33, line 257: Looks like there are two `Notes` sections now, with some overlap, likely from merging. Just make sure that all `includes` are present without repeats! :smile:

kswanson33 (changes requested):
The changes made regarding the outline itself look good! I left some comments about seemingly stray edits to other files.

recipes/output/business-ethics.css
```diff
@@ -2328,21 +2328,6 @@
...
-
-:pass(14) .introduction::after {
-  content: pending(intro-text);
-  move-to: intro-body;
-  class: "intro-text"; }
-
-:pass(14) .introduction:deferred::after {
-  content: pending(chap-outline) pending(intro-body);
-  class: "intro-body"; }
-
```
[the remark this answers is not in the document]
LostMojo: nope! reverting it

styles/designs/cardboard/parts/_chapterintro-components.scss
```diff
@@ -43,6 +43,44 @@ $ChapterOutline__Objective--Link: (
...
+    _name: "OutlineObjectiveLOButton",
+    _subselector: '::before',
+    _properties : (
+        display: inline-block,
+        background: enum('ValueSet:::REQUIRED'),
+        background-size: contain,
+        height: font-scale(0),
+        width: font-scale(1),
+        content: "''",
+        padding-right: .5rem,
```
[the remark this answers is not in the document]
LostMojo: ooo, yes!

yarn.lock
LostMojo: This is one of the bois bein' reverted to it's past self

LostMojo:
Okay, this is going to be a whole lot nicer and have a much more sensible commit history if I start a new branch from the more recent zesty.zendaya so I'm moving the changes onto there (only on the files I want to change) and switching the pull request to that.

Issue #3025, opened 2020-11-06
Astronomy: Learning Objectives


brittweinstein:
Should follow https://github.com/openstax/cnx-recipes/issues/2711

kswanson33:
specs: #2711
changes applied to: learning objectives in astronomy
PR: #3179
astronomy is on katalyst01 with col11992

MartaSztuka:
#### PASSED![Screenshot from 2020-12-22 10-19-08.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/e9c1ee79-0f7f-456a-a7c7-9e6c7b96a825)

PR #3179, opened 2020-12-14
Style learning objectives for astronomy

kswanson33:
#3025

image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2631,9 +2631,9 @@ he@1.2.0, he@1.2.x:
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
 highlight.js@^9.12.0:
-  version "9.12.0"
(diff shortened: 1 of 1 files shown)


styles/designs/cardboard/parts/_learning-shapes.scss
```diff
@@ -13,4 +13,20 @@
       ),
     ))
   ),
+));
+
+@include create_shape('cardboard:::LearningObjectivesCustomShape', (
```
ajluk, line 18: @kswanson33 Could you add a comment that this shape is used only in Astronomy?

ajluk (commented):
Everything looks good. I have just left one comment.

styles/books/astronomy/book.scss
```diff
@@ -27,7 +27,18 @@ $bandColor: #772949;
 ));
 
 
+@include add_settings((
+  LearningObjectivesAstro: (
```
ajluk, line 31: Here maybe the better title would be `LearningObjectivesCustom`. What do you think?

kswanson33: Agreed -- for consistency's sake, either all settings should include 'Astronomy' in the name, or none of them should. I'm with you that none seems like the better choice.

ajluk (approved):

kswanson33:
Git tangle alert! This is already merged into a.a., and all that's left is changes from master that weren't in z.z (and somehow made it into this branch before a.a.). Sorry!

Issue #3188, opened 2020-12-15
accounting - pull in other items from existing themes


lwtchu:
#2712 - documentation

Hello!  I am not sure if these were implemented yet - would you mind going through to check? Thanks!

kswanson33:
- [x] `<term>` and `<term class="no-emphasis">`
Elements with data-type="term" link to index. Elements with class="no-emphasis" are not bold, while the others are bold.

- [x] table cell alignments driven by content
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/372f849c-1e3c-45a8-b9b8-f58d38639ca0)

- [ ] folio structure (except font is Noto Sans and color is #000000)
Preface: 
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/145440d1-045f-4a96-a6f7-528c81e4e1bf)
Chapter header:
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/b43a26c1-1bd4-4350-97ee-35f13007144b)
Odd page #
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/1a5033b0-f6b1-4f42-b505-7ba18835ddf0)
Even page #
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/ce1c07b9-541a-42d6-85a8-056dcd30495e)
**BUG
Key terms, Summary, Questions, Multiple Choice, Exercise Sets & Problem Sets, Appendices, Answer Key, Index: page number is doubled**
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/b4ab5dd4-cee1-416f-be4c-a8f66a3c3242)
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/84490368-3001-40b7-a5e6-2886872c032a)
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/e77d18c4-5c49-4ec1-a32e-357f05038f19)

- [x] standard link color and treatment
![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/d5632e2a-21b1-47ff-9647-0f2ee82de0e3)

- [x] list indentations (stepwise, no label, bulleted, alpha lists)
I scanned through, looks good.
[...]

PR #3195, opened 2020-12-15
style accounting bands

kswanson33:
#3188

image


styles/books/accounting/book.scss
```diff
@@ -1,4 +1,10 @@
-$PagesWithBands: ();
+$PagesWithBands: (
+  (pageName: eoc, firstSelector: '.os-eoc.os-section-summary-container', generalSelector: '.os-eoc'),
```
wilkus27, line 2: I think the first section in EOC is Key Terms. If I'm not wrong, selector is `glossary`.

wilkus27 (changes requested):

styles/books/accounting/book.scss
kswanson33: image

You're right, the first EOC section is `.os-glossary-container`. For some reason I don't fully understand, it doesn't seem to matter what I put in `firstSelector` _or_ `generalSelector`; as long as the `pageName` is set to eoc, all of the eoc pages get the band (when building the pdf with prince). 

That being said, I'll make this change for now. I'd be interested in learning a bit more about this variable, and whether `firstSelector` and `generalSelector` are necessary for deployment?

styles/books/accounting/book.scss
```diff
@@ -1,4 +1,10 @@
-$PagesWithBands: ();
+$PagesWithBands: (
+  (pageName: eoc, firstSelector: '.os-eoc.os-glossary-container', generalSelector: '.os-eoc'),
+  (pageName: eob, firstSelector: '.os-eob[data-type="composite-chapter"]', generalSelector: '.os-eob'),
+  (pageName: appendix, firstSelector: '.appendix', generalSelector: '.appendix'),
+);
+
+$bandColor: #24739E;
```
LostMojo, line 7: Is there a reason the band color is here instead of in _colors.scss?

LostMojo (commented):
Looks pretty good-- but I'm curious why $bandColor is placed where it is

styles/books/accounting/book.scss
kswanson33: Good question! This is how it's done in all other books as well. From what I can tell, it has something to do with the framework design and compile order. I investigated, and found that when `$PagesWithBands` is non-empty, `$bandColor` must be defined in order for the imports from /designs and /design-settings to compile. I would imagine that there's probably a way to disentangle this, but it would require a more serious refactor.

styles/books/accounting/book.scss
wilkus27: For `firstSelector` we apply additional property ` prince-page-group: start`. More details you can find in `bands.scss` file. Here is the path `styles > designs > common > pdf > bands.scss`. In this theme I think it doesn't matter. We don't apply specific settings for first pages of the group, but I think in one of the previous theme this property was necessery. Maybe in Carnival :thinking: I'm not sure. Here is also prince documentation about page groups: https://www.princexml.com/doc/11/advanced-paged/#page-groups.

styles/books/accounting/book.scss
wilkus27: @kswanson33 Yes, exactly, this is the reason. It requires too much work to refactor this, so we decided to hard code band color before imports.

wilkus27 (approved):

LostMojo (approved):

Issue #3034, opened 2020-11-06
Astronomy: Key Terms


brittweinstein:
(uses `<glossary>`) - https://github.com/openstax/cnx-recipes/issues/2742 please use the same specs here but use Dosis header - P1: #3B2746 with this icon: https://drive.google.com/file/d/1hG9R4OQZFAU2QPyWY3zY1bcDMx7JO2AH/view?usp=sharing - screenshot is in the issue for reference

kswanson33:
specs: #2742 & this card
changes applied to: astronomy EOC key terms
PR: #3205
astronomy is on katalyst01 with col11992

MartaSztuka:
#### PASSED
![Screenshot from 2020-12-22 17-44-28](https://user-images.githubusercontent.com/47241114/102912209-691c5f80-447d-11eb-9052-4675b64fa0e0.png)

Issue #1860, opened 2020-02-04
CLEAN UP: refactor titles to use only one component per header type


helenemccarron:
We should have one component per header type (h1, h2, h3, h4, h5, h6), naming should reflect what type of header it is (not using subsubsubsub... title). 

Components should be reused in other components each time we need a title component. Each book will have book specific colors, there might be a couple optional fields (borders for example).

helenemccarron:
A lot of the initial work was done as part of the Space Saving epic. We still need to create more global different kind of titles (with icon, number, etc).

Stackblocks:
Handled piecemeal as part of other issues.

PR #3205, opened 2020-12-16
Style key terms astronomy

kswanson33:
#3034

image


styles/books/astronomy/book.scss
```diff
@@ -33,9 +33,18 @@ $bandColor: #772949;
     'H2WithIcon:::font-family': (_ref: 'typography:::titleFontAstronomy'),
     'H2Icon:::background': url(toDataUri("svg+xml", "designs/cardboard/resources/cardboard-summary2.svg")) no-repeat top left,
     'H3LinkTitle:::font-family': (_ref: 'typography:::baseFont'),
-    'H4Title:::font-family': (_ref: 'typography:::baseFont'),
-    'H4Title:::color': (_ref: 'colorMap:::titleColor'),
-  )
+    'H4Title:::font-family': (),
+    'H4Title:::color': (),
```
ajluk, line 37: @kswanson33 Was it intentional to remove the properties for H5?

ajluk (commented):
Everything looks good! I have just one question.

LostMojo (approved):
Looks good! Just the same question about the H4 headers.

styles/books/astronomy/book.scss
kswanson33: It was intentional, as neither summary nor key terms have H4 headers. 

It would probably be better to change these properties from `REQUIRED` to `OPTIONAL` in the shape. I could add that to this PR, or open a new one. What do you think @ajluk @LostMojo ?

ajluk: Oh, if that is the case I'm not sure, because in the rest of the books I have added these even if they weren't there just to not make them OPTIONAL (to avoid these will be omitted in the other features I have left them REQUIRED), but let me know what do you think?

ajluk: @wilkus27 could you advise?

LostMojo: Hm, I'm not sure. If most other key terms have h4 subheaders, it might be worth it to keep it required to make sure that the h4 doesn't go unstyled for those others, but even if 30% of the books that use this theme don't have h4 in their keyterms, optional seems like a good idea?

wilkus27: In settings we specify REQUIRED and OPTIONAL properties. But remember that components have also hard coded values specified in `-components.scss` file. So @kswanson33  in this case you don't style color and font family for H4, but there are still added other styles for this header (font size etc.). If you don't want to style some element (here H4), because you know it doesn't appear in a book you have to create a separate shape without this component.  For now we decided to have more general shapes (one for EOC titles) rather then detailed (separate for each section from EOC). But it's something we can discuss during the cleaning up fest :wink:

kswanson33: That's a good point about styles being added anyway via the component. I won't change anything about the REQUIRED enums in the shape. I'll still leave these H4 values blank here, though, since it seems misleading to provide values just for filler's sake.

wilkus27: Yes, I agree and I think it's a very good idea to leave blank values when some element doesn't exist in the book. To be honest, I don't know why we didn't come up with this idea earlier :smile:  But, it's something that we didn't use before, so I'm not sure if it's ok to change the convention in the middle of the work on theme. In the previous themes when we had an idea how to impove code we created issues with `Clean up` at the beginning of the name and there described idea for refactoring. Here are some examples: https://github.com/openstax/cnx-recipes/issues/1860 , https://github.com/openstax/cnx-recipes/issues/2040. What do you think to create a document where we can put all of ideas how to improve/refactor code? We can discuss there what are pros and cons, and when we decide that we want to change something, then we will create issue for that and apply it everywhere, not just in a few places.

wilkus27 (approved):

Issue #2889, opened 2020-10-22
accounting - LO links


lwtchu:
![Screen Shot 2020-10-22 at 9 37 04 AM](https://user-images.githubusercontent.com/11299125/96887432-32f53c00-144a-11eb-97e8-c3b8f59b4bd4.png)

- targeted by a class:  `<link class="lo-reference" document="m#####"/>`
- use this icon https://drive.google.com/file/d/12qOsipeGDvc4PVkpJ-0oFPFnJ050TaOt/view?usp=sharing
- a module link with no anchor text typically returns the words `Section #.#`, this should remove the word Section and replace with the LO svg
- have the blue link be the default link color but WITHOUT the underline
- font and size matches the base text font/size

kswanson33:
The specified selector, `class="lo-reference"`, is not present in accounting

lwtchu:
@kswanson33 😬 I'm starting to have a deep dark fear that whenever it gets transformed from CNXML to HTML, it gets wiped out. we might require the forces of BrittanyJ or ...maybe even Phil. I'll wait to see if BrittanyJ has any insights though. I can at least confirm it's on the CNXML on katalyst01 and production.

kswanson33:
@lwtchu Could always just be a me-problem 🌱 (although I did re-run all my scripts, and still no luck.) I'll wait till after the break on this one!

lwtchu:
@kswanson33 oh no i'm looking at the content on CNX when it's baked...and i don't see the class :joy: so i have a strong scary feeling that some work might need to be done. but also I will pocket the fear for after thanksgiving since we need to be positive for the pie.

kswanson33:
🥧🥧🥧

kswanson33:
Status: Styling for LO links is implemented. Changes to recipe are required to remove the "LO" text before the section number. Currently the accounting recipe adds `LO #.#` to the content in plain text. 

@wilkus27 has [suggested](https://github.com/openstax/cnx-recipes/pull/3222#discussion_r547364147) that "LO" be wrapped in a span element that can be targeted and hidden with `display:none`. I think this is the best way too.

After recipe updates, the HTML should look something like this:

![image.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/3f4e4470-d539-4e01-8d80-7dc1a5ace100)

The recipe work will likely be done after the break.

lwtchu:
@kswanson33 following up on this - did you need anything :eyes:

[...]

PR #3222, opened 2020-12-17
LO styling for accounting

kswanson33:
#2889

image


recipes/books/accounting/_config.scss
```diff
@@ -224,7 +224,7 @@ $Config_PartType_Chapter:  (outlineTitle: 'Chapter Outline');
 $Config_ResetSectionCounterForEOCLO: true;
 
 $Config_TargetLabels : append($Config_TargetLabels,
-(selector: 'div[data-type="chapter"] > div[data-type="page"]:not(.introduction)', labelText: "LO " counter(chapter) "." counter(section))
```
kswanson33, line 227: Question: I chose to change the recipe in order to remove the "LO" from the links. Is there a better way to do this?

LostMojo: Hm, I'm not aware of one-- but maybe one of the katalyst folks know?

styles/books/astronomy/book.scss
```diff
@@ -26,8 +26,35 @@ $bandColor: #772949;
   )
 ));
 
+@include add_settings((
+  SummaryTitles: (
+    _selectors: ('.os-summary-container'),
+    'H2WithIcon:::color': (_ref: 'colorMap:::titleColorAstronomy'),
+    'H2WithIcon:::font-family': (_ref: 'typography:::titleFontAstronomy'),
+    'H2Icon:::background': url(toDataUri("svg+xml", "designs/cardboard/resources/cardboard-summary2.svg")) no-repeat top left,
```
LostMojo, line 34: This PR is for Accounting LO-- did you mean to include Astronomy in it?

styles/design-settings/cardboard/_colors.scss
```diff
@@ -31,6 +31,7 @@
   colorMap: (
     fontBodyColor: (_ref: "scheme:::black"),
     titleColor: (_ref: "scheme:::B4"),
+    titleColorAstronomy: (_ref: "scheme:::P1"),
```
LostMojo, line 34: Same as before-- were these astronomy changes intended for this PR?

LostMojo (approved):
Looks good! But you do seem to have added in some Astronomy changes along with the accounting?

styles/design-settings/cardboard/_colors.scss
kswanson33: Rebased!

LostMojo (approved):

recipes/books/accounting/_config.scss
wilkus27: When we change something in the recipe it also has an impact on webview. So when we remove "LO" it will also disappear from here:

https://openstax.org/books/principles-financial-accounting/pages/1-multiple-choice
image

(when recipe changes be deploy to production). I don't know if we want to make such change there. Please contact with Larissa. 

If we don't want to change anything on webview we have to make bigger change in recipe.
We have to modify mixin related with target labels or create new one for custom labels to create something like that:

image

Then we can hide span with "LO" in styles adding `display:none` and instead of that add icon. It won't have an impact on webview.

I put LO and number in spans because I'm not sure if we can put divs inside the `<a>`.

wilkus27 (changes requested):

recipes/books/accounting/_config.scss
wilkus27: But probably spans will be better. We can also add some classes, like `os-text` and `os-number`.

kswanson33: I like that idea. I'll split this into 2 issues, one for styles and one for the recipe update, so that the styles at least can be ready for merge before the break.

wilkus27 (approved):

Issue #3240, opened 2020-12-21
Astronomy: implement Collaborative Group Activities


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/2907

MartaSztuka:
##### PASSED![Screenshot from 2021-01-08 11-31-20.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/d121943e-5572-4f4a-9642-ba4d772dee3d)

PR #3282, opened 2020-12-22
Style Group Activities section from Astronomy

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3240


styles/designs/cardboard/parts/_eoc-components.scss
```diff
@@ -84,3 +84,28 @@ $ModuleWithIndentation__Para: (
...
+    )
+);
+
+$ModuleWithTypeAList__Marker: (
+    _name: "TypeAListMarker",
+    _subselector: '::marker',
+    _properties:(
+        font-weight: bold,
+    )
+);
```
wilkus27, line 111: I wasn't sure what would be better, to create these components here or in `lists-components` file. I decided to put them to `eoc-components` becuase otherwise shape would contain only components from other files, not related with eoc and it might be misleading. Let me know what do you think :slightly_smiling_face:

ajluk: Yes, I think it's ok. I guess hese components can be moved to `lists-components` and we can create a new shape for this kind of a lists if they will be present in different features/modules.

ajluk (approved):

kswanson33 (approved):

Issue #3197, opened 2020-12-15
sociology - sociology in the real world feature


lwtchu:
- class: `sociology-real-world`
- uses #3193 
- title: Sociology in the Real World
- has `<title>`
- icon: [cosmos-world.svg](https://drive.google.com/file/d/1QfMTz9Wq_Iv3H1JL0jxJgXe_bgla9Hk4/view?usp=sharing)
- color: B2: `#0074BC`
- should contain mostly paragraphs

MartaSztuka:
#### PASSED![Screenshot from 2021-01-15 14-36-04.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8d2f80a8-e24f-4e71-bbe8-f61d4becf912)

PR #3329, opened 2021-01-11
style sociology in the real world note

kswanson33:
#3197

image


styles/books/sociology/book.scss
```diff
@@ -9,6 +9,24 @@ $bandColor: #DCB83D;
 @import '../../design-settings/cosmos/_design.scss';
 @import '../../design-settings/cosmos/_settings.scss';
 
+@include add_settings((
+    RealWorld: (
+        _selectors: ('.sociology-real-world'),
+        'Title:::color': (_ref: 'colorMap:::fontBodyColor'),
+        'Title:::font-family': (_ref: 'typography:::titleOption3Font'),
+        'Title:::padding-left': v-spacing(4) + 0.5,
```
wilkus27, line 17: For horizontal spacing we are using `h-spacing` function. If I remember correctly `h-spacing(1)=8px`. If we need custom value we just hard code it with comment why it's custom.

styles/designs/cosmos/parts/notes/_notes-components.scss
```diff
@@ -0,0 +1,70 @@
...
+    _properties: (
+        color: enum('ValueSet:::REQUIRED'),
+        font-family: enum('ValueSet:::REQUIRED'),
+        font-size: font-scale(1),
+        line-height: line-height(1),
+        text-transform: uppercase,
+        font-weight: normal,
+        border-bottom: solid 0.2rem,
+        border-color: enum('ValueSet:::REQUIRED'),
+        padding-left: enum('ValueSet:::REQUIRED'),
```
wilkus27, line 22: Why this padding is set to REQUIRED and not hard coded? There will be other value for it in another notes? Just the question :wink:

styles/designs/cosmos/parts/notes/_notes-components.scss
```diff
@@ -0,0 +1,70 @@
...
+        padding-right: v-spacing(1),
```
wilkus27, line 23: As I mentioned in my previous comment, for horizontal spaces we are using `h-spacing`.

wilkus27 (changes requested):

styles/designs/cosmos/parts/notes/_notes-components.scss
kswanson33: The padding is dependent on the icon size. Since position is absolute for the icon, I didn't see a good way to make the text line up next to the icon without doing it manually. So it wouldn't make sense to hard code the padding here without hard coding the icon size as well. Do you see a better way to do this - I admit I don't think it's a _great_ solution

styles/books/sociology/book.scss
kswanson33: So, the height and width of the title icon are set to v-spacing(4) since it needs to be square, and v-spacing(4) was used for a similar feature in carnival. The title text needs to be left-aligned next to the icon, with 0.5rem of additional spacing. So (with the current implementation) the same spacing should be used for all of them, whether it's v-spacing(4) or something else.

wilkus27: Hmm :thinking: For the icon I think it's ok to use v-spacing for width and heigh. It's something we used to do in previous themes. About icon and text, I got the same effect with such properties:

image

image

I tried it just in the browser, so I'm not 100% sure if it works in PDF. But... I don't know if this solution is something we should go with. The goal is to achive some effect and it could be done in different ways :wink: My only concern was about the value of padding. We expect h-spacing and it's confusing in current form. I think it is ok to add just a comment that it's equal to width of icon plus h-spacing(1) (h-spacing(1)=8px=0.7rem=v-spacing(1)) and it will be ok. I'm curious what do you all think :slightly_smiling_face:

styles/designs/cosmos/parts/notes/_notes-components.scss
wilkus27: I added comment above with proposition how we might do this without depending text on icon. But if you decide to stay with current solution I think it's ok to add just comment, that this value depends on icon and it's different in each note. :wink:

styles/books/sociology/book.scss
ajluk: Yes, I think the same effect could be achieved different ways in styles. So both solutions are good here. The important thing as Natalia is writing above is to make comments where we are using some custom values etc.

kswanson33: Thanks @wilkus27 - this works, and I think it's a better solution!

ajluk (approved):

wilkus27 (approved):

Issue #3200, opened 2020-12-15
sociology 2e - sociological research


lwtchu:
- class: `sociological-research`
- uses #3194
- title: Sociological Research
- has `<title>`
- top bar color: B1: `#38568A`
- bottom box color: B4: `#EFF3FA`
- should contain mostly paragraphs, sometimes figures

MartaSztuka:
#### PASSED![Screenshot from 2021-01-15 14-34-51.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/851f9d79-1c75-40d0-95fd-5af4ccd18e67)

PR #3332, opened 2021-01-11
style sociological research note

kswanson33:
#3200 

image


styles/output/sociology-pdf.css
```diff
@@ -568,4 +568,36 @@ sub {
...
+  background-color: #EFF3FA; }
+
+.sociological-research > .os-note-body > h4.os-subtitle {
+  color: #000000;
+  font-family: Archivo, sans-serif, StixGeneral;
+  font-size: 1.2rem;
+  line-height: 1.5rem;
+  font-weight: 500;
+  margin-bottom: 0rem; }
+
```
wilkus27, line 602: @kswanson33 I see that there is font-weight set to 500 for Archivo. Can you add this option to font import:

image

Previously I added only types which I have found in headers specs.

wilkus27 (changes requested):

styles/output/sociology-pdf.css
kswanson33: Thank you!

wilkus27 (approved):

ajluk (approved):

Issue #3212, opened 2020-12-16
sociology - section quiz


lwtchu:
Documentation #3211 

Class: `class="section-quiz"`
broken up by module sections
title: Section Quiz

MartaSztuka:
@ajluk Headers are unstyled![Screenshot from 2021-01-14 16-50-39.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4d754154-3ab5-4bd8-84b0-0b1d521e6ac3)

MartaSztuka:
#### PASSED![Screenshot from 2021-01-15 08-59-35.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/2ac6bc9a-ca0f-4910-aaa1-12da2a34b831)

PR #3338, opened 2021-01-12
style section quiz links sociology

ajluk:
Issue: #3212


styles/books/sociology/book.scss
```diff
@@ -9,13 +9,30 @@ $bandColor: #DCB83D;
 @import '../../design-settings/cosmos/_design.scss';
 @import '../../design-settings/cosmos/_settings.scss';
 
+
+//Settings
+
+@include add_settings((
+  SeqtionQuiz: (
```
Stackblocks, line 16: "Section" is misspelled throughout the change as "Seqtion". Is this intentional? If it doesn't have a specific purpose I think we should use the "Section" spelling for searchability.

styles/design-settings/cosmos/parts/_lists-settings.scss
```diff
@@ -3,3 +3,9 @@
         _selectors: ('[data-type="page"]'),
     ),
 ));
+
+@include add_settings((
+    ExercisesProblemLists: (
```
Stackblocks, line 8: Should this be the singular "ExerciseProblemList" to match other "Exercise..." pieces? I think we should try to standardize in one direction or the other.

styles/designs/cosmos/parts/_exercises-components.scss
```diff
@@ -0,0 +1,47 @@
+$ExercisesModule__Container: empty_wrapper(ExercisesModuleContainer, '');
```
Stackblocks, line 1: Same Exercises / Exercise comment as above

Stackblocks (changes requested):

styles/books/sociology/book.scss
ajluk: No, it is a typo. Thank you Alan!

styles/design-settings/cosmos/parts/_lists-settings.scss
ajluk: @Stackblocks I fully agree, but I have used plural in this case because we had`ExercisesModuleShape` and `_exercises-components` and shapes files in the themes. I have used the name from there.

styles/designs/cosmos/parts/_exercises-components.scss
ajluk: Yes I kept it for consistency here as well.

Stackblocks (approved):

kswanson33 (approved):

Issue #3216, opened 2020-12-17
sociology - references


lwtchu:
Documentation: #3214
Class: `class="references"`
title: References
broken up by module sections

kswanson33:
![Screen Shot 2021-01-15 at 2.23.43 PM.png](https://images.zenhubusercontent.com/5fa9a1a6a99686111a824015/141177b7-64d2-4879-9f1c-84b5d2d82acb)

@lwtchu @brittany-johnson I'm seeing some weird stuff re: links in References. Most of the URLs in References are plaintext, not wrapped in an `<a>` element (eg. green arrow). However, some of them are links, and whenever that happens, code from `_common-settings.scss` adds an `::after` tag to the element containing the URL. In this case, the content of the link already is the URL. This causes the link text to be repeated twice (eg. red arrow).

Is this a content problem? A styles/recipes problem? And what ultimately should the specs for these URLs be (clickable or static, styled with link styling or black text & no underline)?

lwtchu:
@kswanson33 this is a content problem. can you just leave it as is for now and i'll open a bug on the CM side once the book is done?

kswanson33:
@lwtchu Sounds good, thanks!

lwtchu:
Linking CM card here: https://github.com/openstax/content-managers/issues/2083

MartaSztuka:
#### PASSED![Screenshot from 2021-01-21 10-14-02.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3b531afd-3938-4a12-bc0b-0e887fd724dd)

PR #3353, opened 2021-01-13
style references for sociology eoc

kswanson33:
#3216


styles/design-settings/cosmos/_design.scss
```diff
@@ -10,7 +10,7 @@
 @import 'designs/common/reset';
 @import 'designs/common/pdf/footnotes-shapes';
 
-
+@import 'designs/cosmos/parts/toc-shapes.scss';
```
wilkus27, line 13: `toc-shapes.scss` is already imported on line 17

styles/designs/cosmos/parts/_eoc-shapes.scss
```diff
@@ -1,5 +1,15 @@
...
+@include create_shape('cosmos:::EndnoteModuleShape',(
+    _components:(
+        map-merge($Eoc__Container,(
+            _components:(
+                $Eoc__Reference,
+            )
+        )),
+    )
+));
+
```
wilkus27, line 12: I just noticed that this section is styled the same as Suggested Reading from American Government. I created for it `ModuleWithIndentationShape`. Changes are already on b.b : https://github.com/openstax/cnx-recipes/pull/3369/commits/deafa9ff56fdfee7560af32b671510b42812c660. I think References can use the same shape. Sorry, I didn't know that you are working on something similar. Issue for Suggested Reading doesn't have link to any documentation. Specs are listed inside the issue, so I didn't know that there are other sections with similar specs.

wilkus27 (changes requested):

styles/designs/cosmos/parts/_eoc-shapes.scss
kswanson33: Thanks for letting me know! Good name, too. I'll use that instead.

wilkus27 (approved):

ajluk (approved):

Issue #3186, opened 2020-12-14
sociology - table style


lwtchu:
#3185 

I think only Table style 1 exists. Tables should have standard `<caption>` and  `<title>`

LostMojo:
12 Test Content Location(s)
collection ID: col11762
server: katalyst
chapter/section: assorted chapters
surrounding text: TABLE X.X

MartaSztuka:
#### PASSED![Screenshot from 2021-01-21 16-35-50.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/2b1cd750-0c2c-4ef1-ba4e-1cb4d2e6f20c)

Issue #3185, opened 2020-12-14
cosmos theme - table


lwtchu:
# Caption, title, number
## Table number
font: Archivo
weight: bold, 700
color: R2: `#8B1504`
treatment: all caps
font size: font-scale(0): 9pt, 12px, 1rem

## Table title (if in the `<title>` tag)
font: Archivo
weight: Medium, 500
color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem

## Figure caption
font: IBM Plex Sans
weight: regular, 400
color: Black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem

# Type 1 Regular Table
![Screen Shot 2020-12-14 at 4 17 47 PM](https://user-images.githubusercontent.com/11299125/102142399-f3cee000-3e27-11eb-88c5-dadcd1dc009b.png)

## Header row 
font: IBM Plex Sans
font treatment: bold, 700
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: only top and bottom
cell border color: black `#000000`
cell border weight: .25px

## Non header rows
font: IBM Plex Sans
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: all around unless it is the cells on the left and right (see screenshot)
cell border color: black `#000000`
cell border weight: .25px

## Bottom stroke of table
color: R2: `#8B1504`
cell border weight: 1.5px

ajluk:
@lwtchu Here is the example of styled top titled table from American Government: 
![image](https://user-images.githubusercontent.com/55203598/105499319-939e4a00-5cc1-11eb-90b1-05e14877e016.png)
I have only added padding (0.7rem) to the specs you have provided above for the Top Title.

lwtchu:
@ajluk i think this looks good - do you know if the top alignment in the header row is due to something in the HTML?

ajluk:
@lwtchu It's only centered in the code (as in the specs) but I have added paddings so it can look as top aligned, but it isn't actually: 
![image](https://user-images.githubusercontent.com/55203598/105524879-8d6b9600-5ce0-11eb-8a96-8d0223556956.png)

lwtchu:
@ajluk oh sorry should've been more clear i meant the "why Was This a Problem" but i see it in the HTML... :joy: this looks fine though we can go with what you currently have.

thank you <3

PR #3371, opened 2021-01-14
Issue 3186: Adding Sociology Tables

LostMojo:
#3186 using #3185 

Screen Shot 2021-01-14 at 10 46 57 AM


LostMojo:
With current caption styling

styles/design-settings/cosmos/_typography.scss
```diff
@@ -9,6 +9,7 @@
 @include add_settings((
     typography: (
         baseFont: ('IBM Plex Serif', serif, StixGeneral),
+        sansFont: ('IBM Plex Sans', sans-serif, StixGeneral),
         titleOption1Font: ('Archivo', sans-serif, StixGeneral),
         titleOption2Font: ('IBM Plex Sans', sans-serif, StixGeneral),
```
ajluk, line 14: @LostMojo The font you have added within this PR is already there under name `titleOption2Font`.

ajluk (changes requested):
Everything looks good beside the doubled font needs to be removed in `typography.scss file`

Stackblocks (approved):
Looks good once the doubled font is removed!

styles/design-settings/cosmos/_typography.scss
LostMojo: Should we refactor the "titleOption1Font" font names into something more descriptive if they're going to be used for things besides titles?

ajluk: Yes, it is a good idea! But if you will decide to do it, please remember to check where it has been used (could be many places now) and change it there too or I guess you can keep the old name (with a comment that this need to be cleaned) for now and  open up a clean up issue which we can take care of later? @Stackblocks what do you think?

LostMojo: I went ahead and just changed the `sansFont` to `titleOption2Font` for easy refactoring later, and I think it would be best to just open up a new clean up issue, as it would affect more than just the tables.

Issue #3223, opened 2020-12-17
american government - table of contents


lwtchu:
Documentation: #3154

Has units.

MartaSztuka:
#### PASSED![Screenshot from 2021-01-19 12-10-28.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ea3a5c16-25d8-4ced-8685-d40ec35afee0)

wilkus27:
Reopen to fix Unit styling. Here are the specs: https://github.com/openstax/cnx-recipes/issues/3154#issuecomment-762965934

MartaSztuka:
#### PASSED![Screenshot from 2021-01-22 15-51-43.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fed8f1be-1f0b-4bec-b0e9-5280d9af590d)

PR #3380, opened 2021-01-15
Style Toc in American Government

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3223


LostMojo (approved):
It probably doesn't matter, but I see the sociology css is included!

styles/design-settings/cosmos/parts/_toc-settings.scss
```diff
@@ -6,35 +6,80 @@
             font-family: (_ref: 'typography:::titleOption2Font'),
             border-bottom-color: (_ref: 'colorMap:::tocTitleColor'),
         ),
-        SecondLevelLinkNoChapterPage: (
+        NoUnitChapterLinkPage: (
             color: (_ref: 'colorMap:::fontBodyColor'),
             font-family: (_ref: 'typography:::titleOption2Font'),
         ),
-        SecondLevelLinkNoChapterNumber: (
+        NoUnitChapterLinkNumber: (
```
ajluk, line 13: @wilkus27 As we discussed there is `NoUnitChapterLinkNumber` component present in both shapes (with and without units), it could be useful to add a comment why this name was used for a component present in both shapes.

ajluk (commented):
I have only left a comment about the name of one of the components which can be confusing.

wilkus27:
@LostMojo Yes, these changes are intentional. Sociology has only Index in EOB, but other books without units may contain appendix and answer key, so I added styles for these parts to both shapes (TocNoUnit and TocUnit).

ajluk (approved):

Issue #3154, opened 2020-12-09
cosmos theme - table of contents (toc)


lwtchu:
![Screen Shot 2020-12-09 at 4 33 12 PM](https://user-images.githubusercontent.com/11299125/101697280-2732f880-3a3d-11eb-81ce-453e5b8ccd93.png)
![Screen Shot 2020-12-09 at 4 33 57 PM](https://user-images.githubusercontent.com/11299125/101697281-27cb8f00-3a3d-11eb-9e95-6ea8def641c2.png)

### Text
- font: [Archivo](https://fonts.google.com/?query=archivo)
- color: R1: `#C31427`
- treatment: all caps
- size: font-scale(6): 26.87pt, 36px, 3rem

### Line
- width: 6.5 in
- padding after CONTENTS: 2rem
- line weight: 3px
- color: R1: `#C31427`

wilkus27:
@lwtchu This is how it looks in Sociology:

![image](https://user-images.githubusercontent.com/40228854/103906156-d4ae3000-50ff-11eb-98e4-76f81538747c.png)

End of the TOC:

![image](https://user-images.githubusercontent.com/40228854/103906218-ea235a00-50ff-11eb-9bca-0882284b5df9.png)

In this book there is only Index in EOB.

Should I change something?

lwtchu:
@wilkus27 can we add a space to separate the chapters from the front of book and back of book content? is there a way to add like 1 or 2 h grid to separate it? (see the space between preface and the first chapter) 

thanks!

wilkus27:
@lwtchu Unfortunately it's not so easy to apply these spaces :disappointed: 

At this moment we have such markup:

```
ol
  li.os-toc-preface
  li.os-toc-chapter
  .
  .
  .
  li.os-toc-chapter
  li.os-toc-index
```

The problem is that each book has different elements in EOB. Here, there is only index, but in another book it might be Answer Key, Appendix or something else. So we can't choose one class to which we can apply margin.
To achieve proper effect we have to make some changes in recipes.

**First solution:**

change markup to

```
ol
  li.os-toc-preface
  ol
    li.os-toc-chapter
    li.os-toc-chapter
  li.os-toc-index
```
Then we can add margin to `ol` with chapters. This change might be problematic and it will affect webview. I don't recommend this.

**Second solution**

add additional class to first element from EOB

```
ol
  li.os-toc-preface
  li.os-toc-chapter
  .
  .
  .
[...]

PR #3402, opened 2021-01-21
Fix style for Toc

wilkus27:
Fix style for TOC in American Government based on new specs: https://github.com/openstax/cnx-recipes/issues/3154#issuecomment-762965934


ajluk (approved):

kswanson33 (changes requested):
~~I think the css output for sociology should be included in this change as well, but aside from that it looks good!~~ Disregard, just noticed that sociology doesn't have units.

Issue #3405, opened 2021-01-21
american government - blockquotes


lwtchu:
Documentation #3276

MartaSztuka:
##### PASSED![Screenshot from 2021-01-27 10-40-19.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0ea0c01e-82f6-4189-b9f6-52f0504070c3)![Screenshot from 2021-01-27 10-39-38.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0dfe0b31-e77c-4320-a6aa-f3ed5d37da63)

Issue #3404, opened 2021-01-21
sociology - blockquotes


lwtchu:
Documentation #3276

MartaSztuka:
##### @lwtchu Do we want to add bottom margines? When there's none and blockquote is not one paragraph it look's like this:
 ![Screenshot from 2021-01-27 10-20-27.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/48680869-17b4-40ee-835e-4396ba2e7485)
##### PDF![Screenshot from 2021-01-27 10-21-10.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4ce212fe-8d71-4514-9a2d-e826670b5fc5)

ajluk:
@lwtchu I have checked the example from above that Marta posted and I'm also wondering if should add some top/bottom margins to blockquotes (I have reset them - set the to 0, they had set 1em top/bottom default margins before), because in cases like this when in `blockquote` tag there is just some text (not a paragraph or a list) it will look like this? What do you think?

lwtchu:
@ajluk if it's set to 0 does that mean it's just the normal line-height? if so, that's ok cause in the documentation i don't see any significant spacing.

ajluk:
@lwtchu Yes, that's right, it is just the normal line-height under the blockquote. Above it is the margin from paragraph before it.

lwtchu:
@ajluk i think so. i can close it. please reopen if there's any issues.

Issue #3277, opened 2020-12-22
psychology - blockquotes


lwtchu:
Documentation #3276

MartaSztuka:
##### PASSED![Screenshot from 2021-01-27 09-37-21.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/0cc94c6d-9394-4416-bada-78d59e70f44a)![Screenshot from 2021-01-27 09-37-58.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/7284af49-0f18-460b-8828-9767a81b32b0)

Issue #3276, opened 2020-12-22
cosmos theme - blockquotes


lwtchu:
![Screen Shot 2020-12-21 at 6 06 21 PM](https://user-images.githubusercontent.com/11299125/102833643-46277800-43b7-11eb-97de-ea69f99066f4.png)

Same font as surrounding text, margins on left and right are 0.25 inches in addition to the standard 1inch

ajluk:
@lwtchu This how it looks: 
![image](https://user-images.githubusercontent.com/55203598/105375996-90478780-5c09-11eb-9817-3f4b71632e72.png)
I have noticed accidentally that there also blockquotes in Sociology. Do you want to open a separate card for this book or I can do it? I have added the blockquotes style for this book as well within the same PR.

ajluk:
@lwtchu I have made an investigation about `blockquotes`, because me and Natalia have noticed that the blockquotes have default properties vaules as an html tag and Prince supports and uses them. There is no difference between styled and unstyled blockquote in PDF's. So I'm wondering if we should add the styles in the code just in case Prince will no longer supports the defaults (as it was done before)? Or should we reset the default values and only set the margins in the code? I have checked if I could reset them and in case of left and right margins yes, but top and bottom can still be seen in the pdf after reset. 
Here are the default values: 
 ```
 display: block;
  margin-top: 1em;
  margin-bottom: 1em;
  margin-left: 40px;
  margin-right: 40px;
```
What is you opinion?

lwtchu:
@ajluk could you explain the margins part? i'm not sure i quite follow 😓

ajluk:
@lwtchu Sure! Maybe the best way could be to explain it on the example of browser default properties values for blockquotes (this is how it looks in the html when I'm resetting them): 
![image](https://user-images.githubusercontent.com/55203598/105522127-24ceea00-5cdd-11eb-972b-f4fb75d3849e.png)

lwtchu:
@ajluk hmm if it's not too much trouble i think we should add the styles in the code in case it's no longer supported or if the default changes since we have specific values we already want.

does that make sense D:

ajluk:
Here is the blockquote in the PDF file with reset margin values: 
![image](https://user-images.githubusercontent.com/55203598/105523818-2d282480-5cdf-11eb-90da-237784eaabe8.png)

ajluk:
[...]

PR #3406, opened 2021-01-21
style blockqote cosmos 

ajluk:
Issues: #3277, #3404, #3405, #3276


styles/designs/cardboard/parts/_blockquotes-components.scss
```diff
@@ -2,7 +2,7 @@ $Blockquote__Container: (
     _name: "BlockquoteContainer",
     _subselector: "",
     _properties: (
-        margin-left: 0.25inch,
-        margin-right: 0.25inch,
+        margin-left: 0.25in,
+        margin-right: 0.25in,
```
ajluk, line 6: I have noticed there has been a typo made in the cardboard blockquote component margin property value unit as you can see above. I have regenerated and checked Business Ethics pdf file after I have fixed it and it hasn't changed the  styling, so I haven't created the separate issue for this fix, but let me know if you think It will be safer to do it: 
image

styles/design-settings/cosmos/_design.scss
```diff
@@ -11,19 +11,19 @@
...
+@import 'designs/cosmos/parts/toc-shapes';
+@import 'designs/cosmos/parts/titles-shapes';
+@import 'designs/cosmos/parts/lists-shapes';
+@import 'designs/cosmos/parts/figureimage-shapes';
+@import 'designs/cosmos/parts/notes/notes-iconned-shapes';
+@import 'designs/cosmos/parts/notes/notes-boxed-shapes';
+@import 'designs/cosmos/parts/exercises-shapes';
+@import 'designs/cosmos/parts/link-shapes';
+@import 'designs/cosmos/parts/index-shapes';
+@import 'designs/cosmos/parts/learning-shapes';
```
wilkus27, line 26: I like this change :relaxed: It's more consistent now.

styles/designs/cardboard/parts/_blockquotes-components.scss
wilkus27: Hmm interesting :thinking: I checked it in `-baked.scss` file and I noticed that there are such additional  default specs for blockquotes elements:

image

Maybe there are similar default specs added by Prince. That's why margin left and right don't work here (even they have proper values now). They might be overwritten by this default settings.

wilkus27 (changes requested):

kswanson33 (approved):
Agree with the naming change request, but looks good otherwise!

styles/designs/cardboard/parts/_blockquotes-components.scss
ajluk: Yes, thank you Natalia! As we discussed I have checked that Prince also uses the default properties values predefined for html `blockquote` tag. I have reset the values to make Prince accept our own values (I asked Larissa if would like to do that to be sure that Prince uses values that give him).

styles/books/american-government/book.scss
```diff
@@ -79,7 +79,8 @@ $bandColor: #DCB83D;
 @include use('Figure', 'cosmos:::FigureShape');
 @include use('UnitTitle', 'common:::Hidden');
 @include use('ChapterIntro', 'cosmos:::ChapterIntroShape');
-@include use('LearningObjectivesModule', 'cosmos:::LearningObjectivesModuleShape');;
+@include use('LearningObjectivesModule', 'cosmos:::LearningObjectivesModuleShape');
+@include use('Blockquote', "cosmos:::Blockquote");
```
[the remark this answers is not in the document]
ajluk: Yes, good point! I have also changed the shape name in the previous theme for consistency.

kswanson33 (approved):

wilkus27 (approved):

Issue #3243, opened 2020-12-21
american government - table styling


lwtchu:
Documentation #3185 

Uses type 2 - top-titled

I added the top-titled class to the tables on katalyst01

MartaSztuka:
##### PASSED![Screenshot from 2021-01-26 15-23-40.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/5c2cb6d5-3235-4e53-93b0-19b739130540)

PR #3407, opened 2021-01-22
style top titled tables american government

ajluk:
Issue: #3243


kswanson33 (approved):
I can't speak to the recipe update, but everything else looks good :smile:

Issue #3279, opened 2020-12-22
psychology - connect the concepts


lwtchu:
- class: `class="connect-the-concepts"`
- uses #3193 
- title: Connect the Concepts
- has `<title>`
- icon: [cosmos-connected.svg](https://drive.google.com/file/d/1F3PAuu9LMu-uiamqy_t-qODiVgNniFNL/view?usp=sharing)
- color: B2: `#0074BC`
- contains paragraphs, figures, lists, etc - please check to make sure styling of all elements trickle down to the note

MartaSztuka:
##### PASSED![Screenshot from 2021-01-27 11-24-46.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/32ec406e-9fbe-4daf-8451-e4f3617fb045)

PR #3411, opened 2021-01-22
Style connect the concepts note for psychology

kswanson33:
#3279 
image


styles/books/psychology/book.scss
```diff
@@ -10,6 +10,25 @@ $bandColor: #DCB83D;
 @import '../../design-settings/cosmos/_settings.scss';
 @import '../../designs/cosmos/parts/pdf/_folio.scss';
 
+// Iconned Notes
+@include add_settings((
+  ConnectConcepts: (
+    _selectors: ('.connect-the-concepts'),
+    // 'Container:::box-decoration-break': slice,
```
ajluk, line 17: @kswanson33 The commented property from above is needed in this type of note to prevent adding bottom borders on every page when the note splits to the next pages.

ajluk (changes requested):
Everything looks good beside that the `box-decoration-break: slice` property is needed in this type of notes (with borders).

styles/books/psychology/book.scss
kswanson33: Thank you! I've added it back in.

ajluk (approved):

wilkus27 (approved):

Issue #3330, opened 2021-01-11
Extra space after assessment questions (business-ethics style)


yblock:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [x] styles [ ] both [ ] i don't know

There is extra space after the questions in the collated Assessment Questions sections. In the old style pdf the space was significantly smaller. 
![Screen Shot 2021-01-11 at 12.16.37 PM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/da6d5d38-3fba-4230-94c4-c3cc3a3b874f)

# Acceptance Criteria
There should be less space after the question like it used to be:
![Screen Shot 2021-01-11 at 12.22.47 PM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/e883b815-4a6a-4183-97fc-06f9feaa7d80)

# Bug Location(s)
collection ID: col25722
server: prod
chapter/section:  1.1 (all collated assessment questions)
surrounding text: Which of these concepts relates to utilitarianism?

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
Ideal to get this out in Februrary since we're trying to get Intellectual Property to publish at that time (uses same recipe/style). 
This shuld be a relatively easy fix though

lwtchu:
@wilkus27 @ajluk can y'all take a look at this one - it should be p simple. if it's not, please let me know.

ajluk:
@lwtchu I have removed the bottom margin for the paragraphs in exercises and this is how it looks now: 
![image](https://user-images.githubusercontent.com/55203598/106595053-45a40480-6553-11eb-8ace-c582996accc7.png)

lwtchu:
@ajluk nope this looks good thank you

ajluk:
@MartaSztuka During testing this issue, please check also Answer Key in one additional book from cardboard theme which have it: Accounting vol 1.

MartaSztuka:
#### PASSED
##### Business Ethics![Screenshot from 2021-02-08 11-16-00.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/9f5af1b2-41a0-4725-868d-7b670238ad9a)
[...]

PR #3453, opened 2021-02-02
Remove margin in exercises paragraps cardboard

ajluk:
Issue:  #3330


styles/designs/cardboard/parts/_exercises-components.scss
```diff
@@ -61,10 +61,20 @@ $Exercise__ProblemSolution__Divider: (
...
+
 $Exercise__ProblemSolution__LastPara: (
   _name: "ProblemSolutionLastPara",
   _subselector: " > p:last-child",
   _properties:(
     margin-bottom: 0,
   )
 );
+
+
```
wilkus27, line 80: Can we use existing component? I think it makes the same change. It was added only to Answer Key shape. As I remember, I used `last-child` pseudo selector just in case that there is more then on paragraph in problem/solution.

wilkus27 (changes requested):

styles/designs/cardboard/parts/_exercises-components.scss
ajluk: @wilkus27 as we discussed I have remove one of the paragraphs used in exercises/answerkey. I choosed the one with pseudo selector `:last-child` bacause after checking in the Answer Key in Business Ethics it appeared that there is only one para for each solution so nothing has changed.

wilkus27: Yes, I think it's a good idea :slightly_smiling_face: :+1:

kswanson33 (approved):

Issue #3358, opened 2021-01-13
economics - learning objectives


lwtchu:
Documentation #3175
Version 3

ajluk:
![image](https://user-images.githubusercontent.com/55203598/106908071-3a8dd780-66ff-11eb-9de0-e428b9babc45.png)

MartaSztuka:
##### PASSED![Screenshot from 2021-02-08 13-19-27.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/02b8d214-f343-404a-8e3c-e3493488da45)

PR #3467, opened 2021-02-04
style learning objectives principles economics

ajluk:
Issue: #3358


styles/design-settings/cosmos/parts/_learning-settings.scss
```diff
@@ -20,6 +20,9 @@
         AbstractContainer: (
             font-family: (_ref: "typography:::titleOption2Font"),
         ),
+        Paragraph: (
+            font-family: (_ref: "typography:::titleOption2Font"),
+        ),
```
ajluk, line 25: I have added Paragraph to this shape as well, because in Principles of Economics the markup is little different than it was for other books which have used this shape. Previously there was just a text node in the place of para. Let me know if I should rather create one shape for LO in modules instead and add different selectors to the books.scss files.

wilkus27: @ajluk I baked localy Economics to see this markup and I noticed that Learning Objectives is missing title. It should be added by this config: `$Config_abstractTitle: "Learning Objectives";` in `config.scss` file.

About markup, in carnival there was similar problem and there we decided to make changes in content, that is to wrap text into the paragraph in each book. BUT, it was necessary because paragraph had different style then list from below. Here we don't have such problem. I think it will be good to contant with Larissa and ask her if we want to have consistant markup. If not, I think your change is ok. I mean adding additional component to shape.

wilkus27 (changes requested):

styles/design-settings/cosmos/parts/_learning-settings.scss
ajluk: @wilkus27 Thank you! I don't know why I haven't noticed the missing title :sweat_smile:

kswanson33 (approved):

LostMojo (approved):

wilkus27 (approved):

Issue #3414, opened 2021-01-22
economics - key concepts and summary


lwtchu:
Documentation: #3209
Title: Key Concepts and Summary
class: `class="summary"`

MartaSztuka:
##### PASSED![Screenshot from 2021-02-09 10-55-23.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/cefa0226-391c-473e-83a9-625fb30c170f)

PR #3475, opened 2021-02-05
Style Summary in Economics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3414


styles/books/economics/book.scss
```diff
@@ -13,6 +13,7 @@ $bandColor: #DCB83D;
 
 // Settings
 @include use('BookRoot', "common:::BookRoot");
+@include use('Link', 'cosmos:::LinkShape');
```
Stackblocks, line 16: Is including the `cosmos:::LinkShape` the only thing that needed to be done to style the summary?

wilkus27: @Stackblocks Yes :+1: Summary contains only header and links to sections:

image

In this theme we don't have to create separate settings for each module. All headers from EOC are styled using one setting and one shape, which are added to the book in PR for Key Terms. Since I base this branch on branch for Key Terms, headers are already styled.

ajluk (approved):

Stackblocks (approved):

Issue #3482, opened 2021-02-08
table element alignment not consistent 


yblock:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [ ] styles [ ] both [ ] i don't know

![Screen Shot 2021-02-08 at 11.25.47 AM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/1f27b466-f33d-4aea-9e69-b2a35597de13)

https://openstax.org/books/microbiology/pages/17-1-physical-defenses#fs-id1172100638917

# Acceptance Criteria
LC: Can we make sure the padding is added to the cells even if there's a more rows attribute to the table?

# Bug Location(s)
collection ID: col12087
server: production
chapter/section: 17.1 
surrounding text: provides an overview of the nonspecific defenses

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

yblock:
Possibly only happens when the "morerows" attribute is used on a an entry in CNXML to merge vertical cells.

yblock:
Here is an example of a table that is coded similarly where it is not happening. 
https://openstax.org/books/microbiology/pages/17-2-chemical-defenses#fs-id1172102098156

MartaSztuka:
@wilkus27 Tables look ok, but caption is unstyled![Screenshot from 2021-02-25 10-22-00.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/233f7490-0b02-4da8-a9ba-16e78bec3af9)

wilkus27:
@MartaSztuka I checked and I think it's ok. Here are specs for Carnival: https://github.com/openstax/cnx-recipes/issues/1080. Screenshot in description is from OER.export. @lwtchu can you confirm? :slightly_smiling_face:

lwtchu:
@wilkus27 i think the font for the table number should be muli black?

wilkus27:
Hmm I noticed two interesting things :smile: First, in carnival we don't specify font family for table caption anywhere. So it's inherited from the root ( it's `IBM Plex Serif`). Maybe it's a regression from some cleaning :thinking: I don't know. Maybe settings for captions were removed accidentally. Second, Muli font doesn't exist anymore on fonts google. It's changed to similar one Mulish: https://fonts.google.com/?query=muli&selection.family=Ovo&sidebar.open=true

[...]

PR #3535, opened 2021-02-18
Fix paddings in tables in Carnival

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3482


styles/designs/carnival/parts/_table-components.scss
```diff
@@ -84,6 +84,7 @@ $Table__Data--BottomBorder: (
     _subselector: " > td",
     _properties: (
         border-bottom: enum('ValueSet:::OPTIONAL'),
+        padding: table-v-spacing(1),
```
wilkus27, line 87: I think this issue is a regresion from these changes: https://github.com/openstax/cnx-recipes/pull/1921/files 

image

There was added pseudo selector `:not(:only-of-type)` so only of type `td`s lost their styles.

ajluk (approved):

Stackblocks (approved):

Issue #3503, opened 2021-02-11
economics - implementing narrow-table class


lwtchu:
![image](https://user-images.githubusercontent.com/11299125/107663237-c0a2a480-6c50-11eb-9136-586f016a058e.png)

I added the class `narrow-table` to the table in the screenshot.

Since Prince does not support the caption-side element yet, I think we should see if this could just set a minimum width for the table. 

Can we have it be around 1/2 of the page? I'm not sure how much v-grid that is maybe 8?

Also when we implement it'd be good to answer these questions
- can we implement this library wide? it seems like a generic enough element to utilize across the library
- what would it take for an existing book right now to use this class

ajluk:
@lwtchu I have added the min-width: 20% to the container (div.os-narrow-table-container) not to the table so the caption will have the same width as the table and it works in html, but not in the pdf. The effect of min-width, max-width properties on tables are undefined in CSS and the container is set to `display:table. So I have tried `width: 100%` and it works then. Here is how it looks in pdf: 
![image](https://user-images.githubusercontent.com/55203598/108382897-97ae8080-7209-11eb-9b61-b6d9d059d260.png)
About implementing this solution library wide id depends on how many classes these tables will use. If only default tables will use it or also top-titled. But first to answer this question we need to spot all of the tables in the content to add the class where needed. I have also added a small fix to the recipes to create container class from this class and now all of the books can use it. What concerns me is that if this solution can be applied to all of the similar tables if it needs to be `width` set for them not `min-width`.

ajluk:
@lwtchu I have tried another solution (using `display: flex` on the container and `max-width: 25%` on table caption). And this is how it looks then: 
![image](https://user-images.githubusercontent.com/55203598/108498950-e281d400-72ad-11eb-8934-5f78840e7382.png)
Then table remains in its own size and if the caption text will be longer it will wrap. Here the concern is that the flex property could display not properly in exercises where we use `display: table` on problem containers and in the tables which will be very long and need to break to the second page - then the flex property could move the whole table to the next page, but it neds to be checked as well if will find such table.

lwtchu:
@ajluk are the two options right now either 1. set a min width or 2. have the caption flex?

ajluk:
@lwtchu Yes, but **option 1** is using `width: 20%` not `min-width` and `display: table` on the table container (`os-table`) and in **option 2** the caption is using `max-width: 25%` and the table container `display: flex`.

[...]

PR #3536, opened 2021-02-18
set width narrow table container

ajluk:
Issue: #3503


recipes/books/_common-stuff.scss
```diff
@@ -78,6 +78,9 @@ $Config_Table_Classes: (
   (className: 'column-header'),
   (className: 'text-heavy'),
   (className: 'text-heavy', secondClassName: 'top-titled'),
+  (className: 'narrow-table'),
+  (className: 'top-titled', secondClassName: 'narrow-table'),
+  (className: 'unstyled', secondClassName: 'narrow-table'),
```
wilkus27, line 83: Do you think we should add this change also for polish books? There is separate `common-stuff` file for them? Or it's unnnecessary?

wilkus27 (approved):
I left just one question :slightly_smiling_face:

brittany-johnson (approved):

Issue #3315, opened 2021-01-06
Extra spacing after some PDF paragraphs


yblock:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [x] styles [ ] both [ ] i don't know

In the business-ethics style.
As you can see in the screenshot below, some paragraphs like the first and second in this example have extra space after them. I do not see a reason for this in the cnxml. It seems to only happen when a paragraph contains a footnote inside of it.
![Screen Shot 2021-01-06 at 11.28.58 AM.png](https://images.zenhubusercontent.com/5c5227f6f3a9152a2050e821/59eb9c15-5111-44ba-aac8-507120e84c39)

# Acceptance Criteria

# Bug Location(s)
collection ID: col10001
server: content02
chapter/section: 1.1
surrounding text: Indeed, the Founders viewed intellectual property rights

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
I think this might be due to the reference being moved. Adding this to b.b I don't think it's a super complicated fix and shouldn't be a deal breaker for publishing.

lwtchu:
Moving to c.c. since it's still chillin in the backlog

lwtchu:
@wilkus27 @ajluk can y'all take a look at this one - it should be p simple. if it's not, please let me know.

wilkus27:
@lwtchu I did an investigation and found similar problem here: https://www.princexml.com/forum/topic/860/hiding-footnote-call-footnote-markers. Additional space appears because we hide footnote calls autogenerated by Prince and we are using our own tags. As you can read on the page from the link:

![image](https://user-images.githubusercontent.com/40228854/106608291-80626880-6564-11eb-854c-c9b3f7d262ae.png)

There are three advices how this problem could be resolved, but it might be complicated and time consuming. In this case, do we want to fix this problem? If yes, what is the priority and when we should start working on that?

lwtchu:
Interesting - I don't think this warrants that big of an effort right now. I'll put this in the backlog for right now. It's not ideal...but we have a bunch of higher priority things to do. Thank you for looking into this :)

yblock:
This will be an issue in every book with footnotes. Here is an example from Microbiology 22.1, searchable text "The lower respiratory tract, by contrast, is scantily populated with microbes"

[...]

PR #3540, opened 2021-02-22
remove space after para with footnote reference

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3315


wilkus27:
This is a huge change :thinking: It might cause some regressions. I'm not sure what would be the best way to test it. Please, let me know if you have any ideas :slightly_smiling_face:

lwtchu:
@wilkus27 i think we should test in the original affected book, one footnote from each theme, and one footnote in the preface, table, regular para, and appendix? what do you think?

wilkus27:
For footnotes I think it's ok. We should cover most of the cases with these tests. :relaxed: 

But within this PR I also moved import of `reset` file above the `common-styles` file, because otherwise styles from `reset` overwrite those from the second file:

image

It's hard to predict what regression this change could cause :crying_cat_face:

lwtchu:
@wilkus27 hmmm okay - i say for the english openstax books at least, we will regression test before we release and once we get a QA person in we can coordinate a bigger regression testing cycle. do you think the polish books could wait? or do you want to schedule in a round of testing for them?

recipes/mixins/_modify.scss
```diff
@@ -1251,3 +1251,16 @@
     }
   }
 }
+
+// tag paragraphs that contain footnotes references
```
LostMojo, line 1255: Thank you for leaving notes, this would be very hard to sort through otherwise

LostMojo (approved):

recipes/mixins/_modify.scss
LostMojo: Approved, given that tests work out okay.

Stackblocks (approved):

Issue #2878, opened 2020-10-19
Clean up: Refactor lists shapes in corn theme


ajluk:
Based on Natalia's suggestions from this comment (https://github.com/openstax/cnx-recipes/pull/2798#pullrequestreview-509162030): 
- create two shapes: for lists in exercises from one column sections and from multi column sections
- separate components for lists from one column layout and multicolumn layout
- define settings in book.scss file not in lists-settings.scss if needed

brittweinstein:
@ajluk is this for corn or carboard theme?

ajluk:
@lwtchu could you advise how many books (and maybe which of them) Marta should test within this issue? These changes affects all of the books from corn theme. I think Collega Algebra Coreq should be checked because there are exercises in coreq sections present and maybe Marta could pick one from dev-math and calculus books? What do you think?

@MartaSztuka During testing this issue circled lists with tokens needs to be checked if they still look ok (it's a clean up issue) in section `coreq` in College Algebra Coreq, section Exercises, EoC exercises, EoB Answer Key, Try It notes, Examples in the books from corn theme.

lwtchu:
@ajluk yeah that sounds good. college algebra coreq, elementary algebra and then calculus vol 2?

ajluk:
@lwtchu Thank you! Yes I thin it should be ok.

MartaSztuka:
#### PASSED
[...]

PR #3547, opened 2021-02-23
clean circled lists shapes corn

ajluk:
Issue: #2878


styles/designs/corn/parts/_lists-shapes.scss
```diff
@@ -275,6 +275,82 @@
...
+                    _components: (
+                        $UnorderedList--FirstLevel,
+                    )
+                )),
+            )
+        )),
+    )
+));
+
+
```
ajluk, line 353: I have created two lists shapes used in exercises one for one column layout and second for 3 column layout. The  `ExercisesInOneColumnListsShape` covers also lists in Try It notes and Examples which contains exercises lists as well.
These shapes uses different settings for circled lists. To make this clear and usable I have also created separate circled lists components for both shapes.

styles/design-settings/corn/parts/_lists-settings.scss
```diff
@@ -20,160 +11,46 @@
...
-        ),
         CircledListToken: (
             color: (_ref: 'colorMap:::circledListTokenColor'),
-            margin-right: h-spacing(1),
         ),
     ),
 ));
 
-// Used in Statistics books, where the exercises and solutions are in one column and in College Algebra Coreq for Coreq Skills exercises which are in one column as well
+// Used for corcled lists in Try It Notes and Examples in all of the corn theme books, in Statistics books, where the exercises and solutions are in one column and in College Algebra Coreq for Coreq Skills exercises which are in one column as well
```
Stackblocks, line 20: "corcled" should be "circled"

Stackblocks (approved):

styles/output/additive-manufacturing-pdf.css
```diff
@@ -1642,7 +1642,8 @@ a[role="doc-noteref"] {
 
 [data-type="page"] > section > .circled {
   list-style-type: none;
-  display: inline; }
+  display: inline;
+  padding-right: 0.5rem; }
```
wilkus27, line 1646: Is it intentional to add padding? The same change is in Statistics.

styles/design-settings/corn/parts/_lists-settings.scss
```diff
@@ -20,160 +11,46 @@
...
 @include add_settings((
     ExercisesProblemLists: (
         _selectors: ('[data-type="problem"]'),
```
wilkus27, line 23: Previously it was added only in Statistics and in College Algebra Coreq to one section, which was specified in book.scss file: 

image

So it was ok to use only `[data-type="problem"]` selector here. Now, when we apply these settings to books which contain multicolumn sections this selector is too general. Some of the properties might be overwritten.

wilkus27 (changes requested):

styles/design-settings/corn/parts/_lists-settings.scss
ajluk: @wilkus27 I have checked and it appears that is not overwriting the different styles for the lists applied to sections with 3 column layout. I think that is because their selectors are more specific/strong, but thank you for pointing this out. I had it in my mind as well.

styles/output/additive-manufacturing-pdf.css
ajluk: @wilkus27 As we discussed the padding doesn't do any harm if the list in one column exercises, I'm not sure what was the the intention of adding it (probably there was a sibling to this list positioned next to it somewhere), but it was there before my changes from this PR, just not in the output pdf-css files for some reason. But thank you for pointing this out!

styles/design-settings/corn/parts/_lists-settings.scss
wilkus27: Hmm @ajluk I think it's still too general :crying_cat_face: Some books contain section exercises, which use multicolumn shape. They are children of `[data-type="page"]`.

image

If I'm not wrong it's `[data-type="page"] > section.os-eos [data-type="problem"]`. So for books with section exercises we should probably use `[data-type="page"] > section:not(.os-eos) [data-type="problem"]` for `ExercisesProblemLists` .

ajluk: @wilkus27 Oh, yes that's a very good point! Thank you! I've got this in the back of my head and in the end I have forgot about those section guys :sweat:

wilkus27 (approved):

Issue #2180, opened 2020-04-08
CLEAN UP: Chapter Intro


helenemccarron:
- [ ] Break shapes out of the components: remove all `_components` from the components file and move them in the shapes file
 - [ ] Move lists to the lists.css file
 - [ ] Create a specific component for side chapter outline and a specific component for the full width chapter outline, 
   - Rename all `chapterOutline` components styling the side chapter outline to `sideChapterOutline`
   - Hard code all necessary values for those components, use `REQUIRED` for the font and the color as usual
 - [ ] Create a new set of components for the `fullWidthChapterOutline` reusing the `sideChapterOutline`
   - If any property is conflicting, create a new component rather than use `OPTIONAL`
 - [ ] Analyze `$ChapterIntroContentHeader` and break it down into as many components needed, move it to the title.scss file
 - [ ] Get rid of the `fullWidthBg` function, try to reuse a background component:
  - **Create common BG component**
```$splashElementsBackround: (
   _name: "BG",
   _subselector: '::before',
   _properties: (
       content: "''",
       background-color: enum('ValueSet:::REQUIRED'),
       position: absolute,
       height: 100%,
       width: enum('ValueSet:::REQUIRED'),
       margin-left: enum('ValueSet:::REQUIRED'),
       z-index: -1,
       box-sizing: border-box,
       padding: inherit,
       top: 0,
   )
);
```
 - **Rewrite the shape to be clearer & more readable:**

```('carnival:::ChapterIntroWithObjectivesShape', (
   _components: (
       $ChapterSplashFigureWrapper (
           $ChapterSplashFigure (
               $splashElementsBackround,
               $ChapterSplashMedia,
           ),
           $SplashCaption (
               $splashElementsBackround,
               $FigureNum,
               $FigureLabel,
               $FigureTitle,
               $Caption
           ),
       )
```
  - **Try nesting the BG in the figure and caption in the settings**

```
       SplashFigure: (
           BG: (
               background-color: (_ref: "colorMap:::chapterOutlineColor"),
               width: half-fullBleedsSize(8.5, in),
               margin-left: fullBleedsMargin(-1, in),
           ),
       ),
       SplashCaption: (
           BG: (
               background-color: (_ref: "colorMap:::chapterSplashCaptionColor"),
               width: fullBleedsSize(8.5, in),
               margin-left: fullBleedsMargin(-1, in),
           ),
       )

```
[...]

PR #3557, opened 2021-02-24
clean chapter intro shapes

ajluk:
Issue:  #2180


styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -1,42 +1,46 @@
...
+        caption-side: enum('ValueSet:::OPTIONAL'), // Used in Top Titled Tables only
+        prince-caption-page: enum('ValueSet:::OPTIONAL'), // Used in Top Titled Tables only
+        margin-left:  enum('ValueSet:::OPTIONAL'), // used in Figures only 
+        margin-right:  enum('ValueSet:::OPTIONAL'), // used in Figures only 
+        margin-bottom: enum('ValueSet:::OPTIONAL'), // used in Figures only 
+        font-size: enum('ValueSet:::OPTIONAL'), // used in Figures only
+        line-height: enum('ValueSet:::OPTIONAL'), // used in Figures only
+        position: enum('ValueSet:::OPTIONAL'),// Used in Chapter Intro
+        padding-top: enum('ValueSet:::OPTIONAL'),// Used in Chapter Intro
+        padding-bottom: enum('ValueSet:::OPTIONAL'),// Used in Chapter Intro
```
wilkus27, line 17: In next themes we decided that we use OPTIONAL/REQUIRED only to a few properties in a component. If there is a lots of them, then it's better to create a few different components which can contain hard coded properties. So maybe we can create here `$Caption__Container--Figures`, `$Caption__Container--Tables`, `$Caption__Container--ChapterIntro`. What do you think? In my opinion it will be more clear and readable.

styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -1,42 +1,46 @@
...
     )
 );
 
 $Caption__Title--Label: (
```
wilkus27, line 21: I think it should changed to `$Caption__Title__Label`.

styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -1,42 +1,46 @@
...
     _properties:(
-        font-family: enum('ValueSet:::OPTIONAL'), 
+        font-family: enum('ValueSet:::OPTIONAL'), // Used in Figures, Chapter Intro
         font-weight: bold,
-        color: enum('ValueSet:::OPTIONAL'),
-        display: enum('ValueSet:::OPTIONAL'),
-        margin-right: enum('ValueSet:::OPTIONAL'),
+        color: enum('ValueSet:::OPTIONAL'), // Used in Figures, Chapter Intro not in Tables
+        display: enum('ValueSet:::OPTIONAL'), // Used only in Figures
+        margin-right: enum('ValueSet:::OPTIONAL'), // Used in Figures, Chapter Intro not in Tables
```
wilkus27, line 39: I would split this also to a few components. There is too many optionals here.

wilkus27 (changes requested):
I'm leaving some suggestions for captions. Let me know what do you think :wink: There is a lots of changes here, so I don't finish my review yet. I'll come back to it tomorrow with fresh head :smile:

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -38,92 +34,99 @@ $ChapterOutline--FullWidth: (
...
+        font-size: font-scale(0),
+        margin-bottom: .5rem,
     )
 );
 
-$ChapterOutline__Objective--Learning: (
-    _name: "OutlineObjectiveLearning",
-    _subselector: ' > .learning-objective',
+$ChapterOutline__Objective--Text: ( // Component used only in Margin Chapter Intro 
+    _name: "ChapterOutlineObjectiveText",
```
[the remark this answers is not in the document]
wilkus27: Or `$ChapterOutline--Margin__Objective__Text`. I don't know what would be better :thinking:

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -38,92 +34,99 @@ $ChapterOutline--FullWidth: (
...
+    _subselector: ' > .os-text', 
     _properties: (
-        margin-left: enum('ValueSet:::OPTIONAL'),
+        margin-left: .5rem,
+        display: inline-block, 
     )
 );
+
+$ChapterOutline__Objective--Learning: empty_wrapper(OutlineObjectiveLearning, ' > .learning-objective');
```
wilkus27, line 122: As I see it's only in Full Width Chapter Outline. Maybe it could be `$ChapterOutline--FullWidth__ObjectiveLearning` or `$ChapterOutline--FullWidth__Objective__Learning` . Part `Objective--Learning` isn't proper, since Learning is not type of Objective, but its child. If I understand correctly from the shape: 

image

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -216,29 +177,42 @@ $ChapterSplash__Media: (
...
+            margin-left: fullBleedsMargin(-1, in),
+            z-index: -1,
+            box-sizing: border-box,
+            padding: inherit,
+            top: 0,
+        ),
+);
+
+$SplashCaptionBg: (
+        _name: "SplashCaptionBg",
```
wilkus27, line 200: Can we keep the naming convention for Splash Image components? It could be `$ChapterSplash__FigureBg` and `$ChapterSplash__CaptionBg`.

image

wilkus27: Hmm... there is component `$ChapterSplash__Figure` so maybe it should be `$ChapterSplash__Figure__Bg` or `$ChapterSplash__FigureBg` :thinking: I'm not sure which convention is better.

wilkus27 (changes requested):
Here are bunch of new comments :D Most of them are propositions. Let me know what do you think about it :wink: Still I have a few files to review from this PR, so I don't remove my assignment.

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -18,18 +18,14 @@ $ChapterOutline--Margin: (
     _subselector: ' > .os-chapter-outline',
```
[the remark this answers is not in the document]
wilkus27: As we discussed with @ajluk , since this PR is so huge already, we want to fix only problems like nested components, unnecessary functions and to split components containing too many OPTIONALs. There will be created separate issue for changing naming convention if we decide that we want to change something :wink: 

Please, leave your thoughts about it :slightly_smiling_face:

styles/output/ap-biology-pdf.css
```diff
@@ -1954,14 +1955,15 @@ a {
...
 .introduction > .intro-body > .os-chapter-outline > div.os-chapter-objective > a.os-chapter-objective > .os-text {
-  display: inline-block;
-  margin-left: 0.5rem; }
+  margin-left: 0.5rem;
+  display: inline-block; }
 
 .introduction > .intro-body > .intro-text > h2[data-type="document-title"]:first-of-type {
   float: left;
-  font-size: 12px;
+  font-family: IBM Plex Serif, serif;
```
wilkus27, line 1963: This is a new property here. Is it intentional?

styles/output/biology-pdf.css
```diff
@@ -1896,23 +1896,24 @@ a {
   padding: inherit;
   top: 0; }
 
-.introduction > .os-figure.has-splash > .os-caption-container > span.os-number {
+.introduction > .os-figure.has-splash .os-caption-container > .os-number {
   font-family: Muli;
+  font-size: 1.2rem;
```
wilkus27, line 1901: Here is the new property for this element. Is it intentional? If so, shouldn't be added also for `.os-title-label` and `.os-title`? Maybe we should also define line height?

wilkus27 (changes requested):

styles/output/ap-biology-pdf.css
ajluk: @wilkus27 previously there was 12px, but here I have just used function we dedicated to this: font-size: font-scale(0) so the value is 1 rem (in this case 12px as well): 
image

wilkus27: I mean font family :smile: It wasn't defined before.

ajluk: @wilkus27 It was there previously in the title settings file, but for some reason wasn't in the output. It doesn't change anything actually, because the font is the base font family.

styles/output/biology-pdf.css
ajluk: @wilkus27 Thank you! It is a typo.

styles/designs/carnival/parts/_chapterintro-components.scss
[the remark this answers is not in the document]
ajluk: Yes, that's a good idea, especially since this PR is also needed as a base PR for https://github.com/openstax/cnx-recipes/pull/3658. Thank you! I will open a separate PR with a description and links to you comments about changing naming convention after this one will be approved.

wilkus27 (approved):

Stackblocks (approved):

Issue #3567, opened 2021-02-26
Psychologia: implement headers


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/3559

MartaSztuka:
##### PASSED
![Screenshot from 2021-03-10 11-21-41.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/49a0774b-f8d0-457a-a395-c635fe39e95d)![Screenshot from 2021-03-10 11-21-30.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/b70e340d-6baf-46d2-943b-2251ac66c97e)

PR #3574, opened 2021-02-26
change header color in pl psychology

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3567


styles/design-settings/cosmos/_colors.scss
```diff
@@ -28,6 +29,7 @@
     tableBackgroundColor: (_ref: "scheme:::white"),
     tocTitleColor: (_ref: "scheme:::R1"),
     titleColor: (_ref: "scheme:::R1"),
+    titlePolishColor: (_ref: "scheme:::R4"),
```
ajluk, line 32: I have left similar comment on PR regarding ToC that maybe a better convention could be to add PolishPsychocology at the end of the color name: `titleColorPolishPsychocology`?

ajluk (approved):

styles/design-settings/cosmos/_colors.scss
wilkus27: @ajluk Yeah, you're right, it's not so much descriptive :thinking: Hmm.. What do you think about `titlePlPsychologyColor`? It's shorter and we keep `Color` part at the end like in other names.

ajluk: I think it's even better! :slightly_smiling_face:

brittany-johnson (approved):

Issue #3571, opened 2021-02-26
Psychologia: implement tables


wilkus27:
Documentation: https://github.com/openstax/cnx-recipes/issues/3563

MartaSztuka:
##### PASSED![Screenshot from 2021-03-05 11-46-44.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f61082c1-dfd6-4b80-a45e-352ea1df62eb)

PR #3602, opened 2021-03-02
style tables in pl psychology

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3571


styles/design-settings/cosmos/_colors.scss
```diff
@@ -4,6 +4,7 @@
     R1:     #C31427,
     R2:     #8B1504,
     R3:     #FFE5E5,
+    R5:     #501616,
```
wilkus27, line 7: `R4` is added in another PR

styles/designs/cosmos/parts/_table-shapes.scss
```diff
@@ -7,6 +7,7 @@
         map-merge($Table__Container, (
             _components: (
                 $Table__Title,
+                $Table__TopCaption,  //used only in Polish Psychology
```
wilkus27, line 10: This is the only difference between tables from american books, so I decided to add this component to existing shape rather then create the separate shape for Top Captioned Tables. It causes changes in all books but it shouldn't have any impact, since class of this component exists only in pl psychology. (Tables in pl psychology with fake title are created in different way then in american books, because of using new tool Adaptarr, so different class was needed.)

Let me know if you think that it will be better to create separate shape.

wilkus27:
### Recipe changes

Add class `top-captioned` to `$Config_Table_Classes` in common-stuff file for polish books.
It adds additional class `os-top-captioned-container` to div with `os-table` class that contains `table.top-captioned`.

styles/design-settings/cosmos/_colors.scss
ajluk: @wilkus27 there is also `R5` now on `e.e` :slightly_smiling_face: :, but I think you can change the color to `R6` during resolving conflicts.

styles/designs/cosmos/parts/_table-shapes.scss
ajluk: I thin it's ok since it can't do any harm.

ajluk (approved):

Stackblocks (approved):

Issue #3578, opened 2021-02-26
anthropology - learning outcomes


lwtchu:
This should be at the start of every module.

Documentation #3175 
Version 3
Uses `<section class="learning-objectives">` and standard markup

MartaSztuka:
##### PASSED![Screenshot from 2021-03-09 14-12-17.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/86b0b325-1871-4482-823d-7a33cad9ef66)

PR #3622, opened 2021-03-03
Learning objectives anthropology

ajluk:
Issue: #3578


styles/designs/cosmos/parts/_titles-components.scss
```diff
@@ -134,6 +134,7 @@ $H3--LearningObjectives: (
     line-height: line-height(0),
     font-weight: 700,
     text-transform: uppercase,
+    margin-bottom: 0,
```
ajluk, line 137: I added this property to prevent overwriting margin-bottom form page headers shape. It's hardcoded, so it is used in both learning shapes/settings but it does no harm since bottom margin is not needed in both of them.

wilkus27 (approved):

brittany-johnson (approved):

Issue #3586, opened 2021-02-26
anthropology - podcast


lwtchu:
- class: `media-podcast`
- uses #3235
- title: Podcast
- icon: [carnival-podcast.svg](https://drive.google.com/file/d/1eV_2vo9jUN0vB4MQv4VSgs5Lax1csDOB/view?usp=sharing)
- color: B2: `#0074BC`

wilkus27:
@lwtchu Here's the last one note:

![image](https://user-images.githubusercontent.com/40228854/109930454-fa068700-7cc7-11eb-9ba3-cc4503bd4555.png)

PR #3638, opened 2021-03-04
style podcast note

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3586


brittany-johnson (approved):

styles/books/anthropology/book.scss
```diff
@@ -127,12 +101,11 @@ $bandColor: #DCB83D;
 @include use('Figure', 'cosmos:::FigureShape');
 
 // Notes
-@include use('RealWorld', 'cosmos:::IconnedNoteWithSubtitleShape');
```
Stackblocks, line 130: Should `RealWorld` and `PolicyDebate` be getting removed as part of the `MediaPodcast` addition?

wilkus27: @Stackblocks Yes, these are notes from Sociology. Phil created basic files for Anthropology recipe and styles (to be able to add this book to `book.txt` file) and he just copied everything from Sociology. Most of the elements are the same in these two books but there are some differences between them (e.g notes).

wilkus27:
@Stackblocks Thank you for your comment :relaxed: I left the response for that. Since we have a little deadline for Anthropology and it's the last thing from this book I'm going to merge it without your approved. I hope you are ok with that :smiley_cat:  Let me know if you find something else, then I will make a fix for that in separate PR.

Issue #3601, opened 2021-03-01
DUE 3/19 Add World History (80%) to Cosmos Theme


lwtchu:
# Link to content
- This is a new book. Link forthcoming. Please do not work on this until you have talked to me.
- Test content: https://legacy-content03.cnx.org/content/col10004/latest/

# General Notes 
## Font links
- IBM Plex Serif https://fonts.google.com/specimen/IBM+Plex+Serif?query=ibm+plex+seri#standard-styles
- IBM Plex Sans https://fonts.google.com/specimen/IBM+Plex+Sans?query=ibm+plex
- Archivo https://fonts.google.com/specimen/Archivo?query=archivo
- Archivo Black https://fonts.google.com/specimen/Archivo+Black?query=archivo+black

## Theme assets folder
- https://drive.google.com/drive/u/0/folders/1r-6jPOZZNRL35fhbSgTn6cT5CiQMRn-2

# Design elements in the title

General things:
- [ ] Generic things to pull from other titles - Documentation: #3173 Working card: #3613
- [x] Table of Contents - Documentation: #3154  Working card: #3629
- [x] Headers - Documentation: #3164 Working card: #3630
- [x] Splash image - Documentation: #3171 Working card: #3631
- [x] Chapter Outline - Documentation #3175 Working card: #3608
- [ ] Section Learning Objectives - Documentation #3175  Working card: #3611
- [x] Introduction -  Documentation #3191 Working card: #3609
- [x] Figure & figure title/captions - Documentation #3183 Working card: #3612
- [x] Blockquotes - Documentation #3276 Working card: #3614 
- [x] Tables - Documentation: #3185 Working card: #3632 
- [ ] *NEW* Header row in middle of table - Working card #3665
- [ ] Footnotes: Documentation #3181 Working Card #3610
- [x] Exercises from Tutor - #3661 (if you see exercises populated and no blank spaces, close the card)
- [ ] Centered text - Documentation #3654 Working card: #3655 

### [Cosmos asset folder for icons.](https://drive.google.com/drive/folders/1r-6jPOZZNRL35fhbSgTn6cT5CiQMRn-2?usp=sharing)
Feature-like things:
- [x] In Their Own Words - Documentation #3194 Working card #3633
- [x] Dueling Voices - Documentation #3194 Working card #3635
- [x] Beyond the Book - Documentation #3194 Working card #3636
- [x] The Past Meets the Present - Documentation #3194 Working card #3637
[...]

PR #3666, opened 2021-03-09
Basic files for world-history

kswanson33:
#3601


Stackblocks (approved):

recipes/books/world-history/styleguide/page.note.beyond-book.snippet.xml
```diff
@@ -0,0 +1 @@
+<div/>
```
kswanson33, line 1: The build seems to insist that these snippet files exist, but @brittany-johnson says that they aren't used

LostMojo (approved):

recipes/books/_release.scss
```diff
@@ -725,6 +745,7 @@ $RELEASE_FLAGS: (
     "pl-u-physics",
     "precalculus-coreq",
     "pl-psychology",
+    "world-history",
   ),
 
   move-footnotes: enable-for(
```
ajluk, line 751: @kswanson33 You can alslo safely enable this flag and `modify-column-container` flag, the one flag we shouldn't enable because it needs additional changes is only  `seperate-label-text`.

ajluk (commented):
Looks good! I just left one comment about the release flags which can be also enabled.

Issue #3637, opened 2021-03-04
world history - the past meets the present


lwtchu:
- class: `past-present`
- uses #3194
- title: The Past Meets the Present
- has `<title>`
- uses alternate color schema 4 (yellow)
- should contain mostly paragraphs, images
- **NEW** `list class="critical-thinking"` - added new "Question section" specs on documentation card. Should be a line above the list and bold text for everything under the line.

MartaSztuka:
##### PASSED![Screenshot from 2021-03-18 11-33-33.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fbbc1839-9e38-4fb3-98de-f8449106d5b3)

Issue #3636, opened 2021-03-04
world history - beyond the book


lwtchu:
- class: `beyond-book`
- uses #3194
- title: Beyond the Book
- has `<title>`
- uses alternate color schema 3 (green)
- should contain mostly paragraphs, images
- **NEW** `list class="critical-thinking"` - added new "Question section" specs on documentation card. Should be a line above the list and bold text for everything under the line.

MartaSztuka:
##### PASSED![Screenshot from 2021-03-18 10-39-07.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/672c3a25-3fe3-44b1-a1d4-149fe473cb48)

Issue #3635, opened 2021-03-04
world history - dueling voices


lwtchu:
- class: `dueling-voices`
- uses #3194
- title: Dueling Voices
- has `<title>`
- uses alternate color schema 2 (red)
- should contain mostly paragraphs, images
- **NEW** `list class="critical-thinking"` - added new "Question section" specs on documentation card. Should be a line above the list and bold text for everything under the line.

MartaSztuka:
##### PASSED![Screenshot from 2021-03-18 11-29-03.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/e3641836-f755-493c-8aec-0604ce0d83e1)

Issue #3633, opened 2021-03-03
world history - in their own words


lwtchu:
- class: `own-words`
- uses #3194
- title: Profiles in Anthropology
- has `<title>`
- uses alternate color schema 1 (blue)
- should contain mostly paragraphs, blockquotes
- **NEW** `list class="critical-thinking"` - added new "Question section" specs on documentation card. Should be a line above the list and bold text for everything under the line.

MartaSztuka:
##### PASSED![Screenshot from 2021-03-18 11-34-21.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/628ea2f6-69e8-4794-a16a-cd6655771668)

PR #3675, opened 2021-03-10
Style notes for world history

### LINKED ISSUE #3634: world history - link to learning
- class: `link-to-learning`
- uses #3235
- title: Link to Learning
- icon: [carnival-linktolearning.svg](https://drive.google.com/file/d/1jm3US-spU6F3BwURhHDY5Emxpe5vmyOr/view?usp=sharing)
- color: G3: `#78B042`

kswanson33:
Resolves #3634, #3635, #3636, #3637, #3633


styles/designs/cosmos/parts/notes/_notes-components.scss
```diff
@@ -115,6 +115,20 @@ $Notes--Boxed__Subtitle: (
...
+    margin-left: 0,
+    padding-left: h-spacing(3),
+    padding-top: v-spacing(1),
+    border-top-style: solid ,
+    border-top-width: 0.1rem,
+    border-color: enum('ValueSet:::OPTIONAL'),
+    font-weight: 700,
+  )
+);
+
```
ajluk, line 131: @kswanson33 I was wandering if this list should be moved to `list-components` file? We usually move components like lists, table components, captions to its own components files. But this one is very specific. What do you think?

styles/output/world-history-pdf.css
```diff
@@ -956,4 +956,216 @@ div[data-type="composite-page"].os-eob {
   margin-bottom: 0.7rem;
   font-weight: 600; }
 
+<<<<<<< HEAD
```
ajluk, line 959: I guess the styles should be regenerated, because the conflicts hasn't been entirely resolved in this file.

ajluk (changes requested):
Everything looks good, but the conflicts havn't been entirely resolved in the output pdf files, so the styles should be regenerated.

styles/designs/cosmos/parts/notes/_notes-components.scss
kswanson33: The reason why I haven't put it in `list-components` is that the color of the top border depends on the parent note's title color, as in this example:

image

I don't think there's a way for this component to exist outside of the note shape and still based the border color on the parent note color.

styles/output/world-history-pdf.css
kswanson33: Thanks! I'll fix it when I do the final conflict resolution.

styles/designs/cosmos/parts/notes/_notes-components.scss
ajluk: I'm not sure if I understand correctly your question, but we can put such component to `lists-components`  and then import `list-components` in `note-shapes` (boxed notes shapes), then set the color in the `settings` or in this case in the `book.scss` file. But if you think it is better to keep this in the `notes-components` I guess it also acceptable, because of the specificity :slightly_smiling_face:

ajluk (approved):

wilkus27 (approved):

styles/designs/cosmos/parts/notes/_notes-components.scss
kswanson33: OH you are very smart. Thanks 😄

Issue #2169, opened 2020-04-07
CLEAN UP: Carnival title markup


helenemccarron:
All titles should be in titles.scss. The rule for creating titles is: 
Can I use a generic title without a million `OPTIONAL`s?
 - > If yes, use generic existing component
  - >if no, create a custom named title and use that. Add that custom title in the titles.scss along with the other similar titles (i.e `H2Title` would be followed by `H2ChapterOultine` etc.)

- [x] Remove divider in `appendixTitle` from the recipes since we are not using it. Check Appendix title markup for extra space (work may have already been completed)
- [x] Add comments to all levels to describe where to find them/what they do
- [x] Change the h3 in the h4 component in the recipes. Look for the H3s in the EOC that are being treated as H4s by the styles. Component name is `$H4--Link`
- [x] Refactor the appendTitle to be a generic `H1Number` component
- [x] Rename `MainNumber`, `MainDivider` and `MainText` to H1number, H1divider and H1text
- [x] Add h5 support to modules and wherever necessary

wilkus27:
Other things to clean up:

- [ ] create `PageTitlesShape` for titles from `[data-type="page"]` (Preface, regular module, Appendix)
- [ ] create `ChapterTitlesShape` for chapter title
- [ ] create `EocCompositePageTitlesShape` for titles from EOC composite pages, direct children of `[data-type="chapter"]`
- [ ] create `EocCompositeChapterTitlesShape` for titles from EOC composite chapters (they contain composite pages)
- [ ] create `EocCompositeChapterTitlesWithBorderShape` for EOC titles from Polish U-physics (they are styled differently then in American books)
- [ ] create `EobTitlesShape` for titles from EOB (Index, References, Answer Key)
- [ ] remove titles components from Answer Key shapes
- [ ] remove titles components from Appendix shape
- [ ] remove `GenericEOCShape`, `EOCDefaultCompoundComposite` and `EOCCompoundCompositeWithBorder` shapes from `eoc-shapes` file (they are replaced by EOC titles shapes listed above; new shapes are created in `titles-shapes` file)
- [ ] remove titles components from Index shape
- [ ] move `$Preface__Container` component to `titles-components` file and remove `preface-components` file (NOTE: all Preface files should be removed, but there are still lists components in Preface shape, which will be moved to lists shapes in another PR)
- [ ] remove titles components from Preface shape
- [ ] remove titles components from References shape
[...]

PR #3693, opened 2021-03-15
refactor titles shapes and components in carnival

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2169

Here are things that weren't included in an original issue, but I changed them: https://github.com/openstax/cnx-recipes/issues/2169#issuecomment-799338188


styles/books/pl-u-physics/book.scss
```diff
@@ -1,359 +1,336 @@
...
-  
-  $bandColor: #00819A;
-  
-  $ChapterIntroType: fullWidth;
-  
-  @import 'framework/framework';
-  @import 'design-settings/carnival/design';
-  @import 'design-settings/carnival/settings';
-  @import '../../designs/carnival/pdf/locale/pl/folio';  
-
```
wilkus27, line 15: I fixed also spacing in this file. Code wasn't aligned to the left. That's why it shows that everything here is changed.

styles/designs/carnival/parts/_appendix-components.scss
```diff
@@ -1,3 +1,6 @@
+// To discuss: keep appendix files for appendix glossary or remove them and style appendix glossary title in PageTitlesShape
+// and appendix glossary list in PageListsShape?
+
```
wilkus27, line 3: Please, let me know what is your opinion. Appendix Glossary is module that appears just in one book (Microbiology):

image

Now Appendix shape contains titles components (which I'm removing within this PR), lists components (which will be moved to lists shapes in another PR) and appendix glossary components. When all clean up things will be done we can do two things:
1. keep `AppendixShape` just with appendix glossary components, maybe rename it to AppendixGlossaryShape and inlude it only for Microbiology (old one should be removed from other book.scss files) 

or

2. remove all appendix files and add appendix glossary list component to `PagesListsShape` and title component to `PagesTitlesShape`.

Let know which solution is better. When we have a decision, I will create clean up issue for that.

styles/output/anatomy-pdf.css
```diff
@@ -841,149 +841,44 @@ nav#toc > ol > .os-toc-index > a::after {
   font-size: 1.2rem;
   line-height: 1.5rem; }
 
-.preface {
+.preface.preface {
```
ajluk, line 844: @wilkus27 There is a second preface class added, I guess it is by accident.

styles/designs/carnival/parts/_answerkey-shapes.scss
```diff
@@ -47,11 +44,8 @@
   _components: (
     map-merge($AnswerKey__Container, (
       _components: (
-        $H1,                                                           //ToDo: create separate shape for EOB titles, like in other themes
```
ajluk, line 50: @wilkus27 I guess this comment on the right could be removed, because I have seen that you already created the title shape for EoB part.

styles/output/anatomy-pdf.css
ajluk: I have checked out locally to your branch and this bug is really odd, because in the settings file only one selector is adder and the second one only appears in the output files (for all of the books), maybe regenerating output files could help with that :thinking:

styles/designs/carnival/parts/_appendix-components.scss
ajluk: I guess if the shape is needed only in one book, we can keep it, if we will remove it, we need to set the glossary title color to `OPTIONAL` or `REQUIRED` and then set it for every book even it's not needed there. But that is just a thought how to keep it simple and the second preposition is also ok, because it is consistent to what we do in the other themes and you can add just a comment that the title is used only in Microbiology.

ajluk (changes requested):
@wilkus27 Everything looks good beside the doubled `preface` class which I'm not sure where it came from. A huge work has been done here! :slightly_smiling_face:

styles/design-settings/carnival/parts/_preface-settings.scss
```diff
@@ -2,6 +2,6 @@
 
 @include add_settings((
   Preface: (
-    _selectors: (".preface"),
+    _selectors: (" "),
   ),
 ));
```
wilkus27: Selector is defined in `Preface__Container` component which I moved to `titles-components`. We are using it in `PageTitlesShape`. It's not a perfect solution. Selector should be defined in settings. But because we are going to remove this entire file (when clean up for lists be done) I decided to remove selector from here, rather then create `preface-components` file with different empty `Preface__Container` component.

styles/output/anatomy-pdf.css
wilkus27: Thank you @ajluk for noticing that. I made a fix and add some comment.
Generally this problem appeared because selector was defined in Preface settings (which are going to be removed) and in `Preface__Container`. We need class in this component, because we are using it in `PageTitlesShape`.

styles/designs/carnival/parts/_answerkey-shapes.scss
wilkus27: Yes, entire line is already removed.

styles/designs/carnival/parts/_appendix-components.scss
wilkus27: Thank you @ajluk! Yeah, you are right, first solution seems to be more simple and clear. @kswanson33 I saw you have done something similar for World History from Cosmos. It looks very good. I can create a card to apply your solution to Carnival theme.

ajluk (approved):

Issue #3694, opened 2021-03-15
psychologia: link out of the note 


wilkus27:
When link is very long it goes out of the note:

**Psychologia, module 12.5**
![image](https://user-images.githubusercontent.com/40228854/111178856-d4249080-85ab-11eb-850f-3eec1b0648b2.png)

wilkus27:
### Information for tester

Fix for that affect all books from all thems. Please test one or two more books to check if everything is ok.

MartaSztuka:
##### PASSED![Screenshot from 2021-03-18 15-01-08.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/fed2b5a2-cb96-4df5-86b2-911b8c3747d5)

MartaSztuka:
###### World History![Screenshot from 2021-03-19 12-17-00.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/d3a52527-05d3-4afb-8592-0bfa8a0fb5e8)
###### Introduction to Sociology![Screenshot from 2021-03-19 12-18-26.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/610384a8-1d97-439a-8174-d7fa4e9ff0df)

PR #3706, opened 2021-03-16
Wrap long links

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/3694


styles/designs/common/pdf/_common-styles.scss
```diff
@@ -1,5 +1,6 @@
 a[href^="http"]::after{
   content: " (" attr(href) ") ";
+  overflow-wrap: break-word;
 }
```
wilkus27, line 4: I made this change for library wide. If you think it's too dangerous and it might cause some regressions I can create component with this property and apply it to cosmos theme or just to Polish Psychology. Let me know what do you think about it.

ajluk: I guess there was an similar issue a while ago regarding Entrepreneurship, but I couldn't found it, so maybe we should keep it library wide just to cover it there as well if it wasn't resolved.

ajluk (approved):

styles/designs/common/pdf/_common-styles.scss
Stackblocks: I'd say that it's fine to keep Library Wide, we just need to have Marta do a spot check in one or two other books to see if anything is obviously breaking in an unexpected way. (Based on the code, I do not think this will be a problem)

Stackblocks (approved):

Issue #3665, opened 2021-03-09
world history - header row in middle of table


lwtchu:
Targeted by `class="header-row"`

Same style as #3185 header row - no side borders, just a band in the middle and bold the text.

There's one in Death from the Steppe: Genghis Khan and the Mongols

MartaSztuka:
@kswanson33 Table look's fine, but shouldn't the headers in middle row be  centered, as in the top row?
![Screenshot from 2021-03-19 12-00-11.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/e1d7712b-7c66-4285-8bdf-64c2382db0c1)![Screenshot from 2021-03-19 12-03-05.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c489fae0-acf5-40db-a2d1-c6605cffeff2)

kswanson33:
@MartaSztuka I believe that the header row will still follow whatever alignment is set with `data-align`. Since that attribute isn't set here, it defaults to left. Thoughts @lwtchu ?

lwtchu:
@kswanson33 that's fine. if they want it to be centered we can add a data-align attribute

MartaSztuka:
@kswanson33 @lwtchu if it's fine I'm closing the issue

PR #3715, opened 2021-03-17
style tables for world history

### LINKED ISSUE #3632: world history - tables
Documentation #3185

type 1 - use type 1

kswanson33:
Resolves #3632, #3665. Introduces new style: header row in the middle of a table


styles/designs/cosmos/parts/_table-components.scss
```diff
@@ -275,6 +277,33 @@ $Table__Data--LastRowLastChild: (
     ),
 );
 
+$Table__Row--MiddleHeader: empty_wrapper(TableRowMiddleHeader,  ' > tr.header-row'); // Used in World History
+
+$Table__Data--MiddleHeaderCell: (
+    _name: "TableDataMiddleHeaderCell",
+    _subselector: " > td",
+    _properties: (
+        color: enum('ValueSet:::OPTIONAL'),
```
ajluk, line 286: @kswanson33 Can I ask if is there a reason why all the values set in `table-settings` are optional?

styles/output/pl-psychology-pdf.css
```diff
@@ -1184,6 +1184,14 @@ blockquote {
   font-weight: 500;
   margin-bottom: 0rem; }
 
+.everyday-connection > .os-note-body > ul.critical-thinking {
```
ajluk, line 1187: I have noticed that these note styles with your recent changes with adding specific `critical-thinking` list have been added to `pl-psychology-pdf.css` file withnin this PR, I guess it is because all of the styles haven't been regenerated previously and pl-psychology has been accidentally missed?

ajluk (commented):
Looks good! I have just left two comments with questions. :slightly_smiling_face:

styles/output/pl-psychology-pdf.css
kswanson33: Yeah, that's correct, I had an incomplete list of the books in cosmos when I made changes in previous PRs. Hopefully I got them all now 🤞

styles/designs/cosmos/parts/_table-components.scss
kswanson33: I included settings for the `TableDataMiddleHeaderCell` for `DefaultTable`, but not for the other types (`FakeTitleTable`, `TopCaptionedTable`). This caused other styles that used these settings to break. 

I was thinking that since these other two table types don't have any `MiddleHeaderCell` examples, the added settings should be `OPTIONAL`. But now that you mention it, I could add the settings to the other two table types and change these properties back to `REQUIRED`.  What do you think?

ajluk: @kswanson33 Oh, now I got it! Thank you! In this case I think they should stay `OPTIONAL`, you could also ask Larissa if the other tables will have such cell. If yes we can set them to REQUIRED.

ajluk (approved):

Stackblocks (approved):

PR #3782, opened 2021-04-01
Create initial files for Computer Science

Stackblocks:
- Created subdirectories in the Styles and Recipes directories
- Added dummy lines to books.txt
- Copied over entrepreneurship as a base for the recipe and styles
- Generated the recipes / styles so they can be tracked
Re:3727


Stackblocks:
Error-ing for unknown reasons, going to rebuild on clean branch just in case.

Issue #3662, opened 2021-03-08
DUE March 26 - Add Finance (80%) to the Cardboard Theme


lwtchu:
# Link to content
-  https://legacy-content06.cnx.org/content/col10007/latest

# Design elements in the title

General things:
- [x] Generic things to pull - Documentation #2712 / Working card: #3673 
- [x] Table of Contents - Documentation #1740 / Working card: #3672
- [x] Splash image - Documentation #2642 / Working card: #3670 
- [x] Figure caption/title/number styling - Documentation #2643 / Working card: #3677
- [x] Headers - Documentation #2644 / Working card: #3676 
- [x] Chapter Outline & Introduction - Documentation #2655 / Working card: #3669
- [x] Section number treatment - Documentaiton #2710 / Working card: #3679 
- [x] Learning Outcomes - Documentation #2711 / Working card: #3671
- [x] Tables - Documentation #2741 / Working card: #3674
- [ ] **NEW** Table Caption styling - Documentation & Working card: #3678
- [x] Footnotes - Documentation #2773 / Working Card #3700
- [x] **NEW** Calculator buttons - Documentation & Working card: #3684 
- [x] Equations - Documentation #2894 / Working card: #3697
- [x] **NEW** Stepwise Lists (sort of new - to this theme I think) - Documentation & Working Card: #3701
- [ ] TBD Calculator table - hm let's not do this one right now
- [x] **NEW** Timeline table - Documentation and Working Card #3723 
- [x] **NEW** R code snippet (using highlight.js) - Working Card #3725
- [x] Grey table shading - Working card #3724
- [ ] Verify that exercises are pulled in from exercises.openstax.org (are there exercises or are there errors)

### [Cardboard asset folder for icons.](https://drive.google.com/drive/folders/1wsO8nLf7QJXj6YnB1pmjPle0hVYwMoyP)

Feature-like things:
- [x] Think It Through - Documentation #2737 / Working Card #3698 
- [x] Concepts In Practice  - Documentation: #2737 / Working Card  #3699
- [x] **NEW** Excel Spreadsheet  -  Documentation #3680 / Working Card #3681
- [x] Link to Learning - Documentation #2737 / Working Card #3682

End of Chapter Content
- [x] Key Terms - Documentation #2742 / Working Card #3702 
- [x] Summary - Documentation #2743 / Working Card #3703
- [x] CFA Institute - Working card: #3726
- [x] Multiple Choice - Working card: #3705
[...]

PR #3814, opened 2021-04-09
Add Finance Styling

LostMojo:
Documentation: #3662


LostMojo:
Yes, this is the whole finance book. Sorry about that

recipes/books/finance/_config.scss
```diff
@@ -0,0 +1,130 @@
...
+$Config_PartType_Chapter:  (outlineTitle: 'Chapter Outline');
+/*==========================================================
+  EOB
+==========================================================*/
+
+/* Answer Key
+*/
+
+$Config_TargetLabels : append($Config_TargetLabels,
+(selector: 'div[data-type="chapter"] > div[data-type="page"]:not(.introduction)', labelText: "LO ", labelCounter: counter(chapter) "." counter(section))
```
ajluk, line 124: @LostMojo If this config is needed in finance recipe not just copy/pasta form other book recipe, the flag `seperate-label-text` should be also enabled for finance.

styles/designs/cardboard/parts/_hljs-components.scss
```diff
@@ -0,0 +1,45 @@
+
+$Code__Container: empty_wrapper(CodeContainer, '');
```
ajluk, line 2: @LostMojo I can't see if the components is used anywhere, maybe the one below it was your first thought.

styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -55,3 +55,30 @@ $OrderedList--SecondLevel: (
...
+    _properties: (
+        margin-left: h-spacing(2),
+    )
+);
+
+$StepwiseList--CustomBullet: (
+    _name: "StepwiseListCustomBullet",
+    _subselector: " > li::marker",
+    _properties: (
+        content: '"Step " counter(list-item) ". "',
```
ajluk, line 80: @LostMojo Instead adding counter via the styles you could use ```$Config_CustomList: (
  (selector: '.stepwise', name: "stepwise", beforeLabel: "Step ", afterLabel: ". "),
);``` in the config.scss file and enabled related flag `custom-lists` for finance.

styles/designs/cardboard/parts/_hljs-components.scss
LostMojo: Thanks for catching that!

styles/designs/cardboard/parts/_table-shapes.scss
```diff
@@ -212,4 +258,10 @@
             )
         )),
     )
-));
\ No newline at end of file
+));
+
+@include create_shape('cardboard:::ColoredBgTableRowShape', (
```
ajluk, line 263: That's an interesting solution! But what concerns me that it can be overwritten by other styles if this component isn't included in the table shapes. I think for safety reasons it should be rather included in the table shapes that could have such functionality. What do you think?

styles/designs/cardboard/parts/_calculator-components.scss
```diff
@@ -0,0 +1,12 @@
+$Calculator__Button: (
+    _name: "CalculatorButton",
+    _subselector: "",
+    _properties: (
+        // display: block,
```
ajluk, line 5: I guess this comment could be removed.

ajluk (changes requested):
A lot of work has been done here! I left some comments here and there.

styles/designs/cardboard/parts/_table-shapes.scss
LostMojo: I'll go ahead and incorperate it into the larger table shape 😅

LostMojo: I originally put it separately so that other colors could be included with different shapes, but I don't see us using more than one color at the moment. Incorporating it into an existing shape will probably mean that the color will have to be an optional variable, so it won't break any past books, though.

LostMojo:
Fixes incorporated!

styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -55,3 +55,43 @@ $OrderedList--SecondLevel: (
...
+        flex-direction: row,
+    )
+);
+
+$StepwiseList--CustomBullet: (
+    _name: "StepwiseListCustomBullet",
+    _subselector: " > .os-stepwise-token",
+    _properties: (
+        font-weight: bold,
+        // display: inline-block,
```
wilkus27, line 85: I think it can be removed.

styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -55,3 +55,43 @@ $OrderedList--SecondLevel: (
...
+    )
+);
+
+$StepwiseList--Content: (
+    _name: "StepwiseListContent",
+    _subselector: " > .os-stepwise-content",
+    _properties: (
+        //display: inline-block,
+    )
+);
```
wilkus27, line 97: Can we remove commented line? If so, then this component will have no properties. In such cases we are using `empty_wrapper`:

`$StepwiseList--Content: empty_wrapper(StepwiseListContent, ' > .os-stepwise-content');`

styles/designs/cardboard/parts/_lists-shapes.scss
```diff
@@ -96,3 +96,18 @@
...
+                map-merge($StepwiseList--FirstLevel,  (
+                    _components: (
+                       $StepwiseList--CustomBullet,
+                       $StepwiseList--Content,
+                    )
+                ))
+            )
+        )),
+    )
+));
```
wilkus27, line 113: In general, we are using `--` in names to describe variations of some components. E.g tag `td` in tables :

```
$Table__Data--TopAligned,
$Table__Data--LeftAligned,
$Table__Data--RightAligned,
$Table__Data--Centered
```
When we want to describe different elements we should use `__`. Using example from tables:

```
$Table__Container
$Table__Head
$Table__Data
```

Sometimes it is hard to say if component is a variation or separate element :smile: Can you check names of all of your newly created components whether they are going with this convention? I think components from this file and `hljs_components` might need some fixes.

styles/designs/cardboard/parts/_table-components.scss
```diff
@@ -11,7 +11,16 @@ $Table__Container: (
...
+    _subselector: " > table:not(.timeline-table)",
+    _properties: (
+        width: 100%,
+        border-collapse: collapse,
+    )
+);
+
+$Table--Timeline: (
+    _name: "TableTimeline",
+    _subselector: " > table.timeline-table",
```
wilkus27, line 23: Here is a good example with proper use of `--` in name. :slightly_smiling_face: :+1: We have regular `Table` and its variation `Table--Timeline`.

styles/designs/cardboard/parts/notes/_notes-iconCondensed-shapes.scss
```diff
@@ -0,0 +1,14 @@
+@import 'notes-components';
+@import '../lists-components';
```
wilkus27, line 2: Do we need this import? Shape contains only notes components.

wilkus27 (changes requested):
@LostMojo I added some comments. Some small fixes are needed, but in general it looks good. A lot of work is done here :slightly_smiling_face: 

Beside problems from comments I noticed that output files for other books are not generated. Can you use `./script/run ./script/build-styles` command for this? 

Also when we change something in release flags we should generate `RELEASE_FLAGS.md` file. To do that run `./script/run ./script/release-flag-chart`.

recipes/books/finance/_config.scss
LostMojo: (added to seperate-label-text)

styles/designs/cardboard/parts/notes/_notes-iconCondensed-shapes.scss
LostMojo: We do not! good catch

wilkus27 (approved):

Issue #3822, opened 2021-04-15
sociology recipe - answer key display


lwtchu:
# Brief Description of Change and Screenshots
What kind of change is this? [x] recipe [ ] styles [ ] i don't know

if i do not have anything in the answer key, could there be an option to have it not display a header for that chapter see here for example, this is not ideal (go to answer key): https://rex-web.herokuapp.com/books/introduction-sociology-2e@15.8/pages/chapter-11?arc[…]%2Fopenstax.org%2Fapps%2Farchive-preview%2F20210412.215425

we are trying to balance the 3e and 2e changes for sociology (we re-enabled the answer key for sociology 3e #3663) and still utilize one recipe.

based on this thread: https://openstax.slack.com/archives/CH5514N57/p1618512976008200 

# Acceptance Criteria

# Content Location(s)
collection ID: col11762
server: production
chapter/section:  answer key
surrounding text:  answer key

lwtchu:
phil says it shsould be possible so that's..one good news.

brittany-johnson:
Answer key should not be present in soc 2e but present in 3e. 3e content can be found on `content02 col10084 `for testing

![Screen Shot 2021-04-20 at 1 53 56 PM](https://user-images.githubusercontent.com/24909007/115449040-f8ddda80-a1df-11eb-8bf6-a36d9111e43f.png)
![Screen Shot 2021-04-20 at 1 54 09 PM](https://user-images.githubusercontent.com/24909007/115449056-fbd8cb00-a1df-11eb-8159-d939b7b203ca.png)

lwtchu:
Need to test 1 2 3e when it goes to corgi

PR #3829, opened 2021-04-19
rm soc answer key section if not present

brittany-johnson:
related issue https://github.com/openstax/cnx-recipes/issues/3822


recipes/books/_generator.scss
```diff
@@ -726,6 +728,13 @@
     @include validate_typeOptional($elementType, string);
     @include modify_addFirstElementClass($selector, $elementType);
   }
+
+  @if release-flag-enabled(hack-rm-empty-answ-key) {
+    .os-eob.os-solution-container.no-solutions {
+      move-to: trash;
+    }
+  }
+  @include utils_clearTrash();
```
wilkus27, line 737: I'm not sure about this include. This change appears in all books. I'm wondering if clearing trash in earlier pass might have some unpredictable results :thinking: Maybe we can put it under `if` statement from `732` line? What do you think? Than it will have impact only on sociology which is going to be tested carefully in the near future. Let me know your thoughts :wink:

recipes/books/american-government/_config.scss
```diff
@@ -143,5 +143,5 @@ $Config_PartType_Table_CaptionContent   : (os-title-label: "Table ", os-number:
 $Config_SetTableCaption : (captionType: $CAPTION_TABLE, defaultContainer: caption, hasCaption: true, hasTitle: true, hasTopTitle: true);
 $Config_hasCitation: true;
 
-
+$Config_hasGroupedIntroBody: true;
```
wilkus27, line 146: Is it intentional? If so, I checked output file for American Government and it doesn't change anything there :thinking:

wilkus27 (changes requested):
I left two questions :slightly_smiling_face:

philschatz (approved):

PR #3842, opened 2021-04-29
create one shape for numbered lo

wilkus27:
Two developers were working on the same element (numbered Learning Objectives) in two different books. Since there were discrepancies in styles I left both shapes during resolving conflicts. After checking with Larissa which styles are proper one, I'm creating in this PR one shape common for two books.

Discussion from slack: https://katalysteducation.slack.com/archives/CBW9DFKGF/p1619612730003500


ajluk (approved):

brittany-johnson (approved):
🎉

Issue #3855, opened 2021-05-07
Algebra & Trig - Bold and Italic variable causing line break 


marcjeannott:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [x] styles [ ] both [ ] i don't know

![Screen Shot 2021-05-07 at 9.03.17 AM.png](https://images.zenhubusercontent.com/5b572d7e0e34d640605a661c/9b0ace52-585c-4067-8f07-5a5e82c53138)

This is from Corgi PDF I generated on May 6th. I believe this issue is similar to [Issue 3394](https://github.com/openstax/cnx-recipes/issues/3394)

The different variables seem to be making new lines. This may be because each has two emphasis tags (a bold and an italic) however this same thing is used in the exercises without breaking into new lines. 

# Acceptance Criteria
The variables aren't causing line breaks. 
# Bug Location(s)
Algebra and Trig PDF - 10.8 Section Exercises Folio 1013
collection ID: col11758
server: Prod
chapter/section:  10.8
surrounding text: For the following exercises, use the given vectors to compute

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
This should've been fixed with https://github.com/openstax/cnx-recipes/commit/91345794ab2edd00f74ac5ea88fa308b0f887b56  -- can we make sure it's fixed in the precal styles file also?

lwtchu:
removing the effect="bold" will also fix it but karina and anna (karinanna :joy:) have also done the fix. need to talk to a PHILosopher of content to either validate the content better or transform it or...something else

lwtchu:
I think this got merged - you can test when it goes to staging on corgi maybe :eyes:

Stackblocks:
Displaying OK in updated CORGI output

![image](https://user-images.githubusercontent.com/12678235/119554367-31824e00-bd62-11eb-9b98-3ce310315447.png)

PR #3869, opened 2021-05-18
Algebra and trigonometry: linebreak in vectors

KarinaMendez2:
for: https://github.com/openstax/cnx-recipes/issues/3855


ajluk:
@KarinaMendez2 I have checked if the change you made will break the styling of `p` titles in exercises becuase they utilize the same selector and it's not :relaxed: Here are the titles that utilize also the same selector:
image

Also please remember to regenerate the styles for all the books within the general script: `./script/run ./script/build-styles`, because right now in your PR only styles for Precalculus has been generated and the change also concerns the other math (corn theme) books. 

@lwtchu This problem occurred again because the markup has been changed on production in Algebra Trigonometry - attribute `[data-effect="bold"]` has been added to `strong` tag: 
image
and this selector is also used for exercise paragraph titles in Prealgebra (`p > [data-effect="bold"]`) which are set to display as block elements (starting on new line): 
image

Here is how the same paragraph as shown in the issue decription looked before in Precalculus (I have checked the old content form A&T as well): 
image

ajluk (commented):
@KarinaMendez2 Looks good!  I have only left a comment with some explanation why this issue occurred again  with a friendly reminder to regenerate all css styles for all the books and also please remember to change the base branch to `j.j`

lwtchu:
adding a note here - does the `data-effect="bold"` cause problems in general? i think it's because whoever marked up the content initially put `<emphasis effect="bold">` which is redundant and not accurate so wondering if we should remove it in the content 🤔  or this fix will fix it regardless

ajluk:
@lwtchu the `data-effect="bold"` attribute is only problematic because it is a part of the selector by which we select the exercise paragraph titles: `p > [data-effect="bold"]`  and whenever in exercises this title is used (it is present in all of the column shapes) it will override other paragraphs with a child with this attribute (`data-effect="bold`).

lwtchu:
@ajluk ah okay i willl...still ask since it seems redundant for it to have a attribute like that on it

ajluk (approved):

PR #3942, opened 2021-06-22
poli sci - Remove iframes & adjust figure margins

kswanson33:
Figure margin adjusted to allow for more space above the figure:

Original: 
image

Now:
image

Styles have been rebuilt.


styles/designs/cosmos/parts/_figureimage-components.scss
```diff
@@ -8,5 +8,6 @@ $Figure:(
         display: table,
         margin-left: auto,
         margin-right: auto,
+        margin-top: 1rem,
```
brittany-johnson, line 11: since this affects a few books prob want to let alan know in the card to check figures in the styles affected?

brittany-johnson (approved):

styles/designs/cosmos/parts/_figureimage-components.scss
wilkus27: I'm not sure if we should add margin to all figures. There is no space between figure and subtitle, because we set margin-bottom to 0 for this component:

image

How about making this property for figure `OPTIONAL` and enable it only for iconned notes?

Or we can create new component for figure right after subtitle. Actually we don't need margin top for figures that are e.g. after para in the note.

wilkus27 (changes requested):

styles/designs/cosmos/parts/_figureimage-components.scss
kswanson33: Thanks! I changed the implementation, let me know what you think!

wilkus27 (approved):
Everything looks good :relaxed:

LostMojo (approved):

Issue #3924, opened 2021-06-09
english composition - annotation text treatment


lwtchu:
![Screen Shot 2021-06-16 at 1 39 05 PM](https://user-images.githubusercontent.com/11299125/122274571-4ad17280-cea8-11eb-9536-1ec47e41f1c7.png)

# Public domain text
class: `<para class="public-domain">`
**CHANGE** font: from Courier Prime TO **IBM Plex Sans**
**NEW** indent on left and right, kind of like the quote treatment
Add a line to the side on the left
weight: 1px
color: G1: `#1E9575`

# Student sample text
class: `<para class="student-sample">`
Same as above but with B2: `#0074BC`

Stackblocks:
- [ ] Displays OK with current code & content
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/23866da2-c5a3-4b6e-bac6-cbf1290e34af)

PR #3951, opened 2021-06-24
Issue 3924 Annotation Text

LostMojo:
For issue #3924


styles/books/english-composition/book.scss
```diff
@@ -66,9 +66,16 @@ $bandColor: #DCB83D;
   ),
   PublicDomainText: (
     _selectors:("p"),
-    'PublicDomainText:::font-family': (_ref: 'typography:::fontOption1'),
-    'PublicDomainText:::color': (_ref: "colorMap:::fontBodyColor"),
+    'PublicDomainContainer:::border-color': (_ref: 'colorMap:::publicDomainBorderColor'),
+    'AnnotationText:::font-family': (_ref: 'typography:::titleOption2Font'),
```
ajluk, line 70: @LostMojo Everything looks good. I have only one suggestion to change the name of the `AnnotationText` component a little, maybe to InnerAnnotatationText or something similar, because there is a shape of the same name (see line 62 above) and this can be confusing.

ajluk (commented):
I just left a suggestion about changing Annotation Text component name.

styles/books/english-composition/book.scss
```diff
@@ -86,7 +93,7 @@ $bandColor: #DCB83D;
 @include use('ChapterIntro', 'cosmos:::ChapterIntroShape');
 @include use('LearningObjectivesModule', 'cosmos:::LearningObjectivesModuleShape');
 @include use('AnnotationText', 'common:::AnnotationTextShape');
-@include use('PublicDomainText', 'common:::PublicDomainTextShape');
+@include use('PublicDomainText', 'cosmos:::PublicDomainShape');
```
wilkus27, line 96: I noticed that `AnnotationTextShape` and `PublicDomainTextShape` are created in common file by BrittneyJ:

image

They are added only to english composition for now, but I understand they will be used in other themes (different then cosmos) as well. Am I right @brittany-johnson ?

If yes, I think we should keep all components in common file. Also maybe we should use `AnnotationText` name everywhere? I mean, here are components from common-components file:

image

What do you think to rename first one to `AnnotationText__Container` and second one to `AnnotationText--WithLeftBorder__Container` and apply to it these properties:

image

Then we can use one shape both for Public Domain and StudentSample (e.g `AnnotationTextWithLeftBorderShape`).

Also this component:

image

we can rename to `AnnotationText--WithLeftBorder__Content`. What do you think? :slightly_smiling_face:

wilkus27 (changes requested):

styles/books/english-composition/book.scss
wilkus27: @ajluk we did a review at the same time :smile: In this case, can you look at mine comments and tell what do you think about it?

ajluk: @wilkus27 It seems so :smile: I agree that we can standardize the components properties and make one shape (AnnotationTextWithLeftBorderShape).

styles/books/english-composition/book.scss
LostMojo: That sounds great! I think I also have a way to condense the shape a little

LostMojo: Done! Both .public-domain and .student-samples can use the same shape now

wilkus27 (approved):

Issue #3936, opened 2021-06-15
english composition - unit opener 


lwtchu:
![Screen Shot 2021-06-24 at 12 14 04 PM](https://user-images.githubusercontent.com/11299125/123305481-c30de880-d4e5-11eb-880c-106392423209.png)

done a long long long time ago in the old pipeline so prince can do it, but not sure how to do in baked styles https://github.com/openstax/oer.exports/issues/2693 

Specs for screenshot above

## Unit number
font style: Archivo Black
color: R2: `#8B1504`
weight: regular, 400
size: font-scale(7): 32.25pt, 43.2px, 3.6rem

## Unit title
font style: Archivo Black
weight: regular, 400
color: black `#000000`
size: font-scale(7): 32.25pt, 43.2px, 3.6rem

## Paragraph text 
class: `<section class="unit-opener">` in the `introduction` module.
font style: Archivo
weight: regular, 400
color: black
size: font-scale(1): 10.8pt, 14.4px, 1.2rem

## Block colors
background: R2: `#8B1504`
light box: R3: `#FFE5E5`
radius round corner: 135

lwtchu:
Comment for stuff left to do: https://openstax.slack.com/archives/C022NETP39N/p1624645350115500

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-01 14-21-36.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/839f2145-9776-43f5-be77-39425eb5f1d7)

wilkus27:
@lwtchu As I write here https://github.com/openstax/cnx-recipes/pull/3955#pullrequestreview-696166969 Anna just put unit opener section into the H1 for now, to achieve quick solution, since we need English Composition be ready as soon as possible. I'm trying to rework this now. What I wanted to do was moving section under the H1. And I found solution how to do this but then I realized that all changes in recipes that we make have inpact on webview. We can't move section just in PDF. So my question is, where text from unit opener section should be displayed on cnx/rex? I guess we need to create a whole new module with header and metadata before chapters. I mean collection for unit should contain single module, something like Introduction, and then collections of chapters:

![image](https://user-images.githubusercontent.com/40228854/124426042-bf8e2300-dd69-11eb-8c6e-53856b55d45b.png)

Because on webview each module is a separate webpage. When we do that we have to think about styles in PDF. There will be additional header created for unit introduction module, so we will need new design for that.

[...]

PR #3955, opened 2021-06-25
create unit opener

ajluk:
Issue:  #3936


wilkus27 (approved):
I approved, because visually it looks good and we need English Composition be done as soon as possible, but I think we should rework it. This is how markup looks now:

image

I'm not sure if we should put section inside the header and then section in another section with the same class. I know Anna didn't have enough time to find another solution for that before her vacation. I tried to fix that today but I need to spend more time on that. So maybe I will make a fix in another PR. @LostMojo @lwtchu Let me know what do you think :slightly_smiling_face:

kswanson33 (approved):

Issue #3935, opened 2021-06-14
english composition - handbook outline


lwtchu:
![Screen Shot 2021-06-15 at 2 29 32 PM](https://user-images.githubusercontent.com/11299125/122112167-2614c680-cde6-11eb-9a73-5367806da91b.png)

Documentation #3175 
Version 2 (Chapter Outline)
Title: Outline
Numbering for list: H.# (H.1, H.2, etc) 
Linking will be done in content.

wilkus27:
@lwtchu Do you plan to add some other text to Outline list or there will be always just titles of the sections from Handbook page? If just titles, I think we can create Outline totally by recipe, not in the content, like we do for Chapter Outlines.

lwtchu:
@wilkus27 interesting...can we do it with the h1's?

wilkus27:
Yes, first I have to add numbering to section titles, but then I can copy entire title and use it in Outline.

lwtchu:
@wilkus27 🤩  awesome i'll update the tagging legend

wilkus27:
@lwtchu This is how it looks (without styles for now). I created `div` with `os-handbook-outline` class: 

![image](https://user-images.githubusercontent.com/40228854/123651866-1e184600-d82c-11eb-9559-592449b8b203.png)

Can I ask you for a few content changes? :slightly_smiling_face: 

1. Remove section with `handbook-outline` class.
2. Add ids to section titles that should be numbered.
3. Create some other section after "Paragraphs and Transitions" so we can test if numbering is working good.

lwtchu:
@wilkus27 done 8) sorry it took so long i missed the comment :grimace:

wilkus27:
@lwtchu No problem :wink: I just finished adding styles for it. And this is how it looks:

![image](https://user-images.githubusercontent.com/40228854/123831326-80447a00-d904-11eb-84b4-81ced348491f.png)

I'm moving card to code review :slightly_smiling_face:

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-01 14-27-41.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/70be3ecc-32d5-4cac-b878-e6b202c5b91b)

Issue #3932, opened 2021-06-14
english composition - handbook numbering


lwtchu:
![Screen Shot 2021-06-15 at 2 29 32 PM](https://user-images.githubusercontent.com/11299125/122112167-2614c680-cde6-11eb-9a73-5367806da91b.png)

class: `<document class="handbook">` and `<content class="handbook">`

follows same design for headers in #3164 for appendix/any  end of book content

can we add numbering to the h1's?

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-21 09-13-48.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/2826df84-b7a0-4497-b2cf-272501bcfa6c)

PR #3956, opened 2021-06-28
Fix Handbook and add styles to it

wilkus27:
Issues: https://github.com/openstax/cnx-recipes/issues/3932 and https://github.com/openstax/cnx-recipes/issues/3935


styles/design-settings/cosmos/parts/_titles-settings.scss
```diff
@@ -32,6 +32,29 @@
     ),
 ));
 
+@include add_settings((
+    EobPageTitles: (
+        _selectors: ('body > [data-type="page"]:not(.appendix)'),
```
kswanson33, line 37: should this selector also exclude `preface` pages?

styles/designs/cosmos/parts/_titles-shapes.scss
```diff
@@ -51,6 +51,35 @@
     )
 ));
 
+// Used for Handbook in English Composition.
+// This is EOB section with [data-type="page"], similar as Appendix.
+// To discuss: we can move appendix from PageTitlesShape to this one (but it has different specs for H1),
+// or we can add specs for H2-H4 to EobTitlesShape and use that shape for Handbook (even it's a page, not composite-page)
```
kswanson33, line 57: This seems fine to me. I think that it would also work to combine this shape with `EobTitlesShape`, but it also makes sense to keep them separate since one is for a page and one is for a composite page. The one confusing thing is the name -- not the name of this shape, but `EobTitlesShape` should really be `EobCompositePageTitlesShape` now that this shape exists. That would impact a lot of books, though, so I'm fine with keeping it as-is.

kswanson33 (approved):
Approving, with one question re: selectors!

styles/design-settings/cosmos/parts/_titles-settings.scss
wilkus27: Yes :see_no_evil: Thank you for noticing that! I pushed needed fix.

Issue #3959, opened 2021-06-28
Need to Translate "Access for free at openstax.org" to Spanish for PDFs


openstaxalina:
# Brief Description of Change and Screenshots
What kind of change is this? [ ] recipe [ ] styles [x] i don't know

![Screen Shot 2021-06-28 at 9 50 06 AM](https://user-images.githubusercontent.com/7863733/123657254-41c09980-d7f6-11eb-8bb8-6a1b7aacd7b6.png)

Please let me know if you need JR Language to provide and/or verify the translation. Thank you!

# Acceptance Criteria

# Content Location(s)
collection ID:
server:
chapter/section: 
surrounding text:

lwtchu:
@KarinaMendez2 before you  go,  do you think you could translate t his :eyes:

Stackblocks:
Good news / bad news time:

**Good News:** These two folio elements are no longer in English
**Bad News:** These two folio elements aren't in Spanish either. They're missing in both local and CORGI bakes.

This will need to be resolved, but first we need to ID if this is a showstopper for the impending release of Spanish University Physics, or can it be fixed in post?

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/3360e831-77d5-45a4-a08c-ba4795fb0c3e)

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/e3379068-faf3-4532-b0b7-d787a138daf9)

Stackblocks:
Fixed via https://github.com/openstax/recipes/commit/b8147eb8ec36c15bcbbb4c12b15bf6fdd9cd35b6! 
![image](https://user-images.githubusercontent.com/12678235/130130803-0c668159-9533-4b94-b485-241795778d8a.png)

PR #3970, opened 2021-06-30
Pdf spanish translation

KarinaMendez2:
for: https://github.com/openstax/cnx-recipes/issues/3959#issuecomment-871004812


styles/designs/cardboard/pdf/locale/es/_folio.scss
```diff
@@ -0,0 +1,201 @@
...
+      content: " ";
+    }
+  } 
+  
+  @mixin CornFolioStyles() {
+      font-size: font-scale(-1);
+      font-family: "Noto Sans";
+      color: #000000; 
+      font-weight: 700;
+  }
```
wilkus27, line 132: It's not related with this PR, but I noticed that in every folio file this mixin is named `CornFolioStyles`. We should probably create a separate card for that and fixed names to fit to the theme.

styles/designs/corn/pdf/locale/en/_folio.scss
```diff
@@ -0,0 +1,219 @@
+@import '../pdf/locale/pl/text-varaibles';
```
wilkus27, line 1: I think we can remove this import. From polish file too. I gues that is something from previous idea how to translate words :thinking: I don't know. As I checked file `text-variables` doesn't exist, so it shouldn't change enything.

Also I noticed that your directory of this file is named `en`, not `es`.

styles/designs/cosmos/pdf/locale/en/_folio.scss
```diff
@@ -0,0 +1,187 @@
+//Folio content strings
```
wilkus27, line 1: Here also directory is `en`. It should be `es`.

wilkus27 (changes requested):
I added some comments. Beside that, should this be applied now to any specific book? If yes, you have to add import to `book.scss` file for this book.

Here is an example of Polish Psychology:

image

styles/books/u-physics/book.scss
```diff
@@ -11,7 +11,7 @@ $ChapterIntroType: fullWidth;
 @import 'framework/framework';
 @import 'design-settings/carnival/design';
 @import 'design-settings/carnival/settings';
-@import '../../designs/carnival/pdf/folio';
+@import '../../designs/carnival/pdf/locale/es/folio';
```
brittany-johnson, line 14: Changing this path will make the English folio Spanish - I'm wondering if there's a way we can import the folio conditionally so that we don't have to create a whole new book.scss just for the Spanish translation? @wilkus27 

@KarinaMendez2 In the Spanish content, is there something in the `head` element of the document that specifies the language?

brittany-johnson (changes requested):

styles/books/u-physics/book.scss
wilkus27: Hmm good question :thinking: When we created such solution with @ajluk we assumed that all foreign books will have separate recipes (as Polish books). I saw you had a call with Phil about it. Do you came up with any ideas?

LostMojo (approved):

styles/designs/carnival/pdf/_folio.scss
```diff
@@ -170,8 +170,13 @@ div[data-type="composite-page"].os-eob {
 
 $defaultLeftContent: string(chapter-number) " • " string(chapter-title);
 $defaultRightContent: string(module-number) " • " string(module-title);
-$prefaceContent: "Preface";
-$accessContent: "Access for free at openstax.org.";
+
+$prefaceContent: string(preface-msg);
+$accessContent: string(access-for-free-msg);
+
+[data-pdf-folio-preface-message][data-pdf-folio-access-message] {
```
wilkus27, line 177: I have a few questions :relaxed: These are new attributes, that's right? Where are they added? In the content?
I'm asking because changes from this PR have an impact on all books from carnival and corn theme. And if books don't have these attributes folio for them will be broken.

As I see there is output file only for u-physics. Did you use `./script/build-styles` command or the short one that builds styles only for one specific book?

wilkus27: Ok I looked to the Kitchen PR https://github.com/openstax/kitchen/pull/341/files :grin: They are added here:

image

But as I understand it will be implemented only to u-physics for now, so other books won't have these attributes. We can resolve that in two ways: either we have to add attributes to books in old recipe or we have to create some condition when folio should use old settings and when the new ones.

wilkus27 (changes requested):

styles/designs/carnival/pdf/_folio.scss
wilkus27: Maybe we can add such variable to `book.scss` file:

image

And then in folio:

```
@if $useNewRecipe == true {
   $prefaceContent: string(preface-msg);
   $accessContent: string(access-for-free-msg);

   [data-pdf-folio-preface-message][data-pdf-folio-access-message] {
     string-set: preface-msg attr(data-pdf-folio-preface-message) access-for-free-msg attr(data-pdf-folio-access-message); }
} else {
  $prefaceContent: "Preface";
  $accessContent: "Access for free at openstax.org.";
}
```

But with this solution after we add such variable to `book.scss` file we can't generate PDF from file baked with old recipe. Otherwise folio will be broken. The question is are we ok with that :thinking: If we want to be able to generate PDFs from files baked both with old and new recipe, I think we should create mixin which adds attributes to all books.

@KarinaMendez2 @ajluk @kswanson33 I'm curious what is your opinion about that. Let me know what do you think :wink: Which solution will be better? Or maybe you have some other ideas.

wilkus27 (approved):
Decision: books will not be released without Kitchen recipe, so we are ok that folio is broken for books baked with old pipline.

lwtchu:
CONGRATULATIONS :) i know this was a lot of work nice nice nice nice

PR #3974, opened 2021-07-06
Add GitHub actions

m1yag1:


LostMojo (approved):

.github/workflows/tests.yml
```diff
@@ -0,0 +1,24 @@
+---
+name: tests
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  schedule:
+    - cron: '0 0 * * 0' # weekly
```
philschatz, line 9: I'm curious what the advantage of running weekly is; is it just to catch if 3rd-party packages have somehow changed over time? I'm guessing that this repo will become abandoned once easybake is no longer used

philschatz (changes requested):

.github/workflows/tests.yml
m1yag1: That's just something I always put in my workflows. You could call it a weekly "sanity check". Do you want me to remove it?

m1yag1: I went ahead and removed the weekly run.

PR #3975, opened 2021-07-06
Change Anthro learning objectives to Carnival

LostMojo:
First part of switching Anthropology over to Carnival (and changing carnival to better match its specs #1077 )

This may affect other books in carnival


lwtchu:
@Stackblocks for testing i think we can focus these changes on anthro, uni physics and anatomy & physiology since they're the next books up in q3 to publish  🤔

styles/books/anthropology/book.scss
```diff
@@ -3,15 +3,35 @@ $PagesWithBands: (
...
 @import '../../design-settings/cosmos/_settings.scss';
 @import '../../designs/cosmos/pdf/folio';
 
+//Switching over to carnival, delete cosmos when done
+$ChapterIntroType: fullWidth;
+
+@import '../../design-settings/carnival/_design.scss';
+@import '../../design-settings/carnival/_settings.scss';
+@import '../../designs/carnival/pdf/folio';
+
```
wilkus27, line 19: I think we should remove from this file everything related with cosmos theme and just add carnivals styles for elements one by one. I'm afraid that some of the settings might be overwritten when we have imports from both themes.

wilkus27 (changes requested):

kswanson33 (approved):
This seems like a reasonable approach to me.

Don't forget to base this on the milestone!

styles/books/anthropology/book.scss
LostMojo: Done! I've left some of the old Cosmos commented out, so we can see at a glance the classes & structure used

wilkus27 (approved):

styles/books/anthropology/book.scss
```diff
@@ -3,118 +3,116 @@ $PagesWithBands: (
...
-@include use('Link', 'cosmos:::LinkShape');
-@include use('TocNoUnit', 'cosmos:::TocNoUnitShape');
-@include use('Blockquote', "cosmos:::BlockquoteShape");
-@include use('LearningObjectivesModule', 'cosmos:::LearningObjectivesModuleShape');
+// @include use('BookRoot', "common:::BookRoot");
+// @include use('ChapterIntro', 'cosmos:::ChapterIntroShape');
+// @include use('Link', 'cosmos:::LinkShape');
+// @include use('TocNoUnit', 'cosmos:::TocNoUnitShape');
+// @include use('Blockquote', "cosmos:::BlockquoteShape");
+// @include use('LearningObjectives', 'carnival:::LearningObjectivesShape');
```
wilkus27, line 83: This should be uncommented :wink:

styles/books/anthropology/book.scss
```diff
@@ -3,118 +3,116 @@ $PagesWithBands: (
...
+// ));
 
 
-@include use('BookRoot', "common:::BookRoot");
-@include use('ChapterIntro', 'cosmos:::ChapterIntroShape');
-@include use('Link', 'cosmos:::LinkShape');
-@include use('TocNoUnit', 'cosmos:::TocNoUnitShape');
-@include use('Blockquote', "cosmos:::BlockquoteShape");
-@include use('LearningObjectivesModule', 'cosmos:::LearningObjectivesModuleShape');
+// @include use('BookRoot', "common:::BookRoot");
```
wilkus27, line 78: This one too :upside_down_face:

Issue #3963, opened 2021-06-28
anthropology - table design to carnival t heme


lwtchu:
documentation #1080
should be really close to cosmos current design - only changes are probably some font and color tweaks
accent color should be B2: `#0074BC`

kswanson33:
as per slack discussion with @lwtchu, using the same accent color for tables as for figures (`#C31427`)

kswanson33:
This will impact BasicTables across carnival

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-20 14-35-56.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8c7ab40e-7c5c-4cde-ad6d-0c65e8065493)![Screenshot from 2021-07-20 14-35-21.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/57bbcbf6-4b45-44ae-b68c-b3d22fabdd6d)

Issue #3962, opened 2021-06-28
anthropology - update figure captions to carnival theme


lwtchu:
documentation #1081 - also made some clean up to the documentation so it's clearer. make sure it matches

kswanson33:
This will also impact figures across carnival

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-20 14-39-19.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/8665bc3f-1b2e-4afa-9fa3-8c2876eb7e90)![Screenshot from 2021-07-20 14-39-45.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3416c05e-7cab-44ff-8e56-9b96b02bee17)

Issue #1080, opened 2019-05-20
carnival theme: top-titled tables


lwtchu:
![Screen Shot 2019-06-26 at 4 00 11 PM](https://user-images.githubusercontent.com/11299125/60214649-81310580-982b-11e9-9e5b-bb086ab6de9d.png)

PDF output available in the carnival theme assets folder: https://drive.google.com/drive/folders/1zA3DPstc7Cx1BqtdMYCEB_NGXn1D0XXw

For the number and captions, we can move to below the table.
SOME books have titles on the top and some have titles on the bottom. 
Please use this design for `top-titled` classes. 

Lightweight lines for in between cells (horizontal and vertical): 0.025rem (0.25px)
Header row line weight: 0.1rem top and bottom
Header row font: muli bold
Color for header row font: can we match the header fonts that are book specific? (in screenshot it's #1E9475)
Header column line weight: 0.1rem left and right
Header column font: muli bold
Footer rows: 0.1rem at the top
Caption font: Same as base font
Table body font: alegreya base font size
Header body font: Muli (bold) base font size

Class: `top-titled`

lwtchu:
Investigating some stuff first: #1174

brittany-johnson:
<img width="747" alt="Screen Shot 2019-08-13 at 10 17 43 AM" src="https://user-images.githubusercontent.com/24909007/62953724-a665dd00-bdb3-11e9-868c-e3550310dc85.png">

brittany-johnson:
Table with no class: 
<img width="738" alt="Screen Shot 2019-08-13 at 10 20 45 AM" src="https://user-images.githubusercontent.com/24909007/62953963-212ef800-bdb4-11e9-93eb-548d06f2588d.png">

Table with class `.unnumbered .medium`
<img width="799" alt="Screen Shot 2019-08-13 at 10 20 01 AM" src="https://user-images.githubusercontent.com/24909007/62953988-3572f500-bdb4-11e9-87ad-48ae085316c3.png">

brittany-johnson:
Connected to PR https://github.com/openstax/cnx-recipes/pull/1288

Stackblocks:
Mostly looking good with the exception of 2.14 which @brittany-johnson is investigating.

# Good
![2019-08-14_15-16-32.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/ce045a48-2ba4-4c66-8814-301bba8a52a2)

[...]

PR #3981, opened 2021-07-09
style figures and tables in carnival for anthropology

kswanson33:
#3962 & #3963

image

image


styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -115,6 +122,8 @@ $Caption__Title__Label--Table: (
     _properties: (
         font-weight: bold,
         font-family: enum('ValueSet:::REQUIRED'),
+        text-transform: uppercase,
+        color: enum('ValueSet:::OPTIONAL'),
```
ajluk, line 126: @kswanson33 Can I ask why this property value is set to `optional`? As far as I remember all the tables/figures captions should rather be unified through the whole theme. If yes Fake Titled Tables should also have the captions titles colorized and the value should be set to `required`. But let me know if I'm wrong :slightly_smiling_face:

styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -124,5 +133,6 @@ $Caption__Number--Table:(
     _properties:(
         font-weight: bold,
         font-family: enum('ValueSet:::REQUIRED'),
+        color: enum('ValueSet:::OPTIONAL'),
```
ajluk, line 136: See the comment from above.

ajluk (changes requested):

ajluk:
@kswanson33 I've got also one more question about colored table bottom borders, because they are included in the documentation but you don't add them within this PR? Was it intended? There was a problem with such borders in carnival in the past, but in cosmos theme Natalia figured out how to fix it. Here is the PR with such solution in cosmos: https://github.com/openstax/cnx-recipes/pull/3731/files

LostMojo (approved):

kswanson33:
@ajluk Thank you for catching all that! You're right, I had just overlooked the content-driven alignment, bottom border color, and also the top border, all of which I've now added. I wasn't sure if it was okay to impact `FakeTitleTables` with the caption color, but I've gone ahead and made that change since I agree it's probably what we want.

styles/design-settings/carnival/parts/_table-settings.scss
```diff
@@ -36,12 +36,18 @@
             font-family: (_ref: "typography:::titleOption1Font"),
             font-weight: 700,
             border-bottom: 0.1rem solid black,
+            border-top: 0.1rem solid black,
+        ),
+        LastTableRowData: (
```
ajluk, line 41: @kswanson33 The color bottom border was added to last `td`s in this way in carnival theme previously but not worked for merged cells. I think the only solution is to use `after` element added as Natalia did in cosmos in the PR I pasted above: https://github.com/openstax/cnx-recipes/pull/3731/files. Here is the after element Natalia used in the table shapes to create a bottom border: 
image and then she used in in the caption container instead of `td`: 
image
I hope it's not confusing, but if you will have some questions just let me know :) I have also asked about it Larissa yesterday because I was confused myself if we need such changes in carnival because previously we decided to remove the bottom stroke there and wasn't sure if you discussed it and my comment were just irrelevant :sweat_smile:

styles/design-settings/carnival/parts/_table-settings.scss
```diff
@@ -36,12 +36,18 @@
...
+            border-bottom-color: (_ref:"colorMap:::figureNumberColor"),
```
ajluk, line 42: @kswanson33 Also this color should match the accent color for athropology spec: 
image
It is also a book specific color: 
image
It should be set in the settings in anthropology `book.scss` file. And here we need Larissa advise I guess, because I'm not sure if the book specific colors for bottom strokes should be set for every carnival book or just for anthropology for now. If you don't want to do that within this PR maybe we can set the `border-bottom color` property value to `OPTIONAL` temporarily and set the value only for anthropology. And then create a separate PR adding strokes for the rest of the carnival books.

ajluk: I have asked Larissa if the change should be made just for anthropology here: https://github.com/openstax/cnx-recipes/issues/1080#issuecomment-878879358

ajluk (changes requested):
I have left some more comments in the places which concerned me and also asked Larissa for advise how much wide the changes should be, because I'm not sure here as well :)

styles/design-settings/carnival/parts/_table-settings.scss
kswanson33: In a PM with Larissa, we decided to change the accent color to red instead of blue -- I should have documented that publicly, apologies! I've added a comment to the [documentation](https://github.com/openstax/cnx-recipes/issues/3963#issuecomment-879135506). Sounds like Larissa wants to move towards keeping things like accent colors the same across the theme.

styles/design-settings/carnival/parts/_table-settings.scss
kswanson33: I see 👀, my bad for trying to implement this without looking at that PR! Thank you for pointing out this bug, good to know. I will go back and implement this the cosmos way.

styles/design-settings/carnival/parts/_table-settings.scss
ajluk: Oh, ok. Thank you for the explanation. I wasn't aware that such things will be unified. :smile:

styles/designs/carnival/parts/_table-shapes.scss
```diff
@@ -42,13 +42,13 @@
...
                                 )),
                                 $Table__Row--Last--MergedCells__Data,
-                                $Table__Row--Last__Data,
                             )
                         )),
                     )
                 )),
                 map-merge($Caption__Container--Table, (
                     _components: (
+                        $Table__BottomBorder,
```
ajluk, line 51: @kswanson33 As I understood from documentation guess this component should be added also to Fake Titled Tables shape below. This also requires testing a carnival book containing Fake Title Tables.

ajluk (changes requested):
@kswanson33 There is only one thing I think should be updated with your recent changes (Fake Titled Tables shape), but please correct me if I missed something.

styles/designs/carnival/parts/_table-shapes.scss
ajluk: @kswanson33 As we discussed slack adding a bottom border to Fake Title Tables will be done later on within a separate issue and PR.

ajluk (approved):

Issue #3965, opened 2021-06-29
anthropology - update chapter outline to carnival theme


lwtchu:
Documentation #1076

Version 2 without the list for LO. 

Accent color B2: `#0074BC`

kswanson33:
This will also impact any carnival book that used the old version of the Version 2 outline

MartaSztuka:
##### PASSED
![Screenshot from 2021-07-21 09-05-42.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/3a4201d3-003b-4ceb-8930-acc9a9d17fda)

Issue #3957, opened 2021-06-28
Placement of Chapter Outline "Line" is Awkward with Spanish Translation


openstaxalina:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [x] styles [ ] both [ ] i don't know

![Screen Shot 2021-06-28 at 9 40 35 AM](https://user-images.githubusercontent.com/7863733/123655846-fe196000-d7f4-11eb-8754-9488040e78f3.png)

This looks fine in the English version (below), but perhaps because of the extra characters of the Spanish translation, it looks like a mistake there.

![Screen Shot 2021-06-28 at 9 42 26 AM](https://user-images.githubusercontent.com/7863733/123655980-1d17f200-d7f5-11eb-8434-66c5651cb711.png)

# Does this happen for every instance? If no, is there a discernable pattern of occurrences?
Yes - every chapter opener

# Sample mark up if applicable

# Acceptance Criteria

# Bug Location(s)
collection ID:
server:
chapter/section: 
surrounding text: 

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
Updated documentation to tighten up the design #1076 - might be able to get this update when we implement for anthro

Stackblocks:
- [x] Displays OK with current code & content

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/93edced9-a867-4d8e-bb35-e82e619363c4)

PR #3983, opened 2021-07-09
style splash and chapter objectives for anthropology

### LINKED ISSUE #3961: anthropology - add carnival theme splash treatment
Documentation #1075; specs have been updated to match current formatting. might have some slight adjustments to text font/size

accent color B2: `#0074BC`

kswanson33:
will resolve #3961, #3965, and #3957

image


LostMojo (approved):
Looks good! Thank you for doing this!

styles/books/anthropology/book.scss
```diff
@@ -82,9 +82,17 @@ $ChapterIntroType: fullWidth;
 //   ),
 // ));
 
+@include add_settings((
+  FullWidthChapterIntro: (
+    'SplashFigureBg:::background-color': (_ref: 'colorMap:::chapterOutlineColor'),
+    'OutlineFullWidth:::border-bottom-color': (_ref: 'colorMap:::chapterOutlineColor'),
+    'OutlineFullWidthTitle:::color': (_ref: 'colorMap:::chapterOutlineColor'),
+    'OutlineFullWidthTitleBottomBorder:::border-bottom-color': (_ref: 'colorMap:::chapterOutlineColor'),
```
ajluk, line 90: @kswanson33 This property set for `OutlineFullWidthTitleBottomBorder` from above can be removed since this component no longer exists. But I'm little concerned because this change (I mean removing the component and restyle chapter outline) affects every book in carnival theme, is it intended? If yes this property should be removed from other carnival `book.scss` files as well. If this change only should affect anthropology we need to create a separate shape for Chapter Outline for this book not containing `OutlineFullWidthTitleBottomBorder` and keep OutlineFullWidthTitleBottomBorder (`$ChapterOutline--FullWidth__Title--BottomBorder: titleBottomBorder("OutlineFullWidth", 116.7); from line 103 `component in the `chapter-intro-components` file.

styles/designs/carnival/parts/_chapterintro-components.scss
```diff
@@ -65,6 +65,15 @@ $ChapterOutline--Margin__Objective__Link--Text: (
...
 
+$ChapterOutline__Objective: (
+    _name: "OutlineObjective", 
+    _subselector: ' > div.os-chapter-objective', 
+    _properties: (
+        margin-bottom: .5rem,
+        padding-right: enum('ValueSet:::OPTIONAL'), // Property used only in MarginChapterIntroShape
+    )
+);
+
```
ajluk, line 76: Could you add `--Margin` to the components name  because now it is only used in in the Chapter Intro Margin Shape and change the `padding-right` property  to `REQUIRED`? Or maybe a better idea could be to just leave a comment that if change the styling of the full width chapter intro will be permanent the change of the name and property value will be needed there? What do you think? :slightly_smiling_face:

ajluk (changes requested):
@kswanson33 I have left some comments where I'm little concerned about how this changes affect other books in carnival.

kswanson33:
@ajluk thanks for pointing this out, I've removed the `OutlineFullWidthTitleBottomBorder` & renamed the `ChapterOutline--Margin__Objective`, and rebuilt styles to make sure everything still works (no changes). & I can confirm that this should impact the rest of the carnival theme (many of these anthropology issues seem to include some carnival refactoring)

ajluk (approved):

PR #4055, opened 2021-08-11
add numbered learning objectives settings

LostMojo:
Adds styling for numbered  Learning Objectives


recipes/output/bca.css
```diff
@@ -2421,21 +2421,6 @@
 :pass(14) nav#toc li > a:deferred {
   content: pending(title-spans); }
 
-:pass(14) .introduction .os-chapter-outline {
```
brittany-johnson, line 2424: I think if you add `$Config_hasGroupedIntroBody: true;` to the BCA config the diff should be resolved

brittany-johnson (changes requested):

recipes/output/bca.css
LostMojo: All checks pass now!! Good to merge?

brittany-johnson (approved):

PR #4057, opened 2021-08-12
Make generic styles from rex a webview theme

jarosik10:
for: https://github.com/openstax/cnx-recipes/issues/4105


styles/designs/webview/parts/_titles-shapes.scss
```diff
@@ -0,0 +1,44 @@
...
+@include create_shape('webview:::PageTitlesShape', (
+    _components: (
+        map-merge($MainContent__Container, (
+            _components: (
+                $H1,
+                $H2,
+                $H3,
+                $H4,
+                $H5,
+                $H6,
```
wilkus27, line 12: You probably should add section wrappers to all titles shapes:

image

They are necessary, because some elements have headers (e.g notes have h3) and they are styled differently then page titles. When we specify precise selector (e.g for `section > section > section > h3`) we are sure that other styles are not overwritten (e.g for `[data-type="note"] > h3`).

wilkus27 (changes requested):
Looks good :wink: :+1: I have just a few requested changes.

jarosik10:
@wilkus27 @ajluk 
What would be a proper solution to handle more than one selector inside `_subselector`?
Creating another component?
for example;
from:
```css
$Media--InFigure: (
  _name: "MediaInFigure",
  _subselector: ' > [data-type="media"], > .media',
  _properties: (
    display: block,
    margin: 0,
    text-align: center,
  )
);
```
to:
```css
$Media--InFigure: (
  _name: "MediaInFigure",
  _subselector: ' > [data-type="media"]',
  _properties: (
    display: block,
    margin: 0,
    text-align: center,
  )
);

$Media--InFigure2: (
  _name: "MediaInFigure",
  _subselector: ' > .media',
  _properties: (
    display: block,
    margin: 0,
    text-align: center,
  )
);
```
It doesn't look good to me, because of the code repetition. Do you have any other idea?

jarosik10:
@TomWoodward 
Fixes:

1. Bigger font-size
2. Some of the styles in rex seems to be always overridden/dead. Omitting these selectors would make the styling of some of the cnx-recipies components simpler.

Todo:
Disable default list-style in abstract

1. Change font weight of terms
2. Missing styles for glossary/eoc/eob
3. Add styling for noteish elements. Should I create a shared components for noteish elements? Or make every noteish element to be represented by an individual set of components? The answer to this question depends on how much do we want every noteish element to be customizable and what styling properties we would like to inject from the pdf styles.
4. Check how much the pdf styles for noteish elements differ depending on book design.

I think it’d be good to create some basic shared styling for noteish elements. And even before completing the styles migration from rex to cnx-recipes, i would like to test the injection of book specific styles (from a different design) into the existing styles of webview design. It would check if the existing conception, and already created shapes actually work. For example, it could show me that some properties should be changed to optional, or should not be hardcoded, etc.

jarosik10:
> Bigger font-size

Every style needs a container. In rex it is nested inside:
```css
#main-content:not([data-dynamic-style="true"]), .content-excerpt:not([data-dynamic-style="true"]) {
  font-size: 1.6rem;
  line-height: 2.3rem;
}
```
in cnx-recipies we do not have such a container.

styles/designs/webview/parts/_caption-components.scss
```diff
@@ -0,0 +1,89 @@
...
+);
+
+$Caption__Container--ScaledDown30:(
+    _name: "CaptionContainerScaledDown",
+    _subselector: ' ~ .os-caption-container',
+    _properties: (
+        width: 30%,
+        margin: auto,
+    )
+);
```
wilkus27, line 56: These two components have the same `_name`.  They should be different.

styles/designs/webview/parts/_figureimage-shapes.scss
```diff
@@ -0,0 +1,51 @@
...
+            _components: (
+                map-merge($Caption__Container--Figure, (
+                    _components: (
+                        $Caption__Title--Figure,
+                        $Caption__TitleLabel--Figure,
+                        $Caption__Number--Figure
+                    ),
+                ))
+            ),
+        )),
```
wilkus27, line 33: ### Information for other developers

First step of moving rex styles to Baked Styles project was to get final css file the same as this one from Rex project. Because of that we decided to make some simplifications. As we can see here, this shape doesn't have one parent component. There are a few components at the same level:

image (11)

For that reason there are no selectors in settings:

image (12)

For now it's working but in the next step, when we'll know what is the final design for webview, it should be refactored.

#### To do in the fure
- move settings to their own files (e.g. settings for figures should be in file `figure-settings.scss`)
- split shapes in a way that each of them has one parent component and its selector is defined in settings
- investigate that we need all of the selectors 

For example:
image (13)

Here we need to check if these are two different elements or maybe this is one element that has two attributes, so we can use just one of them for styles. Also there might be media that has class, but doesn't have data-type etc. We should check all cases, then rename those components in a descriptive way.

styles/designs/webview/parts/_titles-components.scss
```diff
@@ -0,0 +1,131 @@
...
+    color: enum('ValueSet:::REQUIRED'),
+    font-size: 1.8rem,
+    margin: '1.5rem 0 1rem 0',
+    font-weight: bold,
+  )
+);
+
+// H5 headers
+$H5: (
+  _name:"H3Title",
```
wilkus27, line 82: Typo :wink: It should be `H5Title`

styles/designs/webview/parts/_titles-components.scss
```diff
@@ -0,0 +1,131 @@
...
+    color: enum('ValueSet:::REQUIRED'),
+    font-size: 1.8rem,
+    margin: '1.5rem 0 1rem 0',
+    font-weight: bold,
+  )
+);
+
+// H6 headers
+$H6: (
+  _name:"H3Title",
```
wilkus27, line 94: Also here. `H6Title`

wilkus27 (changes requested):
It looks very good :relaxed: I found a few small things related to names of components. Beside that everything is ok :tada:  I added also a comment for other developers, so they will know what are the plans and why some things are created differently then in styles for PDFs. 

Also I have one question: there are no output files in this PR. I know you are just generating them locally and use them somehow in a rex. Are we ok with that for now? Or it will be better to push these files here so everyone has access to them?

One of the next step will be also to modify `build-styles` and `build.js` scripts so they can generate files for webview styles properly.

wilkus27 (approved):

Issue #4052, opened 2021-08-06
finance - tighten spacing for tables


lwtchu:
so for the default tables, what's the smallest we can go? same question as #4050 - how small can we go? can we match it to the timeline table? around 2px all around?

trying to achieve this look
![Screen Shot 2021-08-04 at 3 19 03 PM](https://user-images.githubusercontent.com/11299125/128575026-945b458d-f2a7-4e88-b6d0-559b07e535e8.png)

***IF YOU NEED A CLASS*** i can suggest `condensed-table` and you can ask a CM to add it to all the tables

LostMojo:
@lwtchu Should this affect all cardboard tables or just the ones in finance?

lwtchu:
in book..scss for finance, when it goes `@include use('DefaultTable', 'cardboard:::TableWithBordersShape');`  just change 'DefaultTable' to 'CondensedDefaultTable'

Stackblocks:
- [x] Displays OK with current code & content

![image](https://user-images.githubusercontent.com/12678235/134697715-e41ef4c5-50a0-4771-945b-62c784df9be5.png)

Issue #4050, opened 2021-08-06
finance - update timeline table design


lwtchu:
![Screen Shot 2021-08-06 at 4 48 55 PM](https://user-images.githubusercontent.com/11299125/128574385-9e8b96dd-0c07-4294-8857-ea88a972976c.png)

`<class="timeline-table">`

specs  stay mostly the same please make the following changes
- [x] remove top and bottom border
- [x] remove left and right border
- [x] tighten spacing - this is about 2px all around. what's the smallest we can go without exploding the style?

ajluk:
Hi @lwtchu, welcome back! :relaxed:  I have a question about timeline tables - are they book specific (only present in finance book) or can they be spotted in different books from cardboard theme? I'm asking because we added styles for these tables within shapes created for default and fake titled tables and the shapes were used by different books. So the timeline tables styles were present in css files for other cardboard books and the books could utilize them. But during working on removing the outer borders for timeline tables I needed to create separate shape for these tables and I wanted to confirm with you if it will be ok to remove the old timeline tables styling added within default and top titled shapes and include the new styling with removed outer borders only in finance book? Or rather should I include the shape also in other cardboard books just in case they can be there and also should they have removed outer borders? Thank you!

lwtchu:
@ajluk let's add this shape in other cardboard books as well - and keep it with the timeline-table tag. the old timeline-table style should be overridden with this one.

ajluk:
@lwtchu Oh, ok. I will remove the old styles then. Thank you!

ajluk:
@lwtchu the padding in the tables cells isn't 2px actually, but 0.7rem which gives 8.39px I have set it to 2px (0.16rem) now and this is how timeline tables look like then: 
![image](https://user-images.githubusercontent.com/55203598/130515369-58f10ba8-c12d-4491-9023-4297ce2fda7b.png)
![image](https://user-images.githubusercontent.com/55203598/130515387-d605ff79-fc51-4078-92dd-9c2aa67e0228.png)
![image](https://user-images.githubusercontent.com/55203598/130515748-570c6c8c-02df-4287-b7f3-3598bbc19e5b.png)
![image](https://user-images.githubusercontent.com/55203598/130515798-93f623a5-613b-4f7d-9fe0-c260b96cde91.png)
Should I keep it or make the padding a little bigger? Let me know what do you think?

lwtchu:
We can leave it as is-- what is it now? Book specific?

[...]

PR #4077, opened 2021-08-30
Tighten finance tables (revived)

LostMojo:
#4050 #4052 

Accidentally merged the previous branch too early, and then reverted it, so here's try two


LostMojo:
This is just #4052 (which accidentally got merged but quickly reverted) but with merge conflicts sorted out

PR #4081, opened 2021-09-02
extend carnival exercises shape to style injected exercises

kswanson33:
for #4089


wilkus27 (approved):
Looks good :wink: :+1: I have one question, do injected exercises appear only in EOC sections? I see they are added only to one shape.

Issue #4031, opened 2021-08-03
OPS-0012 Composition - Indentation


srm-openstax:
# Brief Description of Change and Screenshots
What kind of change is this? [ ] recipe [ ] styles [ ] i don't know

![Screen Shot 2021-08-03 at 11 13 24 AM](https://user-images.githubusercontent.com/86622862/128049943-748ff0ea-6a2d-4f6c-b3a1-16e349e91531.png)

From Shelley: Is there a way to add the icon without creating the extra indentation? It's odd that the two paragraphs don't align.

# Acceptance Criteria

# Content Location(s)
collection ID:
server:
chapter/section: 18.1, p. 5
surrounding text:

lwtchu:
We'll need to test and check with print specs in case this goes into the gutter of the spine. I don't anticipate this blocking any work, so we will take some time to explore options.

Stackblocks:
- [x] Displays OK with current code & content

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/40f66ce8-c601-4033-b436-4f34d9fb49a1)

PR #4091, opened 2021-09-09
Fix indentation in annotation

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/4031

### DIFF
--- a/styles/designs/cosmos/parts/_annotation-components.scss
+++ b/styles/designs/cosmos/parts/_annotation-components.scss
@@ -17,7 +17,7 @@ $AnnotationSectionContainer: (
   _subselector: '',
   _properties:(
     display: flex,
-    margin-left: h-spacing(-3),
+    margin-left: h-spacing(-4),
   )
 );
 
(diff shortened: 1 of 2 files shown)


LostMojo (approved):

styles/designs/cosmos/parts/_annotation-components.scss
```diff
@@ -17,7 +17,7 @@ $AnnotationSectionContainer: (
   _subselector: '',
   _properties:(
     display: flex,
-    margin-left: h-spacing(-3),
+    margin-left: h-spacing(-4),
```
LostMojo, line 20: Does this only effect english-composition?

LostMojo (approved):

styles/designs/cosmos/parts/_annotation-components.scss
wilkus27: Yes, shape for annotations is included only to this book. I built styles for all books just in case, but it didn't change any other output file.

Issue #4105, opened 2021-09-15
Move styles from Rex


wilkus27:
Create basic files and directories for styles from Rex in `cnx-recipes` repo.

### Testing
The output css files are in [repo](https://github.com/jarosik10/rex-test-styles). It uses gh-pages to make the css file accessible:
https://jarosik10.github.io/rex-test-styles/main/webview-generic.css.

To test the styles in REX, you have to visit the page with `content-style` param set to the link of css file.

Example:
https://openstax.org/books/american-government-2e/pages/1-1-what-is-government?content-style=https://jarosik10.github.io/rex-test-styles/main/webview-generic.css

The current version of styles are inside `main` directory: `rex-test-styles/main/webview-generic.css`.
Every PR with a new set of styles will have a separate `webview-generic.css` file. 
Like: `/rex-test-styles/<pr_name>/webview-generic.css`.

lwtchu:
closing since it looks like everything is done. let me know if should be open again

PR #4106, opened 2021-09-15
Add lists terms and font

jarosik10:
css file is uploaded: https://jarosik10.github.io/rex-test-styles/add-lists-terms-and-font/webview-generic.css
Testing was described: https://github.com/openstax/cnx-recipes/issues/4105


styles/design-settings/webview/_colors.scss
```diff
@@ -11,7 +11,7 @@
   colorMap: (
     headingTitleColor: (_ref: "scheme:::grayDark"),
     figureCaption: (_ref: "scheme:::gray"),
```
wilkus27, line 13: Since you are using now `baseColor` in Figure caption settings can we remove this color?

wilkus27 (changes requested):

styles/design-settings/webview/_colors.scss
jarosik10: Good catch!

PR #4107, opened 2021-09-15
Add learning objectives

jarosik10:
css file is uploaded: https://jarosik10.github.io/rex-test-styles/add-learning-objectives/webview-generic.css
Testing was described: #4105


styles/designs/webview/parts/_learning_shapes.scss
```diff
@@ -0,0 +1,12 @@
...
+@include create_shape('webview:::LearningObjectivesShape', (
+    _components: (
+        map-merge($LearningObjectives__Container, (
+            _components: (
+                $LearningObjectives__Title,
+                $LearningObjectives__Paragraph,
+            ),
+        )),
+    ),
+));
```
wilkus27, line 12: I reviewed PR with lists first and approved it, but can you move this component https://github.com/openstax/cnx-recipes/blob/172d9e480945e5e3853dc00a60daba05790365ee/styles/designs/webview/parts/_lists-shapes.scss#L7 here to this shape?

wilkus27 (changes requested):
One small thing with list :wink:

styles/designs/webview/parts/_learning_shapes.scss
wilkus27: Update: after discussion we decide that we leave this as is, since in rex `os-abstract` list doesn't have any parent. We will move it in the next step while refactoring.

wilkus27 (approved):

wilkus27 (approved):

PR #4108, opened 2021-09-15
Add links

jarosik10:
css file is uploaded: https://jarosik10.github.io/rex-test-styles/add-links/webview-generic.css
Testing was described: #4105


styles/designs/webview/parts/_link-shapes.scss
```diff
@@ -0,0 +1,8 @@
+@import 'link-components';
+
+@include create_shape('webview:::Link', (
```
wilkus27, line 3: Can rename shape to `LinkShape`. We will keep this in convention.

wilkus27 (changes requested):

wilkus27 (approved):

PR #4121, opened 2021-09-21
Key Terms spacing

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/4102


styles/books/philosophy/book.scss
```diff
@@ -141,6 +141,12 @@ $bandColor: #DCB83D;
     ),
 ));
 
+@include add_settings((
+  KeyTerms: (
+    'KeyTermsDefinition:::margin-left': h-spacing(1)/4,
```
wilkus27, line 146: Code looks good :wink: :+1: I'm not sure about value :thinking: First of all we shouldn't divide `h-spacing` function or use non integer values inside. If there is a need for other value than `h-spacing(0)`, `h-spacing(1)`, `h-spacing(2)`, etc. we just hard code it. So here you can put just `2px`. But 2px is a very small value. It's almost not visible for the eye, so maybe we can set margin just to `0`? I think that maybe we should show to Larissa PDF with margin 0, and with margin `h-spacing(1)`, so she can decide if it's better to have any space or none.

Also I just noticed that in the issue there is a question if we can apply this change to theme. It's easy to do that, so maybe we also should ask Larissa if it should be done in this PR or in another one later.

wilkus27 (changes requested):

wilkus27 (approved):

PR #4137, opened 2021-09-27
Add input styles

jarosik10:
css file: https://jarosik10.github.io/rex-test-styles/add-input-styles/webview-generic.css


styles/output/webview-generic.css
```diff
@@ -3,6 +3,23 @@
...
+  cursor: pointer; }
+
+button {
+  overflow: visible; }
+
+button, input, optgroup, select, textarea {
+  color: inherit;
+  font: inherit;
+  margin: 0; }
+
```
wilkus27, line 22: ### :memo: To Do:

Analize if we can create components for that, and maybe some shape.

wilkus27 (approved):

PR #4142, opened 2021-09-30
Add tables styles

jarosik10:
css file: https://jarosik10.github.io/rex-test-styles/add-tables/webview-generic.css


styles/designs/webview/parts/_caption-components.scss
```diff
@@ -87,3 +87,43 @@ $Caption__Number--Table: (
     font-weight: bold,
   )
 );
+
+$FigureCaption__Container--InTable: (
+  _name: "FigureCaptionContainerInTable",
```
wilkus27, line 92: I think we should rename this to `Caption__Container--FigureInTable`.

styles/designs/webview/parts/_caption-components.scss
```diff
@@ -87,3 +87,43 @@ $Caption__Number--Table: (
...
+  _name: "CaptionInTable",
+  _subselector: ' .os-caption-container',
+  _properties: (
+    padding: 8px,
+    border-top-width: 0.1rem,
+    border-top-style: solid,
+    border-top-color: enum('ValueSet:::REQUIRED'),
+    font-size: 0.9em,
+  )
+);
```
wilkus27, line 109: This can be removed. The same component is in line `59`.

styles/designs/webview/parts/_caption-components.scss
```diff
@@ -87,3 +87,43 @@ $Caption__Number--Table: (
...
+
+$Caption--InTable__Number: (
+  _name: "CaptionInTableTitle",
+  _subselector: ' .os-number',
+  _properties: (
+    display: inline-block,
+    padding-right: 0.25em,
+    font-weight: bold,
+  )
+);
```
wilkus27, line 129: Here the same situation. First one is in line `71` and second in line `81`.

styles/designs/webview/parts/_lists-components.scss
```diff
@@ -111,3 +111,36 @@ $StepwiseList__NestedList--Ordered: (
...
+  )
+);
+
+$List--InTable--Unstyled__Item: (
+  _name: "OrderedListInTable",
+  _subselector: ' li',
+  _properties: (
+    list-style: none,
+  )
+);
```
wilkus27, line 146: Very interesting styling :thinking: 

### :memo: To do:

Use `list-style-type` in first component like here:

image

 and remove second one

### For now:

I think we should rename this. What do you think about something like that:

`NoneStyleList` for first component and `NoneStyleList__Item` for second one? I don't want to use `Unstyled` word because it might be confusing with component from line `64`.

styles/designs/webview/parts/_tables-components.scss
```diff
@@ -0,0 +1,153 @@
...
+  )
+);
+
+$Caption__Title__Before: (
+  _name: "CaptionTitleBefore",
+  _subselector: ' ::before',
+  _properties: (
+    margin-right: 0.5rem,
+    color: enum('ValueSet:::REQUIRED'),
+    font-weight: bold,
```
wilkus27, line 77: These should be moved to `caption-components.scss`.

styles/designs/webview/parts/_caption-components.scss
```diff
@@ -14,7 +14,7 @@ $Caption__Container--Figure: (
 
 $Caption__Title--Figure: (
   _name: "CaptionTitleFigure",
-  _subselector: " > .os-title-label",
+  _subselector: " > .os-title",
```
wilkus27, line 17: Good catch! :wink: :+1:

styles/output/webview-generic.css
```diff
@@ -1269,6 +1269,148 @@ a:hover {
...
+.os-table table caption sup {
+  top: auto;
+  line-height: inherit; }
+
+.os-table table caption > .title {
+  display: block;
+  font-size: 1.8rem;
+  font-weight: bold; }
+
+.os-table table caption > .title ::before {
```
wilkus27, line 1309: Does it work? Shouldn't be `.title::before` without space?

styles/output/webview-generic.css
```diff
@@ -1269,6 +1269,148 @@ a:hover {
...
+
+.os-table table ol {
+  padding-left: 1.5em; }
+
+.os-table table ol[data-bullet-style="none"] {
+  padding-left: 1.5em; }
+
+.os-table table ol[data-bullet-style="none"] li {
+  padding-left: 1.5em; }
+
```
wilkus27, line 1378: Something is wrong :thinking: There are no properties from these components :open_mouth:

wilkus27 (changes requested):

styles/designs/webview/parts/_tables-components.scss
jarosik10: There are two different caption elements can be inside a table: wrapper with a class `.os-caption-container` or a specific [table caption html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption): `caption`. That's why the table specific one is inside `tables-components` instead of `caption-components`.

styles/output/webview-generic.css
jarosik10: mistake :smile:

styles/output/webview-generic.css
jarosik10: These components had the same `name` property. That was causing this issue ;D

styles/designs/webview/parts/_tables-components.scss
wilkus27: Hmm :thinking: Ok i think this is leftover from something. There should be no `<caption>` tag in tabels after baking. If it exists it means table is not baked properly. But because styles for this exists in rex, lets keep them for now. And yeah, probably is better to not move them for now to another file. Let's mark this as :memo: **To do**.

styles/designs/webview/parts/_figureimage-components.scss
```diff
@@ -132,3 +132,19 @@ $Image--ScaledDown30: (
...
+$Figure--InTable: (
+  _name: "FigureInTable",
+  _subselector: ' .os-figure',
+  _properties: (
+    margin: 0,
+  )
+);
+
+$Image--InTable: (
+  _name: "FigureInTable",
```
wilkus27, line 145: Oh just noticed that the name is the same as for component from above.

wilkus27 (approved):
I found one another small thing, but beside that everything looks good :wink: So I'm approving.

Issue #4113, opened 2021-09-16
OPS-0012 English Composition unwanted vertical line appeared in table


srm-openstax:
We have coded as normal without adding any style. However, the vertical line appears unwanted at the bottom of the table. Kindly check and resolve this.
![vertical-line](https://user-images.githubusercontent.com/86622862/133600853-9ef22f7d-22f6-4eeb-8b31-12e21a61fd1e.png)
![vertical-line1](https://user-images.githubusercontent.com/86622862/133600871-22c2d032-a37c-4d17-b77f-92b3c55362be.png)

# Bug Location(s)
collection ID: col00001
server: https://copper-anteater-egyq5vxq.ws-us15.gitpod.io/
chapter/section: 5.5
surrounding text: Set aside time for several research sessions to find credible

lwtchu:
![Screen Shot 2021-09-16 at 1 40 56 PM](https://user-images.githubusercontent.com/11299125/133667495-a6bd9326-3e7b-4a35-9858-440bdd1e2515.png)
The style isn't consistent when it gets to the third cell. I checked the content and couldn't see anything obvious.

wilkus27:
**Note for a tester**: fix has an impact on all books from cosmos theme.

Stackblocks:
- [X] Displays OK with current code & content

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/b45a43f0-9028-4e2d-a47c-718d3e3a85fc)

PR #4146, opened 2021-10-01
Fix tables eng comp

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/4113


styles/designs/cosmos/parts/_table-components.scss
```diff
@@ -272,6 +272,26 @@ $Table__Data--LastRowLastChild: (
...
+        padding-top: table-v-spacing(1),
+        padding-bottom: table-v-spacing(1),
+        padding-left: table-h-spacing(1),
+        padding-right: table-h-spacing(1),
+        border-width: .02rem, // .25px in rem
+        border-style: solid, 
+        border-color: enum('ValueSet:::REQUIRED'),
+        border-right-style: none,
+        border-left-style: none, 
+        border-right-color: transparent,
```
ajluk, line 291: @wilkus27 I have noticed that the `border-right-color` property value is set to transparent, but but for left border it was enough to just use `border-left-style: none` to make the border "disappear". If that also worked in pdf in this case I think that ` border-right-color: transparent,` can be entirely removed from the two above components :thinking:

ajluk (commented):
Natalia I just left one comment about property which does no harm, but seems unused and can be removed. Everything else looks good though! :relaxed:

wilkus27:
@ajluk Thank you for noticing that! I checked and this property is added to a few other components. So I decided to add this also here rather then remove it.

ajluk (approved):

PR #4147, opened 2021-10-01
Add styles for keyterms

jarosik10:


styles/books/generic/webview.scss
```diff
@@ -160,6 +160,13 @@
     ),
 ));
 
+//Keyterm
+@include add_settings((
+    Keyterm: (
```
wilkus27, line 165: Can you use Camel Case here? (`KeyTerm`) :wink:

styles/books/generic/webview.scss
```diff
@@ -187,3 +194,5 @@
 @include use('Blockquote', 'webview:::BlockquoteShape');
 //Code
 @include use('Code', 'webview:::CodeShape');
+//Keyterm
+@include use('Keyterm', 'webview:::KeytermShape')
```
wilkus27, line 198: Same here.

wilkus27 (changes requested):
Looks good :+1: I left two small comments about naming.

wilkus27 (approved):

Issue #3864, opened 2021-05-17
(WEB STYLING) English Comp - Literary Narrative Trailblazer banner doesn't span the full page width in HTML view


Stackblocks:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [ x ] styles [ ] both [ ] i don't know

![2021-05-17_15-07-39.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/4a414f09-bed0-4cea-b1fb-87147e3fb0e3)

# Does this happen for every instance? If no, is there a discernable pattern of occurrences?
It happens in the sole instance in the test content

# Sample mark up if applicable

# Acceptance Criteria
Banner displays properly in all versions of content

# Bug Location(s)
collection ID: col10001
server: Content06
chapter/section: All
surrounding text: Start of section

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server: 
chapter/section:
surrounding text:

lwtchu:
This  will break the PDF if we fix. Saving for web if we end up  doing it

KarinaMendez2:
Modifying the styles for web will affect pdf

ajluk:
@lwtchu I guess I have found a solution that could work for both pdf and webview. Here is how it looks now in pdf: 
![image](https://user-images.githubusercontent.com/55203598/135897574-a7ff2c6e-c9c7-49ca-b9f1-d26fce810bcc.png)
And here is baked.xhtml file: 
![image](https://user-images.githubusercontent.com/55203598/135897747-ddb37485-a26e-4169-8538-ac1d38ba4578.png)
I fixed it by removing the width of the title with colored bar and add minus right margin to stretch the bar but left the content in place. I have used the exact solution as I did in my old PR for Stats Lab note https://github.com/openstax/cnx-recipes/pull/2599/files (Statistics.) Let me know if this is the desired result :slightly_smiling_face:

lwtchu:
oh nice ! what a fancy fix ~ it looks good. I'll throw it in this milestone

MartaSztuka:
#### PASSED
tested on old content
[...]

PR #4150, opened 2021-10-04
english-composition: fix full page note bar html view

ajluk:
Issue: #3864


kswanson33 (approved):
is the changed component only used in english-comp? if not, styles should be regenerated, but i'm going to assume it is & approve because i figure you probably would have remembered :)

ajluk:
@kswanson33 Yes, it's only used in one book :+1: I have regenerated the styles just to be sure I won't spoil the main branch :sweat_smile: and there were no changes.

PR #4224, opened 2021-11-16
Add inlined exercises styles

jarosik10:


wilkus27 (approved):
For now we can leave this as is, but probably we don't need separate shape. We can just add this properties to regular exercises components. But it has to be investigated if we really need them everywhere or there are some exceptions.

PR #4226, opened 2021-11-16
Add footnotes styling

jarosik10:


styles/designs/webview/parts/_footnotes-components.scss
```diff
@@ -0,0 +1,32 @@
...
+  ),
+);
+
+$Footnote--Class: (
+  _name: "FootnoteClass",
+  _subselector: ' .footnote',
+  _properties: (
+    font-size: 1rem,
+  ),
+);
```
wilkus27, line 32: ### :memo:  To Do:

Investigate if we need all components. Some of the selectors might not exist anymore.

wilkus27 (approved):
Everything looks good :slightly_smiling_face: I left one note to do in the future.

Issue #3818, opened 2021-04-14
contemporary math - learning objectives


lwtchu:
Documentation: #2190

lwtchu:
![Screen Shot 2021-11-03 at 12 01 57 PM](https://user-images.githubusercontent.com/11299125/140121818-4815dc9f-6220-4eb5-8fce-5fc87b15e05f.png)


This isn't displaying as it should be in the recently generated PDF on CORGI 
Job: 2082

kswanson33:
![image](https://user-images.githubusercontent.com/26280712/142071453-8b25e367-604d-4232-ae8a-d6d5625e2a79.png)
html structure of the learning objectives seems to be a bit funky. Styles expects:
1. learning objectives to be wrapped in an `<abstract>`
2. the nested list to be `ul`, not `ol`

Should this be a content fix, or should I code a workaround?

lwtchu:
Update - request to add style functionality for the section with a class. Content also updated to have no enumerated list in tagging

MartaSztuka:
@lwtchu there's still a problem with learning objective
##### baked locally on w.w.
![Screenshot from 2021-11-18 12-42-18.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/219d94d3-c032-4d5b-a879-551184ad9ee4)
![Screenshot from 2021-11-18 12-45-13.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f9814434-8af2-4d0d-83b4-b5cd8593d237)
##### CORGI (job 2366) 
![Screenshot from 2021-11-18 12-48-31.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/f07fa690-6244-4e83-903a-4ed82de7c7a9)
##### CORGI (job 2371) - I've repeat job 2082 from issue description, there are errors.
![Screenshot from 2021-11-18 12-55-31.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/c9d32303-c29c-46e5-b2a3-f105e7a11b5e)

kswanson33:
This looks like a content issue: I can tell because the LO list is numbered in the screenshots. For my development purposes, I had Larissa change the LO list to an unnumbered list, so that styles can work.

[...]

PR #4230, opened 2021-11-17
fix learning objectives type for contemporary math

kswanson33:
Also adds an LO selector for all of corn

#3818

image


wilkus27 (approved):
Beside additional comment everything is good :relaxed:

Issue #4237, opened 2021-11-19
Appendix Glossary headers in World History are missing styles


ajluk:
During testing World History kitchen epic issue Marta has found out that the headers in Appendix Glossary are unstyled: https://github.com/openstax/cnx-recipes/issues/4125#issuecomment-974212132. As I have checked the shape for Appendix Glossary is missing the headers styling in cosmos theme.

ajluk:
@lwtchu Here is how it looks now: 
![image](https://user-images.githubusercontent.com/55203598/142663979-fbe9d95c-15b1-4942-a745-66fca74c6157.png)
There are also the specs visible on the screenshot: 
![image](https://user-images.githubusercontent.com/55203598/142661838-50317588-963b-4f63-adb5-e5383a29577c.png)
Should I change something?

lwtchu:
@ajluk oh nice thank you <3 this looks good

MartaSztuka:
##### PASSED
![Screenshot from 2021-11-23 11-17-49](https://user-images.githubusercontent.com/47241114/143006890-3d53576c-a590-4fd0-b440-8373c185de8c.png)

PR #4238, opened 2021-11-19
style h3 headers

ajluk:
Issue: #4237


styles/designs/cosmos/parts/_titles-components.scss
```diff
@@ -138,6 +138,18 @@ $H3--LearningObjectives: (
...
+  _subselector: ' > h3[data-type="title"]',
+  _properties: (
+    color: enum('ValueSet:::REQUIRED'),
+    font-family: enum('ValueSet:::REQUIRED'),
+    font-size: font-scale(1),
+    line-height: line-height(1),
+    margin-bottom: v-spacing(1),
+    font-weight: 600,
+  )
+);
```
wilkus27, line 152: I'm wondering is there a need to create new component? It has the same properties as regular H3 :thinking:

wilkus27 (changes requested):

styles/designs/cosmos/parts/_titles-components.scss
ajluk: Yes, it has but I wasn't sure if it will, so I have left it this way. Thanks for noticing! :slightly_smiling_face:

wilkus27 (approved):

PR #4268, opened 2021-11-30
Add annotation styling

jarosik10:


styles/designs/webview/parts/_common-components.scss
```diff
@@ -0,0 +1,35 @@
...
+  _name: "StudentSampleText",
+  _subselector: '.student-sample',
+  _properties: (
+    font-family: enum('ValueSet:::REQUIRED'),
+    padding: '1rem 2rem',
+    margin: '1rem 0 2rem',
+    border-left-style: solid,
+    border-left-color: enum('ValueSet:::REQUIRED'),
+    border-left-width: 0.2rem,
+  )
```
wilkus27, line 34: In styles for PDF this part is a little bit messy :grin: These are a new elements so we wasn't sure how to handle this.
What do you think to clean up this here. All of these elements are annotations. So I would suggest to create two shapes: `AnnotationShape` and `AnnotationWithBorderShape`. Also two components: `Annotation` and `Annotation--WithBorder`. They should have empty subselector, so then in settings we can specify para with class. And in that way `.annotation-text` would use `AnnotationShape` and `.public-domain` and `.student-sample` - `AnnotationWithBorderShape`.

Also I think we should create regular files for annotations, not keep them in `common-components`. I think in Baked Styles it was done that way, because at the beginning there was only one component, so someone decide it's better to keep it in common catalog then create bunch of new files.

@ajluk what is your opinion?

wilkus27 (changes requested):
I have some suggestions, but I want to know your opinion too @jarosik10 @ajluk

styles/designs/webview/parts/_common-components.scss
jarosik10: @wilkus27
I was thinking about a approach like that, but what about the fact that `.public-domain`, `.student-sample` don't have `color` property? Should we just set the color to the default color from rex?

jarosik10: Sounds good.

ajluk: @wilkus27 As we discussed the solution you have proposed could be a better way to handle such components, shapes in the future, so I agree off course :slightly_smiling_face:

wilkus27 (approved):

PR #4270, opened 2021-11-30
add map files to gitignore?

kswanson33:
Experimenting with removing `map` files. Arguments in favor are:
1. Map files sometimes create unimportant merge conflicts 
2. They aren't very useful, + they're long and create extra fluff in the repo
3. The map files can still be generated by devs locally by compiling the style if we need them

The purpose of a map file is to help trace a chunk of `css` back to the `Sass` that created it. However, since we have so many files & imports in this repo, the map file rarely points to useful lines of `Sass` code (see below). Given that point 3 above can always be used in a pinch, I think there's no real reason to include `.css.map` files in this repo.

image


LostMojo:
These seem like good enough reasons to me

Stackblocks:
Seeing as they are not load-bearing code and they can be generated locally if needs be I'm pro ignoring them.

wilkus27:
Looks very good to me :wink: :+1: The only benefit of these merge conflicts is that it force us to regenerate files for all books, what is good, because sometimes when we make a change that affects entire theme we forgot about it and generate styles only for book that we're working on. So then there are unrelated changes in PRs that are created for other issues. But probably we should find some another solution for that problem :smile: Maybe some reminder or something like that. I don't have a specific idea for now :thinking:

wilkus27:
@kswanson33 we discussed this with @ajluk and we are not sure what would be the best. If tests will take as much as generating styles locally (more less half an hour) it might be bothersome :disappointed: We are wondering if it's possible to add versions to styles too. Maybe that would help. When someone is making theme change doesn't have to bother about all books, because they would be styled with one specific version of styles. We could update styles versions for books once in sprint or month. The question is if it's possible to add something like versions to this workspace :thinking:

PR #4277, opened 2021-12-02
Fix build-styles & add timer

kswanson33:
As I was looking into adding a css compile check to our workflow (result: since it takes 20+ minutes to recompile the whole library, it isn't practical right now) I found some opportunities for small improvements.

1. The `build-styles` script was compiling things multiple times; this fixes that problem.
2. Added timing info for `build.js`


script/build-styles
```diff
@@ -6,27 +6,30 @@ arg1=$1
 
 styles_root="./styles/books"
 
-# Pull in the BOOK_CONFIGS
-source ./books.txt || exit 1
```
kswanson33, line 10: Getting recipe names from books.txt was causing a couple of problems: first, that anytime a recipe was listed multiple times (i.e. calculus is listed 3 times for each of its 3 volumes), its style would be needlessly compiled multiple times. Second, using the directory structure automates the process of searching for & compiling styles based on what is actually there, and prevents data entry errors from causing problems (misspelled recipes, missing recipes).

styles/build/build.js
```diff
@@ -49,6 +50,8 @@ try {
     outputStyle: 'nested',
     outFile: outputFile
   })
+  let time2 = new Date().getTime()
+  console.log("time to render (s): " + ((time2-time1)/1000))
```
kswanson33, line 54: Having output here shouldn't impact anyone but us, since I don't believe richb-press ever calls this script. It was useful for me when investigating ways to make things faster, and I think it'll continue to be useful to us going forward.

LostMojo (approved):

PR #4288, opened 2021-12-06
add styles to hide components visually but not from screenreaders

kswanson33:
also imports these styles into English Composition


ajluk (approved):
Looks good! It is a cool way to visibly hide elements! (I learned about it while reviewing :slightly_smiling_face: )

Issue #4289, opened 2021-12-06
References need restyling after Kitchen changes made to alphabetize them


ajluk:
The section tags around references paragraphs has been removed in Kitchen during sorting references alphabetically:
![image](https://user-images.githubusercontent.com/55203598/144890635-dbd59fd1-acfc-48f6-90e7-cc406ce52886.png)
The shape used to create indention in cosmos theme used section wrappers so the styles need to be changed for References in Philosophy.

ajluk:
@MartaSztuka during working on this issue a part of the clean up in modules with indentation shapes has been done. For testing purposes it will be safer to test more that one book from cosmos theme. The good choice could be English Composition, where sections Works Cited, Further Reading and Works Consulted from EoC need to be checked and also sections Works Cited and Further Reading in Handbook pages need to be tested. The third book could be Political Science because it is one of the new books (section Suggested Reading in the EoC needs checking).

MartaSztuka:
#### PASSED
##### Philosophy
![Screenshot from 2021-12-08 13-04-39](https://user-images.githubusercontent.com/47241114/145205596-ecc184d5-6be8-48e8-8cbd-1b27377357d9.png)

##### Political science
![Screenshot from 2021-12-08 13-02-52](https://user-images.githubusercontent.com/47241114/145205416-38fc1244-7175-470b-83d0-5b2e77c4c0f8.png)
![Screenshot from 2021-12-08 13-03-01](https://user-images.githubusercontent.com/47241114/145205454-f8785035-09da-46d7-8cbf-6cc1ccb0c2c4.png)

##### Writing guide
![Screenshot from 2021-12-08 13-07-03](https://user-images.githubusercontent.com/47241114/145205846-7d75021c-0d23-4329-99d0-98eac4ad238c.png)
![Screenshot from 2021-12-08 13-06-27](https://user-images.githubusercontent.com/47241114/145205852-b2caa98c-baf9-4bb7-bd82-91589fb24891.png)
![Screenshot from 2021-12-08 13-08-23](https://user-images.githubusercontent.com/47241114/145205975-d87b4f03-b2a5-4b10-8561-11a00d137b80.png)

Issue #4155, opened 2021-10-11
Clean up cosmos theme: Check if shapes used for indentation styling can to be refactored in common one.


ajluk:
Currently there are two shapes used for styling sections with indentation in cosmos theme(`ModuleWithIndentationShape`, `SectionWorksCitedShape`). Within working on #4138 another one  has been created to avoid sectors overwriting and need to add more special classes: `HandbookModuleWithIndentationShape`. It needs to be checked during the clean up if these shapes can refactored and combined as one.

lwtchu:
@ajluk since this is merged shoudl we close it

ajluk:
@lwtchu The PR referenced under this issue is only related one, so the issue should remain open until someone will look into the shapes mentioned in the descriptions and and refactor them if it is possible.

lwtchu:
@ajluk just doing some clean up - i think we missed this 🤔 do you remember what this card was for

lwtchu:
@ajluk okay i'll move this to a later sprint

ajluk:
During working on #4289 a part of the clean up `ModuleWithIndentaion` shapes has been done. The HanbookModuleWithIndentaion shape has been removed and the EoC `ModuleWithIndentaionShape` has been moved out from eoc files to module files to clarify  it can be also used by sections other than EoC (f.ex. Hanbook sections in English Composition ). Two shapes using indentation remained untouched because they contain inner elements like titles (SectionWorksCitedShape): 
![image](https://user-images.githubusercontent.com/55203598/145192151-390cc652-01e2-4087-991e-aea7ecd22e5f.png)
 and chapter area(EobReferencesWithoutNumberingShape) :
![image](https://user-images.githubusercontent.com/55203598/145192237-8604748e-6988-4e87-a51b-dd9ff4ba2c81.png)
 During working on this issue it will be needed to analyze if we can create one generic shape for EoB and EoC modules with indentation or it will be better to keep the remaining 3 shapes as they are.

LostMojo:
`HandbookModuleWithIndentationShape` no longer exists, and it looks like the common components in the other two styles will be handled by the big refactor!

PR #4291, opened 2021-12-06
Restyle chapter references for philosophy

ajluk:
For: #4289
Also partly covers: #4155


styles/books/english-composition/book.scss
```diff
@@ -175,8 +175,8 @@ $bandColor: #DCB83D;
 @include use('UnitOpener', 'cosmos:::UnitOpenerShape');
 @include use('PunctuationEmphasis', 'cosmos:::PunctuationEmphasisShape');
 @include use('SectionWorksCited', 'cosmos:::SectionWorksCitedShape');
```
ajluk, line 177: I have left this shape untouched because it contains titles with specific styling only for section Works Cited in Handbook a page: 
image
If we would like these sections to utilize `ModuleWithIndentationShape` the titles need to be extracted from this shape or contains properties with OPTIONAL values set in book.scss file and I thought both solutions seems too much effort in this case, because this `SectionWorksCitedShape` is only used in English Composition.

styles/books/english-composition/book.scss
```diff
@@ -175,8 +175,8 @@ $bandColor: #DCB83D;
...
-@include use('HandbookWorksCited', 'cosmos:::HandbookModuleWithIndentationShape');
-@include use('HandbookFurtherReading', 'cosmos:::HandbookModuleWithIndentationShape');
+@include use('HandbookWorksCited', 'cosmos:::ModuleWithIndentationShape');
+@include use('HandbookFurtherReading', 'cosmos:::ModuleWithIndentationShape');
```
wilkus27, line 179: I'm not sure about this. Handbook it's a EOB page like Appendix. This shape is for EOC sections. I think we should come to this later and create general shape for modules with indentations, that can be used for EOC and EOB. And probably it should be done in a separate PR.

wilkus27: Or add comment next to this shape in `eoc-shapes.scss` file that it should be moved out of EOC :thinking: All handbook fies are already removed, so maybe that make no sense to add them again :smile: But I would keep issue #4155 opened. We can add a comment there too.

wilkus27 (changes requested):

styles/books/english-composition/book.scss
ajluk: @wilkus27 Thank you! I was going to move out the ModuleWithIndentationShape from EoC, but I just forgot at some point. I have move it out within the next commit. I will leave a comment in a clean up issue about it.

wilkus27 (approved):
Everything looks good :+1:

PR #4311, opened 2021-12-16
add devcontainer for styles devs

kswanson33:
To test out:

1. VSCode may give you a "reopen in container" button when it detects the devcontainer.json file
2. If not, click the green icon in the lower left corner of VSCode
image

Then select "Reopen in Container" from the menu at the top.

image

Try running a few scripts in the container, such as the build script for an individual style, or `./script/build-styles`

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,13 @@
+
+{ 
+    "name": "Styles Devcontainer",
+    "build": {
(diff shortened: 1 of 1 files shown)


LostMojo:
So others don't make the same error I did: 

If when you try to start up your docker container, it runs into an error, but doesn't seem to be setting up a _new_ docker container when you check the docker's log--

you might need to run `docker system prune -a` to get rid of any past build images before restarting your container.

Issue #3952, opened 2021-06-22
Contemporary Math - Tables in solutions crowding text


Stackblocks:
# Brief Description of Scenario and Screenshots
What kind of issue is this? [ ] recipe [ ] styles [ ] both [ x ] i don't know

![2021-06-22_11-59-54.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/5507b3e5-746c-4d23-91ae-0f760119989e)

# Does this happen for every instance? If no, is there a discernable pattern of occurrences?
Happens consistently

# Sample mark up if applicable

# Acceptance Criteria

# Bug Location(s)
(leave blank if the same as above)
collection ID: col10008
server: content04
chapter/section: 1.1
surrounding text: Notice that the last bin doesn’t follow

# Test Content Location(s)
(leave blank if the same as above)
collection ID:
server:
chapter/section:
surrounding text:

lwtchu:
I think these are tables specifically in LISTS. Maybe we can add a lil more padding

lwtchu:
Broken content can't work on this yet

LostMojo:
Caused by the table being a sibling to the raw li text

MartaSztuka:
@LostMojo spacing between tables and paragraphs is still too small.
![Screenshot from 2022-01-12 15-04-25.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/ea67d914-06b3-4a84-a68e-be6276e4a211)![Screenshot from 2022-01-12 15-04-40.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/a552ddc6-d1fd-4b2d-9866-b525eba02f85)

lwtchu:
@LostMojo what spacing is this at now :eyes:

PR #4335, opened 2022-01-04
3952 math table spacing

LostMojo:
#3952

### DIFF
--- a/styles/designs/corn/parts/_table-components.scss
+++ b/styles/designs/corn/parts/_table-components.scss
@@ -3,7 +3,6 @@ $Table__Container: (
     _subselector: "",
     _properties: (
         margin-bottom: table-v-spacing(2),
-        margin-top: table-v-spacing(1),
         display: enum('ValueSet:::OPTIONAL'),
         margin-left: enum('ValueSet:::OPTIONAL'),
         margin-right: enum('ValueSet:::OPTIONAL'),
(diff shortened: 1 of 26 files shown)


LostMojo:
Screen Shot 2022-01-04 at 3 19 21 PM

styles/designs/corn/parts/_table-components.scss
```diff
@@ -3,6 +3,7 @@ $Table__Container: (
     _subselector: "",
     _properties: (
         margin-bottom: table-v-spacing(2),
+        margin-top: table-v-spacing(1),
```
wilkus27, line 6: I think we should be more specific :thinking: There is no space between table and text in certain situations. This change is added to all tables. If table be after element with bottom padding (e.g para in module) there will be double space there (bottom padding of para and top padding of table).

wilkus27 (changes requested):

styles/designs/corn/parts/_table-components.scss
ajluk: Natalia has already written what I'm concerned about as well. Maybe we can tag such tables with specific class in content and add top margins only to them? :thinking:

LostMojo: So I've changed it to only take effect in tables in lists-- but due to how the html is set up, there's not a way to always catch when there's not text above a list-bound table-- I think it should be fine.

LostMojo: I've limited created a different shape to only target tables in lists-- this should cover most of our bases, and Larissa says we probably don't have to worry too much about the edge case of a list item starting with a table!

wilkus27 (approved):

PR #4361, opened 2022-01-12
Spanish Stats - Paragraph and list titles

KarinaMendez2:
for: https://app.zenhub.com/workspaces/ce-styles-5cdc84baee941c07853fbbf2/issues/openstax/cnx-recipes/4332

**Paragraph title example:** 
image

image

**List title example:** 
image

### DIFF
--- a/styles/design-settings/corn/parts/_titles-settings.scss
+++ b/styles/design-settings/corn/parts/_titles-settings.scss
@@ -167,3 +167,15 @@
         ),
     ),
 ));
+
+@include add_settings((
+    ListTitles:(
+        _selectors: ('[data-type="list"]'),
+    ),
+));
+
+@include add_settings((
+    ParagraphTitles:(
+        _selectors: ('p'),
+    ),
+));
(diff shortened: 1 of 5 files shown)


KarinaMendez2:
**Paragraph title example:** 
image

image

**List title example:** 
image

styles/design-settings/corn/parts/_titles-settings.scss
```diff
@@ -167,3 +167,15 @@
         ),
     ),
 ));
+
+@include add_settings((
+    ListTitles:(
+        _selectors: ('[data-type="list"]'),
+    ),
+));
```
LostMojo, line 175: I worry that these shapes might be a little too broad? Do we want them the same even if they're in a note or an exercise?

KarinaMendez2: according to the card, it should target any list and paragraph title in the book, can you pls confirm? @lwtchu

lwtchu: Yeah if it helps with context - list and para titles are seldom used cause we tried to ban it. But of course, some people still use them. Occasionally in some notes there are paragraph titles, but they will look like this - bold, left aligned, same as base font and with a line return after. I think this is ok to apply across the book.

LostMojo (approved):

Issue #4391, opened 2022-01-24
Problems with spaces in EOC


wilkus27:
1. Additional space after section Summary
2. Not enough space after modules with exercises

Look: https://github.com/openstax/cnx-recipes/issues/4340

wilkus27:
Problem with Summary is fixed for books from labels. Spaces after exercises sections are added to all books in Cosmos and Cardboard theme.

Issue #4340, opened 2022-01-06
0013-Philosophy EOC sections space issue


srm-openstax:
We have received the below corrections from customer. Kindly check and resolve these issues.

**Why is there an empty space after the summary but not after any of the other EOC sections. They should run one after the other.**
**There should be more space between the end of the questions and Further Reading.**
**There should be more space between the end of Further Reading and References.**

![Philosophy_EOC_Section](https://user-images.githubusercontent.com/86622862/148325392-9481d2ac-643e-4f74-9efa-b7103ef82e20.png)

# Bug Location(s)
collection ID: col00005
github openstax/osbooks- link: https://coffee-grouse-ulafn81g.ws-us25.gitpod.io/
chapter/section: Chapter 2/Section 6 (Global)
surrounding text: What is a thesis statement, and how should you go about developing

lwtchu:
@srm-openstax We can take a look but please bear in mind that some page break issues are difficult to control in an auto-generated setting

lwtchu:
# Acceptance Criteria

Can we investigate if there's any page break rules around the EOC sections that might be driving this spacing? We might have to tweak them a bit

wilkus27:
@lwtchu before I start working on that I want to make sure if this is intentionall that Summary is before Key Terms?

![image](https://user-images.githubusercontent.com/40228854/149152326-3b5065d9-0050-4f22-9919-f202700b5ba1.png)

Or should I change that to be consistant with other books?

lwtchu:
@wilkus27 i think this is fine. :eyes:

wilkus27:
Thank you @lwtchu! I fixed problems with spaces for this book and in the meantime I noticed that problem with empty space after Summary appears also in Finance and Political Science. Also issue with no space after sections with exercises occurs in entire theme. So I created separate card for that https://github.com/openstax/cnx-recipes/issues/4391. But everything is fixed within one PR.

lwtchu:
@wilkus27 oh thank you for being so thorough ❤️

Stackblocks:
- [x] Displays OK with current code & content
![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/678f471b-b1d5-4b92-878f-697554a85d4a)

PR #4392, opened 2022-01-24
fix problems with spaces in EOC

wilkus27:
Issues:
1. https://github.com/openstax/cnx-recipes/issues/4340
2. https://github.com/openstax/cnx-recipes/issues/4391

### DIFF
--- a/styles/books/english-composition/book.scss
+++ b/styles/books/english-composition/book.scss
@@ -142,7 +142,7 @@ $bandColor: #DCB83D;
 
 @include add_settings((
   DiscussionQuestions: (
-    _selectors: ('section'),
+    _selectors: (''),
     'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
   ),
 ));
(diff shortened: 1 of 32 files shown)


styles/books/english-composition/book.scss
```diff
@@ -142,7 +142,7 @@ $bandColor: #DCB83D;
 
 @include add_settings((
   DiscussionQuestions: (
-    _selectors: ('section'),
+    _selectors: (''),
     'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
```
wilkus27, line 146: Discussion Questions is untypical part of book that exists just in the English Composition. It's just section without any class at the end of regular page:

image

Because of that I decided to create a new shape, just for the injected exercises within page. Currently it was styled by `ExercisesModuleShape` which is used for EOC sections. In this PR I'm adding margin after those sections, so here it would be added to selector `section` and this is not what we want :wink:

ajluk (approved):

Issue #3555, opened 2021-02-24
polish u-pysics: regression testing


wilkus27:
Solution for additional space after paragraphs which contain footnotes references (https://github.com/openstax/cnx-recipes/issues/3315) might cause some regression. So pl-u-physics should be tested page by page.

Link to PR: https://github.com/openstax/cnx-recipes/pull/3540

MartaSztuka:
#### Fizyka vol 2
##### @wilkus27 there's double space in EoC chapter 1, after question 67.![Screenshot from 2021-03-23 16-19-44.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/4c36fab5-54c2-49d0-aa98-477ffc5568a7)
##### Answer Key chapter 7 ![Screenshot from 2021-03-24 17-31-23.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/de93d89f-1937-4ae0-a5f6-009ca342db0f)

wilkus27:
It happens because settings for paragraphs from problems and solutions overwrites settings for `p.has-noteref`:

![image](https://user-images.githubusercontent.com/40228854/112614292-6127df00-8e21-11eb-9c96-ce12b4ed6760.png)

lwtchu:
@wilkus27 can this be closed or should i put it in the backlog?

wilkus27:
@lwtchu we can put this to the backlog. Fixing selectors doesn't help, problem still occurs. We need to spent more time on this to figure out why it doesn't work. We discussed with team that it's not urgent, so we come back to this later.

lwtchu:
@wilkus27 thanks !

wilkus27:
After updating to Prince 14, there is no more problem with empty space after exercise:

![image](https://user-images.githubusercontent.com/40228854/151555807-67b05c3f-9dd9-4571-a8d2-4d7d94d81695.png)

But footnote is in the column not at the end of the page.

In Answer Key it looks good:

![image](https://user-images.githubusercontent.com/40228854/151556031-ea80f142-465a-4e6e-8dbc-aec25aeee4ce.png)

ajluk:
Screenshot for reference: 
![image](https://user-images.githubusercontent.com/55203598/151586682-9ae251ba-7e87-4673-a757-f6fbc18457e5.png)

ajluk:
[...]

PR #4417, opened 2022-01-28
change footnote float properties

ajluk:
For: #3555

### DIFF
--- a/styles/designs/common/pdf/_footnotes-components.scss
+++ b/styles/designs/common/pdf/_footnotes-components.scss
@@ -2,7 +2,7 @@ $Footnote__Container: (
   _name: "Container",
   _subselector: '',
   _properties: (
-    -prince-float-placement: footnote,
+    float: footnote,
     -prince-float-reference: page,
     color: enum('ValueSet:::REQUIRED'),
     line-height: enum('ValueSet:::REQUIRED'),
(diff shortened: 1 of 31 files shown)


styles/designs/common/pdf/_footnotes-components.scss
```diff
@@ -2,7 +2,8 @@ $Footnote__Container: (
   _name: "Container",
   _subselector: '',
   _properties: (
-    float: footnote,
+    -prince-float-placement: footnote,
+    -prince-float-reference: page,
```
ajluk, line 6: In Prince 14 footnotes in multicolumn layout are floating to the bottom of the column instead of the bottom if the page, this is a default behavior which can be changed by split up the float shorthand property into two: placement and reference properties as in the lines from above.
Link to documentation: https://www.princexml.com/doc/styling/#floats

styles/designs/common/pdf/_footnotes-components.scss
```diff
@@ -3,6 +3,7 @@ $Footnote__Container: (
   _subselector: '',
   _properties: (
     float: footnote,
```
ajluk, line 5: I have changed back `-prince-float-placement` property to `float: footnote` to add support for Prince 12 if we would like to maintain it.

wilkus27 (approved):

PR #4426, opened 2022-01-31
build.js to build.dart

KarinaMendez2:


styles/build/build.dart
```diff
@@ -1,14 +1,54 @@
...
+    throw Exception('Cannot find styles root directory.');
+  }
+
+  Map<String, String> envVars = Platform.environment;
+
+  var platform = envVars["PLATFORM"] ?? 'none';
+  var platformIncludesPath = "${stylesPath}framework/platform/${platform}";
+  if(File(platformIncludesPath).existsSync()){
+    throw Exception("Specified platform '${platform}' does not have an associated directory.");
+  }
```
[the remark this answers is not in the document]
KarinaMendez2: the [platform class ](https://api.dart.dev/stable/2.16.1/dart-io/Platform-class.html) is what I found to retrieve information about the environment, do you mean to just skip/comment out these lines or to change how the env variables are set?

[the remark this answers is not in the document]
kswanson33: oh yeah that's confusing with `Platform` being both a class to access environment variables and the name of our env variable. I was talking about the `PLATFORM` env variable, not the class. Since `PLATFORM` isn't really used in node (it was used back when they tried to do web styles, but I believe web vs. pdf is its only purpose), and probably won't be used in the future, I don't think we should translate it into dart. We can always add it back in if we end up using it, but from what I've been following about the web styles, it doesn't seem likely that we'll need an environment variable in order to compile them. does that make sense lol

styles/build/build.dart
```diff
@@ -1,14 +1,45 @@
...
+  var endTime = DateTime.now();
+  print("compile time is: ${endTime.difference(startTime)}");
+
+  if ((result.css.length > 0) & (outputFile.length > 0)) {
+    new File(outputFile).writeAsStringSync(result.css);
+    new File("${outputFile}.map").writeAsStringSync(json.encode(result.sourceMap.toJson()));
+  } else {
+    throw Exception("Empty output");
+  }
+}
```
LostMojo, line 45: no final newline? otherwise, looks good!

LostMojo (approved):

styles/build/build.dart
```diff
@@ -1,14 +1,45 @@
 import 'dart:io';
+import 'dart:convert';
 import 'package:sass/sass.dart' as sass;
+import 'package:source_maps/source_maps.dart';
+import 'dart:io' show Platform;
```
kswanson33, line 5: do we need this import still?

styles/build/build.dart
```diff
@@ -1,14 +1,45 @@
...
-}
+  var endTime = DateTime.now();
+  print("compile time is: ${endTime.difference(startTime)}");
+
+  if ((result.css.length > 0) & (outputFile.length > 0)) {
+    new File(outputFile).writeAsStringSync(result.css);
+    new File("${outputFile}.map").writeAsStringSync(json.encode(result.sourceMap.toJson()));
+  } else {
+    throw Exception("Empty output");
+  }
```
kswanson33, line 44: nice

kswanson33 (approved):

PR #4478, opened 2022-02-15
Update noteish elements styling

bethshook:
Updating baked styles to reflect style changes in this pr: https://github.com/openstax/rex-web/pull/1256/files

Would appreciate any feedback since I am new to the baked styles framework.

### DIFF
--- a/styles/designs/webview/parts/_learning-components.scss
+++ b/styles/designs/webview/parts/_learning-components.scss
@@ -12,7 +12,7 @@ $LearningObjectives__Container: (
 );
 
 $LearningObjectives__Container--Note: (
-  _name: "LearningObjectivesContainer",
+  _name: "LearningObjectivesNoteContainer",
   _subselector: ' .learning-objectives[data-type="note"]',
   _properties: (
     padding: 0,
--- a/styles/designs/webview/parts/_learning-components.scss
+++ b/styles/designs/webview/parts/_learning-components.scss
@@ -11,13 +11,20 @@ $LearningObjectives__Container: (
   )
 );
 
+$LearningObjectives__Container--Note: (
+  _name: "LearningObjectivesContainer",
+  _subselector: '&[data-type="note"]',
+  _properties: (
+    padding: 0,
+  )
+);
+
 $LearningObjectives__Title: (
   _name: "LearningObjectivesTitle",
   _subselector: ' [data-type="title"]',
   _properties: (
     display: block,
     margin-top: 0,
-    margin-bottom: 1rem,
     padding: '0 1.5rem 1rem',
     border-bottom-style: solid,
     border-bottom-color: enum('ValueSet:::REQUIRED'),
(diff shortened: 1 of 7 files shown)


styles/designs/webview/parts/_notes-components.scss
```diff
@@ -111,9 +113,26 @@ $Note__Body: (
...
     border-top: none,
     background-color: enum('ValueSet:::REQUIRED'),
     color: enum('ValueSet:::REQUIRED'),
+    overflow-x: 'auto !important',
+  ),
+);
+
+$Note__Body__FirstContent: (
+  _name: "NoteBodyFirstContent",
+  _subselector: ' > *:not(section):not(header):first-child',
```
LostMojo, line 126: Do you want to include span[data-type=media] in this? This may overlap with the Media--InNote component below

LostMojo: Otherwise, this looks very good!

styles/designs/webview/parts/_learning-components.scss
```diff
@@ -11,13 +11,20 @@ $LearningObjectives__Container: (
   )
 );
 
+$LearningObjectives__Container--Note: (
+  _name: "LearningObjectivesContainer",
```
wilkus27, line 15: Two components can't have the same `_name`. It also should be changed.

styles/designs/webview/parts/_notes-components.scss
```diff
@@ -3,22 +3,28 @@ $Note__Container: (
...
-    padding: 1.5rem,
     border-style: solid,
     border-width: 0.2rem,
     border-color: enum('ValueSet:::REQUIRED'),
     background-color: enum('ValueSet:::REQUIRED'),
   ),
 );
 
+$Note__Container__FirstExercise: (
+    _name: "ExerciseContainerFirst",
```
ajluk, line 14: A small thing, the name will be more descriptive if it will match the variable name: `FirstExerciseContainer`.

styles/output/webview-generic.css
```diff
@@ -305,9 +304,17 @@ ol {
...
   color: #555; }
 
+.learning-objectives[data-type="note"] {
+  margin: 3rem 0;
+  padding: 1.5rem;
+  border-style: solid;
+  border-color: #dcdcdc;
+  border-width: 0.2rem;
+  background-color: #ededed; }
+
```
wilkus27, line 317: Padding is not changed. Probably because this component has the same name as the first one (see my previous comment). We will see what happen when you change the name. I'm afraid changing just padding won't be enough. Probably you should create two shapes. One for learning objectives as note and one for learning objectives in another tag (section?). And each of the shapes should contain all needed components with all properties.

wilkus27 (changes requested):

styles/designs/webview/parts/_notes-components.scss
bethshook: Good catch - Bartek didn't touch that so I think I will leave as is.

wilkus27 (approved):
Everything looks good :relaxed:

PR #4495, opened 2022-02-17
add linter to devcontainer

kswanson33:
Originally I was trying to see if this would help us with the dart migration ... it didn't, but it still might be good to have. A couple thoughts:
- `sass-lint`, the old linter used in this repo, hasn't been supported since 2017
- `stylelint` is the new way to go, with its extension for `scss` linting (thanks Karina)
- The `npm` version has to be updated in order for the `scss` linter to work; this might require rebuilding node-sass to be compatible (not that we'll ... be using node-sass much longer). Not sure what other side effects it may have.

Usage:
image
- linter running in the container gives as visual of what problems are and (sometimes) offers a quick fix
- running `npx stylelint <filepath(s)>` gives a list of violations & line numbers
- running the above with the `--fix` flag will automatically fix anything that is ... automatically fixable, like single quotes -> double quotes and some indentation (although it gets confused about indentation depending on how bad it is). I've found it eats comments sometimes tho so I've had to be careful to manually put the comments back in

Do we need a linter? How should we use it?
- I personally like linting & having standardized code format, but it isn't a huge deal
- Since we have so many unlinted files, making linting part of CI doesn't seem feasible yet
- We could do something like cookbook does & test committed files for linter issues in CI ... but even that might be annoying for a while. I've included `accounting/book.scss` as an example of all the changes I had to make to get a regular file down to 0 linter issues.
- We don't have to run linter stuff in CI at all to benefit from a linter. Just having it running in the devcontainer will give us some formatting rules & we can choose to ignore them if we want
- but also it might be nice to have some linter enforcement _eventually_
- running `npx stylelint <filepath> --fix` solves a lot of problems quickly too


.node-version
```diff
@@ -1 +1 @@
-10.15.3
+14.16.1
```
kswanson33, line 1: used the same version as enki

package.json
```diff
@@ -18,14 +18,14 @@
     "sass-true": "^4.0.0",
     "sassdoc": "^2.6.0",
     "standard": "^12.0.1",
-    "stylelint": "^9.10.1",
+    "stylelint": "^14.5.0",
```
kswanson33, line 21: has to be later than 14 for compatibility w/ `stylelint-config-standard-scss`

.node-version
ajluk: @kswanson33 I have tried to run the one book script (`PLATFORM=pdf node ./styles/build/build.js ./styles/books/dev-math/book.scss ./styles/output/dev-math-pdf.css`) on your branch just to check if changing the node version will affect it, and it is unfortunately: 
image

ajluk: The same error also occurs while running the general styles script (`./script/build-styles`): 
image

ajluk: The both scripts were working after running `npm rebuild node-sass`, but the error I pasted via the slack regarding not working linter script still persists :disappointed: 
image

ajluk (commented):
@kswanson33 As I mentioned on the slack thread I couldn't try the actual linting, because of the below error: 
image, but I have tried styles scripts on this branch because of the changing of the node version and left comments (I needed also update node-sass). Otherwise it all looks really good! :slightly_smiling_face:

lwtchu:
@kswanson33 should we set this to a sprint?

kswanson33:
@ajluk I was finally able to reproduce this bug. It had something to do with dependencies not being installed in `node_modules`. Deleting my `node_modules` folder & running `yarn install` fixed the bug. I checked, and reopening the container does not automatically create `node_modules`, which is odd because `./script/setup` does run `yarn install`. Digging a little deeper, I found out that `node_modules` is in `.dockerignore`. I'm going to do some reading about that and decide what to do to keep the container set up well, but in the meantime, you should be able to run the linter after running `yarn install`.

script/setup
```diff
@@ -5,10 +5,11 @@ source ./script/bootstrap || exit 111
 # See https://github.com/openstax/setup-scripts for details on this file
 
 # JavaScript stuff
-if [[ -f "./package.json" ]]; then
-  do_progress_quiet "Installing JavaScript Packages" \
-    yarn --prefer-offline
-fi
```
ajluk, line 11: @kswanson33 I have only one question about installing packages outside the container. If these lines are commented out does it mean that we need to update manually the packer if will be not working in the container for some reason?

Issue #4505, opened 2022-02-21
contemporary math - exercise alignment


lwtchu:
![Screen Shot 2022-02-21 at 5 00 07 PM](https://user-images.githubusercontent.com/11299125/155035929-66924a32-871e-43bf-ad7c-f4919628defd.png)

https://github.com/openstax/osbooks-contemporary-math 

I believe this is an injected exercise. Is there a way to align the text for exercises even if the numbers are double digit? We can also talk options and talk if this impacts more books.

ajluk:
@lwtchu Here is how the change in the exercise numbers styling looks like in pdf in contemporary math:

1. When `.os-number` (the link with exercise number) has width set to **2%**:

![image](https://user-images.githubusercontent.com/55203598/156202367-abc6d443-f529-4861-8cdc-edfb7778c724.png)

   - within 3 digits number case (it looks worse there-see first number):
   
![image](https://user-images.githubusercontent.com/55203598/156202521-9a13250c-7c93-45e6-aefd-6732111a4457.png)

2. When the width is set to **3%**:
![image](https://user-images.githubusercontent.com/55203598/156202846-18f923fa-ce82-42c5-b2d2-8413160114e6.png)

  - it looks good also for 3 digits, but off course the number takes more place 
![image](https://user-images.githubusercontent.com/55203598/156203106-43285831-7f1e-40b5-a45d-cdf56feb4693.png)

ajluk:
The decision was to fix the styling only for one column exercises and answer key, because the fix is not working in 2/3 column exercises/answer key.

ajluk:
@MartaSztuka please test exercises and answer key in Contemporary Math (also content/section exercises at the end of the modules) Chemistry, Statistics, Accounting (one of the volumes). @lwtchu should we test more books? I didn't mention other math books because there the exercises are in 3 columns, simmilar situation is in physics (2 columns).

ajluk:
@MartaSztuka the solution which worked for 3 digit numbers didn't properly with a large amount of math and tables in Chemistry, so there was a decision to apply the solution which works only for 2 digits.

ajluk:
@lwtchu Could you take a look and advice if we need to do more tests or test more books? :slightly_smiling_face:

lwtchu:
@ajluk i think this is good and very thorough :)

PR #4525, opened 2022-03-04
fix exercise numbers styling one column exercises all themes

ajluk:
Issue: #4505

### DIFF
--- a/styles/books/american-government/book.scss
+++ b/styles/books/american-government/book.scss
@@ -64,7 +64,7 @@ $bandColor: #DCB83D;
 //EoC
 @include add_settings((
   ExercisesModule: (
-    _selectors: ('.os-eoc.os-review-questions-container', '.os-eoc.os-critical-thinking-container', '.os-eoc.os-suggested-reading-container'),
+    _selectors: ('.os-eoc.os-review-questions-container', '.os-eoc.os-critical-thinking-container'),
     'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
     'ProblemNumber:::width': 2%,
     'ProblemNumber:::text-align': right,
--- a/styles/books/american-government/book.scss
+++ b/styles/books/american-government/book.scss
@@ -61,14 +61,13 @@ $bandColor: #DCB83D;
     ),
 ));
 
+//EoC
 @include add_settings((
-  ReviewQuestions: (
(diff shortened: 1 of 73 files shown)


styles/books/american-government/book.scss
```diff
@@ -61,14 +61,13 @@ $bandColor: #DCB83D;
...
+//EoC
 @include add_settings((
-  ReviewQuestions: (
-    _selectors: ('.os-eoc.os-review-questions-container'),
-    'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
-  ),
-  CriticalThinking: (
-    _selectors: ('.os-eoc.os-critical-thinking-container'),
+  ExercisesModule: (
+    _selectors: ('.os-eoc.os-review-questions-container', '.os-eoc.os-critical-thinking-container', '.os-eoc.os-suggested-reading-container'),
```
ajluk, line 67: I did some cleaning while working on this issue, because I noticed that many exercises are using the shapes without any additional changes defined in book.scss files. Also there are no `after`, `before` elements in these shapes so using many selectors won't break anything.

wilkus27, line 67: Is selector for Suggested Reading intentional? It has different shape (ModuleWithIndentation).

styles/books/history/book.scss
```diff
@@ -62,13 +62,11 @@ $bandColor: #DCB83D;
...
 // EOC exercises
 @include add_settings((
-  ReviewQuestions: (
+  ExercisesModule: (
     _selectors: ('.os-eoc.os-review-questions-container'),
     'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
-  ),
-  CriticalThinking: (
-    _selectors: ('.os-eoc.os-critical-thinking-container'),
-    'ProblemNumber:::color': (_ref: "colorMap:::fontBodyColor"),
```
wilkus27, line 71: Selector for Critical Thinking is missing.

wilkus27 (changes requested):
A few small things :wink:

styles/books/american-government/book.scss
ajluk: Oh, no. Thanks Natalia!

wilkus27 (approved):

Issue #4545, opened 2022-03-16
Mikroekonomia - tables


MartaSztuka:
# Caption, title, number
## Table number
font: Archivo
weight: bold, 700
color: `#8f007e`
treatment: all caps
font size: font-scale(0): 9pt, 12px, 1rem

## Table title (if in the `<title>` tag)
font: Archivo
weight: Medium, 500
color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem

## Figure caption
font: IBM Plex Sans
weight: regular, 400
color: Black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem

# Type 1 Regular Table
![Screenshot from 2022-03-16 13-27-42](https://user-images.githubusercontent.com/47241114/158590130-8a2aa291-7d71-449d-9032-891aed9bcb8c.png)

## Header row 
font: IBM Plex Sans
font treatment: bold, 700
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: only top and bottom
cell border color: black `#000000`
cell border weight: .25px

## Non header rows
font: IBM Plex Sans
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: all around unless it is the cells on the left and right (see screenshot)
cell border color: black `#000000`
cell border weight: .25px

## Bottom stroke of table
color: `#8f007e`
cell border weight: 1.5px

# Type 2 top-titled
## First row in header row
This will be a fake title. 
font: IBM Plex Sans
font treatment: bold, 700
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
*alignment: centered*
*cell border: only bottom*
cell border color: black `#000000`
cell border weight: .25px

- specs below are the same as type 1
## Header row 
font: IBM Plex Sans
font treatment: bold, 700
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: only top and bottom
cell border color: black `#000000`
cell border weight: .25px

## Non header rows
font: IBM Plex Sans
font color: black `#000000`
font size: font-scale(0): 9pt, 12px, 1rem
alignment: content driven
cell border: all around unless it is the cells on the left and right (see screenshot)
cell border color: black `#000000`
cell border weight: .25px

## Bottom stroke of table
color: `#8f007e`
cell border weight: 1.5px

MartaSztuka:
[...]

PR #4552, opened 2022-03-17
change table colors in pl-economics

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/4545


styles/output/ap-physics-2e-pdf.css
```diff
@@ -1377,7 +1377,9 @@ a {
   display: table-cell;
   text-decoration: none;
   color: #000000;
-  font-weight: bold; }
+  font-weight: bold;
+  width: 2%;
+  text-align: right; }
```
wilkus27, line 1382: Probably changes from some previous PR.

ajluk (approved):

styles/output/ap-physics-2e-pdf.css
ajluk: These are changes from my previous PR actually :smile: I though the ap-physics-2e styles were created later, but maybe it was the other way round.

Issue #4543, opened 2022-03-16
Mikroekonomia - chapter intro


MartaSztuka:
The introduction for Economics is a little bit different so here's a card to capture what might be happening in the introduction of Economics. This will deviate from the cosmos theme a little, but still uses same specs for each piece.
![Screenshot from 2022-03-16 12-53-42](https://user-images.githubusercontent.com/47241114/158584377-0309d4ea-c644-4005-a2bf-593e94ec045a.png)

### Splash image
- [x] #4540 

### Bring It Home note
- [x] #4542 

### Introduction title
Introduction title follows the SAME header styling in #3164 for module titles
font style: Archivo
weight: semi-bold, 600
font-size: font-scale(3): 15.55pt, 20.4px, 1.7rem
color: R1: `#8f007e`

### Chapter Objectives
- [x] #4541

MartaSztuka:
#### PASSED
![Screenshot from 2022-03-21 14-18-38.png](https://images.zenhubusercontent.com/5cb5ec14fec3b63424e0f03c/86c59452-0dce-4f85-95f8-0fe0cf45d0b6)

PR #4564, opened 2022-03-18
change intro title color

ajluk:
For: #4543

### DIFF
--- a/styles/books/pl-economics/book.scss
+++ b/styles/books/pl-economics/book.scss
@@ -147,6 +147,11 @@ $bandColor: #217867;
 ));
 
 @include add_settings((
+  IntroductionPageTitles:(
+    _selectors: ('[data-type="page"].introduction'),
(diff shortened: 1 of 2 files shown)


styles/books/pl-economics/book.scss
```diff
@@ -185,7 +185,6 @@ $bandColor: #217867;
 // Titles
 @include use('PageTitles', 'cosmos:::PageTitlesShape');
 @include use('ChapterTitles', 'cosmos:::ChapterTitlesShape');
-@include use('IntroductionPageTitles', 'cosmos:::IntroductionPageTitlesShape');
```
ajluk, line 188: Within this PR I'm removing this shape include from book.scss file since pl-microeconomics doesn't need it. We can always add it back if these H3 titles will be present in intro pages.

ajluk: @wilkus27 Rest of the titles are included in in Chapter Intro shape, so I unsigned myself from the issue.

Issue #4575, opened 2022-03-22
College Physics 2e table separated from its title


oscryan:
# Brief Description of Scenario and Screenshots
<img width="733" alt="Screen Shot 2022-03-22 at 9 20 56 AM" src="https://user-images.githubusercontent.com/1895432/159503918-4c913b89-ebb7-452b-8654-7d5bb3b60c8e.png">
<img width="731" alt="Screen Shot 2022-03-22 at 9 20 50 AM" src="https://user-images.githubusercontent.com/1895432/159503924-8994f2cd-aabe-4b44-968a-ed012b59b8b2.png">

# Does this happen for every instance? Is there a discernable pattern of occurrences? Please describe below.
Only found one so far
# Acceptance Criteria

# Sample mark up if applicable
```
<!-- Please paste here. -->
```

# Location(s)
github openstax/osbooks- link: 
collection slug (if applicable):
chapter/section: 
surrounding text:

oscryan:
Also, very big table avoids breaking even when this makes the previous page very empty
<img width="732" alt="Screen Shot 2022-03-22 at 8 47 55 AM" src="https://user-images.githubusercontent.com/1895432/159521700-d6744012-b443-4b62-8cb9-bf934cffd561.png">
<img width="733" alt="Screen Shot 2022-03-22 at 8 48 03 AM" src="https://user-images.githubusercontent.com/1895432/159521710-296a43c4-9fc7-4127-99c4-b57af4d2cf70.png">

lwtchu:
To whomever working on this card: can you tell me what the rules are for the page breaks for tables? Ideally it should be able to break within the `<tbody>` and NOT `<thead>` and the number/caption is not orphaned

ajluk:
@lwtchu Here are the page breaks rules for all tables:

lwtchu:
@wilkus27 is there a way to write something where if it's not the first or last row, the table can break?

ajluk:
@lwtchu I guess you accidentally tagged Natalia, but I will check it out. :smile:

lwtchu:
@ajluk :joy: i'm so sorry i was messaging her in the other two cards I was on and just typed it out of habit bahhh I'm sorry

ajluk:
@lwtchu I have added page break rules to first and last tr like so:
![image](https://user-images.githubusercontent.com/55203598/160176448-a756c666-9719-4cce-a17d-8f461b5f0d07.png)
[...]

PR #4579, opened 2022-03-24
add caption prince specific properties

ajluk:
For: #4575

### DIFF
--- a/styles/designs/carnival/parts/_caption-components.scss
+++ b/styles/designs/carnival/parts/_caption-components.scss
@@ -9,7 +9,7 @@ $Caption__Container--Figure: (
         margin-right: auto,
         margin-bottom: v-spacing(1),
         font-size: font-scale(0),
-        line-height: 1.4rem,
+        line-height: 1.4rem, 
     )
 );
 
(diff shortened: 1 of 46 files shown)


styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -102,7 +102,10 @@ $Caption__Container--Table: (
     _name: "CaptionContainerTable",
     _subselector: ' .os-caption-container',
     _properties: (
-        font-size: font-scale(0)
+        font-size: font-scale(0),
+        display: enum('ValueSet:::OPTIONAL'), // Used in Basic Tables 
+        caption-side: enum('ValueSet:::OPTIONAL'), // Used in Basic Tables 
+        prince-caption-page: enum('ValueSet:::OPTIONAL'), // Used in Basic Tables
     )
 );
```
wilkus27, line 110: There is so many OPTIONALs. Shouldn't we create a separate component in this case?
I'm not sure about names :thinking: New component could be named by `$Caption__Container--BasicTable`, but existing one should be left as is, or it's better to rename it to `$Caption__Container--NonBasicTable`?

wilkus27 (changes requested):

styles/designs/carnival/parts/_caption-components.scss
ajluk: Natalia I didn't add NonBasicTable to the existing table component name, because there is also separate one used in top titled tables, so I left the name as it was.

wilkus27 (approved):

Issue #4528, opened 2022-03-08
Anatomy and Physiology - bullets in lists in tables falling out of the cell


yblock:
# Brief Description of Scenario and Screenshots
bulleted list inside of table item renders with bullet outside of the cell in PDF

<img width="653" alt="Screen Shot 2022-03-08 at 12 08 08 PM" src="https://user-images.githubusercontent.com/5865292/157299169-78e3bdcd-96e0-4a77-924d-a15ce1ae747a.png">

# Does this happen for every instance? Is there a discernable pattern of occurrences? Please describe below.
Appears to happen in all cases of bulleted lists within table cells

# Acceptance Criteria
bullet point placement will be adjusted so that it falls within its respective cell. 

# Sample mark up if applicable
```
 <list id="fs-id1491159" list-type="bulleted"><item>Secretes several hormones, including hCG, estrogens, and progesterone, to maintain the pregnancy and stimulate maternal and fetal development</item>
```

# Location(s)
github openstax/osbooks- link: https://github.com/openstax/osbooks-anatomy-physiology
collection slug (if applicable): anatomy-and-physiology
chapter/section: 
surrounding text: Secretes several hormones, including hCG

KarinaMendez2:
There's a very global style being applied to `[data-type="page"] ul`'s that sets `padding: 0;`, this is affecting all scenarios in the book, by removing this padding, the screenshoted sections in this card get fixed but other areas like: 

![image.png](https://images.zenhubusercontent.com/5c54789f0a972e6fd1ece1a0/21e8555a-d18a-475a-af40-31e1ce859971)

get affected. I also found a comment for this book saying:

![image.png](https://images.zenhubusercontent.com/5c54789f0a972e6fd1ece1a0/68453357-7776-4a41-8e5d-3b5674869bb8)

I suggest a styles cleanup for lists instead of targeting one by one scenarios.

ajluk:
@lwtchu I have added styles for lists in tables based on suggestion in a comment in code that the margin should be smaller in tables lists than in the lists in the rest of the book. There has been a function created for margins in tables - `table-h-spacing`. When I set it to `1` the margin is equal to 0.7rem (8.39px) but it doesn't look good: 
![image](https://user-images.githubusercontent.com/55203598/160160198-9d00f92b-6bd6-460a-a462-432f421ec935.png)
so I changed it to table-h-spacing(2) which is equal to 1.4rem (16.79px). Here is how it looks: 
[...]

PR #4581, opened 2022-03-25
add page list shape anatomy

ajluk:
For: #4528

### DIFF
--- a/styles/books/anatomy/book.scss
+++ b/styles/books/anatomy/book.scss
@@ -216,6 +216,9 @@ $ChapterIntroType: fullWidth;
 @include use('Footnote', 'common:::FootnoteShape');
 @include use('FootnoteCall', 'common:::FootnoteCallShape');
 
+// Lists
+@include use('PageLists', 'carnival:::PageListsShape');
+
 //EOC
 @include use('KeyTerm', 'carnival:::EocKeyTerm');
 @include use('EocSummary', 'carnival:::EocSummary');
--- a/styles/output/anatomy-pdf.css
+++ b/styles/output/anatomy-pdf.css
@@ -1706,6 +1706,39 @@ a[role="doc-noteref"] {
   top: -0.7em;
   padding-right: 0.16rem; }
 
+[data-type="page"] > ul:not([data-labeled-item="true"]) {
+  margin-left: 24px; }
+
+[data-type="page"] > ol {
+  margin-left: 24px; }
+
+[data-type="page"] > [data-type="list"] > [data-type="title"] {
+  font-weight: bold; }
+
+[data-type="page"] > [data-type="list"] > ul[data-bullet-style='none'] {
+  list-style-type: none; }
+
+[data-type="page"] > section > ol {
+  margin-left: 24px; }
+
+[data-type="page"] > section > ul:not([data-labeled-item="true"]) {
+  margin-left: 24px; }
+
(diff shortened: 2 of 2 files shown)


styles/designs/carnival/parts/_lists-components.scss
```diff
@@ -202,3 +202,36 @@ $TitledLists__Title: (
         font-weight: bold,
     )
 );
+
+// Tables lists
+// Lists in table cells need to have margin set withs table spacing functions
+
+$Table__UnorderedList--FirstLevel: (
```
wilkus27, line 209: Because we are in lists file, component name should start with list name. This is also what we do in other themes. If list for some element should be styled differently, we treat this as list of different type. So probably we should rename this to `UnorderedList--FirstLevel--Table`. The same for all components from below.

styles/designs/carnival/parts/_table-components.scss
```diff
@@ -216,5 +216,9 @@ $Table__BottomBorder: (
         position: relative,
         top: -0.1rem,
         margin-bottom: v-spacing(1),
-    )    
+    )
 );
+
+// Table data container used in table list shape
+$Table__Data--Lists: empty_wrapper(TableDataLists, ' > td');
+$Table--Lists: empty_wrapper(TableDataLists, ' > table');
```
wilkus27, line 224: Do we have to specify that it is created for lists? Maybe we can reuse empty wrappers in other elements later. Maybe we should mark somehow when component is just a wrapper :thinking: This is a proposition also for refactoring. I'm not sure what name convention would be the best.

- `Table--Wrapper`, `Table__Data--Wrapper` or
- `TableWrapper` , `Table__DataWrapper`

Let me know what do you think :wink:

wilkus27 (changes requested):
I have two suggestions :wink:

styles/designs/carnival/parts/_table-components.scss
```diff
@@ -216,5 +216,9 @@ $Table__BottomBorder: (
         position: relative,
         top: -0.1rem,
         margin-bottom: v-spacing(1),
-    )    
+    )
 );
+
+// Table data container used in table list shape
+$Table__Data--Wrapper: empty_wrapper(TableDataLists, ' > td');
+$Table--Wrapper: empty_wrapper(TableDataLists, ' > table');
```
wilkus27, line 224: Names in `empty_wrapper` function also should be changed. Now they are the same and have part `Lists`. If I remember correctly this function adds part `Wrapper` automatically :thinking: So when you put just `Table` it will be `TableWrapper`.

wilkus27 (changes requested):
One small thing :wink:

styles/designs/carnival/parts/_table-components.scss
ajluk: I chose the first option, seemed more clear to me. Thank you Natalia! And yes I was also thinking we should mark the empty containers/wrappers somehow, maybe we can pick the word wrapper for it, like a reserved one.

wilkus27: Ok :+1: I will add this point to General Rules https://docs.google.com/document/d/15CNiQQDEDo9M49IFkjhc9bTkzMZ6b1PXeEZ5J7LeQwI/edit# file.

wilkus27 (approved):

Issue #2288, opened 2020-05-07
Chapter headers need more space


MartaSztuka:
Refers both to ToC and Chapter header, especially when line breaks in two.
Anatomy
![Screenshot from 2020-05-07 16-14-17](https://user-images.githubusercontent.com/47241114/81304782-d2ca3380-907d-11ea-9dc0-58cc955e83f1.png)
![Screenshot from 2020-05-07 16-18-35](https://user-images.githubusercontent.com/47241114/81305291-70256780-907e-11ea-840d-05e2b4089d55.png)
AP Biology
![Screenshot from 2020-05-07 16-16-04](https://user-images.githubusercontent.com/47241114/81304993-16bd3880-907e-11ea-82a9-5e5467178415.png)
Biology 2e
![Screenshot from 2020-05-07 16-17-31](https://user-images.githubusercontent.com/47241114/81305157-49673100-907e-11ea-8aff-a24f2ca8becc.png)
Chemistry 2e
![Screenshot from 2020-05-07 16-21-20](https://user-images.githubusercontent.com/47241114/81305612-d4e0c200-907e-11ea-84a8-44b694868fa4.png)
Chemistry AF
![Screenshot from 2020-05-07 16-22-30](https://user-images.githubusercontent.com/47241114/81305786-09547e00-907f-11ea-8faa-89a44bb137bd.png)
Concepts of Biology
![Screenshot from 2020-05-07 16-23-44](https://user-images.githubusercontent.com/47241114/81305940-36089580-907f-11ea-9f0f-1781f713014f.png)
HS Physics
![Screenshot from 2020-05-07 16-25-24](https://user-images.githubusercontent.com/47241114/81306116-65b79d80-907f-11ea-9446-13b18e232798.png)
Microbiology
![Screenshot from 2020-05-07 16-26-22](https://user-images.githubusercontent.com/47241114/81306239-8849b680-907f-11ea-84cf-23be4c218827.png)
College Physics
![Screenshot from 2020-05-07 16-27-41](https://user-images.githubusercontent.com/47241114/81306391-b5966480-907f-11ea-8907-3dede857d05d.png)
University Physics
![Screenshot from 2020-05-07 16-28-38](https://user-images.githubusercontent.com/47241114/81306514-da8ad780-907f-11ea-9b4f-b4ed6d9524c1.png)

lwtchu:
What are the specs currently? There's been a few changes since I last saw these

brittweinstein:
[...]

PR #4583, opened 2022-03-25
fix toc unit link style in Carnival

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/2288

### DIFF
--- a/styles/design-settings/carnival/parts/_toc-settings.scss
+++ b/styles/design-settings/carnival/parts/_toc-settings.scss
@@ -158,7 +158,7 @@
     TocUnitLink: (
       font-family: (_ref: 'typography:::sansFont'),
       font-size: font-scale(4),
-      line-height: line-height(3),
+      line-height: font-scale(4), // should be the same as font size
       font-weight: bolder,
       color: (_ref: 'colorMap:::tocUnitColor'),
       border-bottom: "0.1rem solid black",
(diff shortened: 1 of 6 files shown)


styles/design-settings/carnival/parts/_toc-settings.scss
```diff
@@ -158,7 +158,7 @@
     TocUnitLink: (
       font-family: (_ref: 'typography:::sansFont'),
       font-size: font-scale(4),
-      line-height: line-height(3),
+      line-height: font-scale(4), // should be the same as font size
```
ajluk, line 161: @wilkus27 Is the usage of `font-scale()` function as the `line-height` value intentional? Or should it be rather line-height(4)?

ajluk (commented):
@wilkus27 I left one question, because I wasn't sure if the `line-height` function should be replaced by `font-scale` or the value should rather be the same and it was just a typo.

styles/design-settings/carnival/parts/_toc-settings.scss
wilkus27: @ajluk I explained this in the issue: https://github.com/openstax/cnx-recipes/issues/2288#issuecomment-1079026629. Sorry I should left a comment in the PR as well :smile:

ajluk (approved):

Issue #4595, opened 2022-03-31
carnival theme (A&P) - reduce font size for References


wilkus27:
Font size for paragraphs in References is bigger then in other parts of the book:

![image](https://user-images.githubusercontent.com/40228854/161045833-89f48a47-82ad-468f-bf07-368b00734276.png)

wilkus27:
@lwtchu I set a basic font size for paragraphs and applied two columns. This is how it looks:

![image](https://user-images.githubusercontent.com/40228854/161052130-77f2baef-daed-4642-98f4-17a4524f090d.png)

wilkus27:
Decision: it doesn't reduce page count a lot. We are not going to apply it.

lwtchu:
@wilkus27 sorry I didn't realize the font was bigger and got too focused on the two column - not a big deal, but can we set the font to the base size but leave as one column?

Stackblocks:
- [x] Displays OK with current code & content

![image.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/a98d25dd-de72-45ec-ac2d-9c686eb0e094)

PR #4596, opened 2022-03-31
reduce font size in references

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/4595


wilkus27:
Decision: it doesn't reduce page count a lot. We are not going to apply it.

PR #4618, opened 2022-04-20
Refactor caption container

wilkus27:
Issue: https://github.com/openstax/cnx-recipes/issues/4620


styles/designs/carnival/parts/_caption-components.scss
```diff
@@ -103,25 +103,6 @@ $Caption__Title: (
 $Caption__Container--Table: (
     _name: "CaptionContainerTable",
     _subselector: ' .os-caption-container',
-    _properties: (
-        font-size: font-scale(0),
```
wilkus27, line 107: There is no need to set font size. `font-size(0)` is `1rem`, so it's a basic font size set by book root (`12px`).

ajluk (approved):
Looks good! :relaxed:

Issue #4619, opened 2022-04-20
Baked styles refactor - appendix glossary


lwtchu:
- [x] Carnival - change files names from Appendix to Appendix Glossary
- [x] Other parts of regular Appendix are styled by other shapes (e.g. PageTitles).
- [x] Carnival - check if section (group) should be styled like in other themes. Now there is only a section wrapper. - Note from Larissa: when you get to this can someone explain? should we approach this in a different way? 
- [x] Unordered list - define component in lists file no appendix one.
- [x] Unordered list - review specs (probably should be styled the same way everywhere).
- [x] In Cosmos there are two additional components. Should they be styled in other themes as well? - note from Larissa - what are the two other components? term and list elements?
- [x] Use one components names convention in every theme.

ajluk:
@lwtchu The section (group) in carnival theme only inherits the margin-bottom (0.7rem) from other common styles: 
![image](https://user-images.githubusercontent.com/55203598/164728228-bc793676-1694-4326-b529-aeddce9c5514.png)
But in cardboard and cosmos the section has it's own margin-bottom set to 1.4rem: 
![image](https://user-images.githubusercontent.com/55203598/164728529-ecbbc501-dfb3-4792-8806-797ba4727e95.png)
Should we standardize it be the same on every theme - which option should we keep then: 0.7 or 1.4rem

ajluk:
@lwtchu about the additional appendix elements styled in cosmos these are list item (`li`) which was created for adding indentation, should we also add the indentation for the other themes?:
![image](https://user-images.githubusercontent.com/55203598/164735159-9f84840c-7ab9-47bd-833d-f29409ccdbb2.png)
and `strong` elemnt which has additional padding-right: 3px set: 
![image](https://user-images.githubusercontent.com/55203598/164735502-134bc0eb-6e75-49b5-b4d2-3d78c0c63ab9.png)
Should we also add the padding to string elements in the other themes? This is how it currently looks in Microbiology (carnival): 
![image](https://user-images.githubusercontent.com/55203598/164735730-308529ce-9bff-45f5-9585-8737f12c612a.png)

ajluk:
One more thing about the `column-gap` between glossary columns, do we also want to standardize it in every theme? In carnival the value of the gap is not set and it looks like this by defult: 
![image](https://user-images.githubusercontent.com/55203598/164738977-540be065-6407-4098-a659-6ca943843f64.png)
[...]

PR #4624, opened 2022-04-22
refactor appendix files theme wide

ajluk:
for:  #4619

### DIFF
--- a/styles/designs/cardboard/parts/_appendixglossary-components.scss
+++ b/styles/designs/cardboard/parts/_appendixglossary-components.scss
@@ -1,5 +1,5 @@
-$Appendix--Glossary__Container: (
-  _name: "Container",
+$AppendixGlossary__Container: (
+  _name: "AppendixGlossaryContainer",
   _subselector: '',
   _properties: (
     column-count: 3,
@@ -7,11 +7,19 @@ $Appendix--Glossary__Container: (
   )
 );
 
-$Appendix--Glossary__Group: (
-  _name: "Group",
+$AppendixGlossary__Group: (
+  _name: "AppendixGlossaryGroup",
   _subselector: ' > section',
   _properties: (
-    margin-bottom: 0.7rem,
+    margin-bottom: v-spacing(1),
+  )
+);
+
+$AppendixGlossary__Term: (
+  _name: "AppendixGlossaryTerm",
+  _subselector: " > strong",
+  _properties: (
+    padding-right: 3px,
   )
 );
 
--- a/styles/designs/cardboard/parts/_appendixglossary-components.scss
+++ b/styles/designs/cardboard/parts/_appendixglossary-components.scss
@@ -15,12 +15,3 @@ $Appendix--Glossary__Group: (
   )
 );
 
-$Appendix--Glossary__Group__UnorderedList: (
-  _name: "UnorderedList",
-  _subselector: " > ul",
-  _properties: (
-    font-size: font-scale(-1),
-    list-style-type: none,
-    margin-left: 0, // needs to be checked if it is needed when the computer-science content will avaialable 
-  ),
-);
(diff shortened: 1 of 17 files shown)


styles/designs/cardboard/parts/_appendixglossary-components.scss
```diff
@@ -3,24 +3,15 @@ $Appendix--Glossary__Container: (
...
+    column-gap: h-spacing(2),
   )
 );
 
 $Appendix--Glossary__Group: (
   _name: "Group",
   _subselector: ' > section',
   _properties: (
-    margin-bottom: column-v-spacing(1),
+    margin-bottom: 0.7rem,
```
wilkus27, line 14: Maybe we should use `v-spacing(1)` here? It's better not to use hard coded values.

styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -105,3 +105,31 @@ $StepwiseList__CustomBullet: (
 );
 
 $StepwiseList__Content: empty_wrapper(StepwiseListContent, ' > .os-stepwise-content');
+
+// Appendix
+
+$UnorderedList--AppendixGlossary: (
+  _name: "AppendixGlossaryGroupUnorderedList",
```
wilkus27, line 112: Can we remove part `Group` from the names? It appears in a few components.

styles/designs/cardboard/parts/_lists-components.scss
```diff
@@ -105,3 +105,31 @@ $StepwiseList__CustomBullet: (
...
+  )
+);
+
+$ListItem__Term--AppendixGlossary: (
+  _name: "ListedTerm",
+  _subselector: " > strong",
+  _properties: (
+    padding-right: 3px,
+  )
+);
```
wilkus27, line 135: Strong tag is not part of a list. Can we move this component back to Appendix Glossary file? Maybe it could be named `AppendixGlossary__Term` :thinking:

styles/designs/cardboard/parts/_appendixglossary-components.scss
```diff
@@ -3,24 +3,15 @@ $Appendix--Glossary__Container: (
   _subselector: '',
   _properties: (
     column-count: 3,
-    column-gap: h-spacing(4),
+    column-gap: h-spacing(2),
   )
 );
 
 $Appendix--Glossary__Group: (
```
wilkus27, line 10: What do you think to change name everywhere from `Appendix--Glossary` to just `AppendixGlossary`? Previously there where some files/shapes for Appendices, but now we have only Appendix Glossary, so maybe we don't have to treat it as a type of Appendix :thinking:

styles/designs/carnival/parts/_appendixglossary-shapes.scss
```diff
@@ -0,0 +1,26 @@
+@import 'appendixglossary-components';
+
+@include create_shape('carnival:::AppendixGlossaryShape', (
+  _components: (
+    map-merge($Appendix--Glossary__Container, (
+      _components: (
+        map-merge($Appendix--Glossary__Group, (
+          _components: (
+            $H3--AppendixGlossary,
```
wilkus27, line 9: Titles are not imported to this file.

wilkus27: And lists.

wilkus27 (changes requested):
I left a few ideas. Let me know what do you think :wink:

styles/designs/cardboard/parts/_lists-components.scss
ajluk: Yes, that's right. I kept the name :slightly_smiling_face:

styles/designs/cardboard/parts/_appendixglossary-components.scss
ajluk: Yes, good point!

styles/designs/carnival/parts/_appendixglossary-shapes.scss
wilkus27: Hmm :thinking: That's interesting. What do you think to put this into cnx-recipes clean up epic https://github.com/openstax/cnx-recipes/issues/4635? I mean, to investigate this problem during that work?

wilkus27 (approved):

Issue #4636, opened 2022-04-29
cnx-recipes dockerfile


kswanson33:
AC: Take out installs from dockerfile that are no longer needed. Make sure that all required scripts can still run in the container (& beware of caching: run build with the `--no-cache` flag)

kswanson33:
the only real thing to test is that the dockerfile has been reduced & scripts still works, & multiple devs have already verified this, so I'm just going to close

PR #4641, opened 2022-04-29
pare down dockerfile

kswanson33:
#4636


wilkus27 (approved):
Everything looks good :relaxed: Scripts are working in a container and Enki as well.

Issue #4621, opened 2022-04-21
Refactor `./script/run` for cnx-recipes devs


kswanson33:
[this script](https://github.com/openstax/cnx-recipes/blob/main/script/run) remains useful to dev workflows, but many dependencies in the repo have changed since it was last updated, and we have found ways that it is ... inefficient

Acceptance Criteria:
- [x] `./script/run` allows for portable lightweight development & can containerize `./script/build-styles` and `./script/build-impacted-styles` and any single `dart ./script/build/build.dart <in> <out>` script. (do we use more scripts than that these days??)
- [x] `./script/run` uses Dockerfile in cnx-recipes
- [x] file changes from running various build scripts persist outside the container & after the container has been stopped
- [x] run script has been simplified; no longer uses docker-compose
- [x] run script doesn't leave containers running or create extra images (check that it can be run many times without creating a new image each time)
- [x] files created by the container have [permissions](https://gist.github.com/philschatz/b2d0f8957c98d3e69ba825722929b652#volume-permissions) f[ixed](https://github.com/openstax/enki/blob/main/dockerfiles/10-fix-perms.sh)

lwtchu:
Humanreadable Goal: create lightweight container-ized version ; but too many skeletons. need to deskeletonize script

lwtchu:
MJ learning bash for this. so this is why it's in icebox

lwtchu:
Update: Weird docker issues blocking

PR #4655, opened 2022-05-06
Docker Run migration

LostMojo:
#4621 

Changes the script/run command to use a single docker container to run commands.

### DIFF
--- a/script/run
+++ b/script/run
@@ -9,9 +9,8 @@ do
   shift
 done
 
-if [ ! -d node_modules ]
-then
-  docker-compose run --rm command "./script/setup"
-fi
-docker-compose run --rm "${env_vars[@]}" command "$@"
-docker-compose run --rm command /bin/bash -c "find ! -path './node_modules/*' ! -path './venv/*' ! -path './vendor/*' -user root -exec chown \"\$(stat -c '%u:%g' ./)\" '{}' \\;"
+docker build -t cnx-recipes-image .
+docker run -d -t --rm --mount type=bind,source="$PWD",target=/code --name styles_temp_container cnx-recipes-image /bin/bash
+echo "Executing $@..."
+docker exec "${env_vars[@]}" styles_temp_container "$@"
+docker stop styles_temp_container > /dev/null    # mutes "styles_temp_container" output


script/run
```diff
@@ -9,9 +9,8 @@ do
...
 done
 
-if [ ! -d node_modules ]
-then
-  docker-compose run --rm command "./script/setup"
-fi
-docker-compose run --rm "${env_vars[@]}" command "$@"
-docker-compose run --rm command /bin/bash -c "find ! -path './node_modules/*' ! -path './venv/*' ! -path './vendor/*' -user root -exec chown \"\$(stat -c '%u:%g' ./)\" '{}' \\;"
+docker build -t run-image .
+docker run -d -t --rm --mount type=bind,source="$PWD",target=/code --name runcontainer run-image /bin/bash
```
kswanson33, line 13: nice! runcontainer could also be named something like styles_container or styles_temp_container (more descriptive if you have multiple containers going)

script/run
```diff
@@ -9,9 +9,8 @@ do
...
+echo "Executing $@..."
+docker exec "${env_vars[@]}" runcontainer "$@"
+docker stop runcontainer > /dev/null    # mutes "runcontainer" output
```
kswanson33, line 16: do we want to mute output?

kswanson33 (changes requested):
The changes to `script/run` are looking good! I'll pull this branch down and test it locally, but to my knowledge it should work.

I'm wondering if we still need to worry about env vars. Do you ever use them? If not, it might be good to park this with the Katalyst folks & check

script/run
LostMojo: Well, it usually just prints out its name, "runcontainer" which might be confusing to some folks. I just did this to clean it up a little, but if we want it, it can be unmuted

script/run
LostMojo: ooo, good idea, changed

Dockerfile
```diff
@@ -108,28 +59,20 @@ FROM build-os-dependencies as build-dependencies
...
 ENV SKIP_MY_POSTINSTALL=true
 
-RUN ./script/setup
-
 FROM build-dependencies as code
 
 # Install code
 COPY . ./
-
-# dart pub get doesn't find the pubspec file until after the above line is run
```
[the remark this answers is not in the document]
LostMojo: hm, I'm not sure why it's not showing the restored file, but I know I did it :T for monday I guess

kswanson33:
tested the container & all seems good :-) once extraneous files are removed, this can be merged! i'd recommend:

```bash
git rebase j.j    # brings this branch up to date. `git merge j.j` is another option
git reset --soft j.j
# Discard all chances except those to /script/run ...
git commit -m "..."
git push -f    # force push so that the commit tree will be clean on github as well as locally
```

kswanson33 (approved):

PR #4658, opened 2022-05-12
Include webview-generic.css sourcemap

philschatz:
That way debugging the styles in REX is easier because the original LESS/SASS files will be in the browser dev tools instead of the compiled CSS file.

Alternatively, is there a command that enki could run to build the CSS file? Then the CSS file (& source map) would not need to be committed to the repo at all

cc @TylerZeroMaster


philschatz:
Closing after discussion with @kswanson33 that this was not desired yet

Issue #4647, opened 2022-05-02
marketing - unit opener


lwtchu:
![Screen Shot 2022-05-02 at 2 30 58 PM](https://user-images.githubusercontent.com/11299125/166312899-e8493786-860d-4d05-91c3-81871ebe2a50.png)

Similar to https://github.com/openstax/cnx-recipes/issues/3936

Change the Archivo title fonts to Roboto Condensed
Base font to Noto Sans
Red to: B2: `#344456`
Faint red to: Grey1: `#F9F9F9`

Stackblocks:
- [x] Displays OK with current code & content

![2022-05-24_15-22-04.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/f0227263-31c8-4704-9099-c9e21ca65b76)

![2022-05-24_15-20-47.png](https://images.zenhubusercontent.com/5af1f4c24b5806bc2bd257f2/3c3283b2-7c4f-4ec8-bee7-9de7fead9e10)

PR #4666, opened 2022-05-18
Marketing unitopener

LostMojo:
#4647 

Screen Shot 2022-05-18 at 3 17 54 PM

(Sorry for partial screenshot)


kswanson33 (commented):
Looks fine to me, but I think @ajluk is the real expert since she worked on this for cosmos. I'm wondering if some of this could be reused between themes?

ajluk:
@kswanson33 I did only part of the job and @wilkus27 finished it up as I remember :smile: But about making the components/shape reusable between themes I think it is a good idea. I checked that the components utilizes  the same properties, so we can move the components and shapes files to `designs/common/pdf` where we already have footnotes files and only create a `unitopener-settings` file in a theme which utilizes it. @wilkus27 I have added you here if you would like to participate :slightly_smiling_face:  @LostMojo regarding this issue there is a change needed to be made also in recipe: `BakeUnitPageTitle`direction needs to be added to the marketing recipe file.

wilkus27:
@ajluk everything looks good to me too :wink: Let me put a question about moving unit opener components to `common` file to Baked Styles refactoring spreadsheet. During refactoring for `common` files we have to decide what exactly we want to keep there, so we will see later what would be the best for unit opener element :relaxed:

kswanson33 (approved):

PR #4687, opened 2022-05-25
use css var to scale font size

jivey:
Background/Context: REX is adding a text resize feature that will scale the font size and line height of the book content.

This change updates the webview font size style to use a `calc` to scale it based on a CSS var. On the REX side, we can change this var and have everything update. A downside is all future typography styles will need to use this function.


ajluk (approved):

lwtchu:
@jivey yeah go ahead and merge :eyes: sorry should've kept an eye on this. but usually the person who opens the PR merges after it's reviewed

PR #4698, opened 2022-05-31
Blockquote Refactor

LostMojo:


styles/design-settings/carnival/parts/_blockquote-settings.scss
```diff
@@ -4,4 +4,3 @@
       _selectors: ("blockquote"),
     ),
   ));
```
kswanson33, line 6: looks like blockquote-settings can be deleted from all themes too!

kswanson33 (changes requested):
much simpler! seems to me like you got everything except the `_blockquote-settings.scss` files (and their imports in the `_settings.scss` files) :tada:

kswanson33 (approved):

